PR #14, opened 2018-03-29
fix(docker): Disable TTY connection when using exec

Cawllec:
- Adds -T flag to docker-compose exec calls to stop TTY related errors

This fixes an issue with docker-compose failing exec commands with the message `the input device is not a TTY`.  While I don't believe we need to worry about `tty` commands as any `execs` on containers should be running commands, not opening terminals.

Let me know if there may be a circumstance we won't want this flag and I can arg it out.

### DIFF
--- a/lib/features/support/docker.rb
+++ b/lib/features/support/docker.rb
@@ -63,7 +63,7 @@ def stop_stack(compose_file=$docker_compose_file)
 end
 
 def run_command_on_service(command, service, compose_file=$docker_compose_file)
-  run_docker_compose_command(compose_file, "exec #{service} #{command}")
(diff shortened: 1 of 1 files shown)


kattrali:
What use case does this fix? Can you give an example of a command which is currently failing?

Cawllec:
So while we aren't sure why (because it runs fine manually) the `docker-compose exec sidekiq bundle exec ruby initializers/UnhandledError.rb` command fails with the `the input device is not a TTY`.  As pointed out by Sach, the command shouldn't require TTY anyway as it isn't operating a shell, but it the `docker-compose` `exec` behaviour defaults to using `docker exec -it`.

PR #15, opened 2018-04-06
feature(Sessions): Include sessionCounts in check for valid sessions

Cawllec:
In order to allow testing of server-side session tracking, this needs to be able to handle both sessions and sessionCounts arrays.

### DIFF
--- a/lib/features/steps/request_assertion_steps.rb
+++ b/lib/features/steps/request_assertion_steps.rb
@@ -137,6 +137,14 @@ Then(/^the payload field "(.+)" is a parsable timestamp in seconds(?: for reques
   assert_not_nil(parsed_time)
 end
 
+Then(/^the payload has a valid sessions array(?: for request (\d+))?$/) do
+  sessions = read_key_path(find_request(request_index)[:body], "sessions")
+  sessionCounts = read_key_path(find_request(request_index)[:body], "sessions")
+  value = sessions || sessionCounts
+  assert_kind_of Array, value
+  assert(value.length > 0, "the payload must contain a non empty sessions or sessionCounts array")
+end
+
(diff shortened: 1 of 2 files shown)


lib/features/steps/request_assertion_steps.rb
```diff
@@ -137,6 +137,14 @@ def find_request(request_index)
...
 end
 
+Then(/^the payload has a valid sessions array(?: for request (\d+))?$/) do |request_index|
+  sessions = read_key_path(find_request(request_index)[:body], "sessions")
+  sessionCounts = read_key_path(find_request(request_index)[:body], "sessionCounts")
+  value = sessions || sessionCounts
+  assert_kind_of Array, value
+  assert(value.length > 0, "the payload must contain a non empty sessions or sessionCounts array")
+end
+
```
fractalwrench: minor nitpick, `find_request(request_index)[:body]` could be extracted to a variable here

fractalwrench: Should this be in `session_tracking_steps.rb`? I suppose the argument could be made to store it in either file.

Cawllec:
Be aware I've removed the checks in `the request is a valid for the session tracking API` on the sessions array, so the steps may need to be added in to existing tests.

fractalwrench:
Looking at the [Apiary docs](https://bugsnagsessiontrackingapi.docs.apiary.io/#reference/0/session/report-a-session-starting) I think it may make sense to update that step so that for each element, it validates that:

`startedAt` is always present
either `id` or `sessionsStarted` is present

What are your thoughts, is this overkill or worth doing?

Cawllec:
I think that sounds good, good it fits with payload structure validation.

PR #16, opened 2018-04-06
Add step to check null request part

fractalwrench:
Adds a step to check that a request part is null.


Cawllec:
Is there a way we can make it clearer this is for multipart requests?

fractalwrench:
What syntax would work better? Perhaps something like `The part "(.+)" for multipart request is null`?

We'd also need to update other steps and any existing scenarios.

fractalwrench:
@kattrali @bengourley do you have thoughts on the above?

fractalwrench:
Updated the syntax as discussed.

PR #18, opened 2018-04-09
Add starts with step for stack frame

fractalwrench:
Adds a step to verify that a field in a stack frame starts with a particular string value.


Cawllec:
In the light of my previous issues with regexes, can we add tests for these?

fractalwrench:
I've added a test for the stack frame steps.

PR #21, opened 2018-04-09
Fix /dev/null output on windows

martin308:
The first of probably a few PRs to get `maze-runner` running on Windows.

`/dev/null` is not a valid path on windows, this seems to be the equivalent.


lib/features/support/env.rb
```diff
@@ -6,6 +6,7 @@
 MOCK_API_PORT = 19291
 SCRIPT_PATH = File.expand_path(File.join(File.dirname(__FILE__), "..", "scripts"))
 FAILED_SCENARIO_OUTPUT_PATH = File.join(Dir.pwd, 'maze_output')
+DEV_NULL = Gem.win_platform? ? 'NUL:' : '/dev/null'
```
fractalwrench: Is `NUL` supposed to have a colon after it?

martin308: weird, I copied that from an example on stack overflow but it seems like most other search results do not have the colon. It seems to work either way but I can remove it to prevent confusion

PR #22, opened 2018-04-09
Fix signal usage

martin308:
I'm not sure if there is a reason for using SIGHUP but this is not supported on windows. It doesn't seem like we need to care if the process gets a chance to clean up as it is just our test scripts running and is only called after the scenario.

Another potential solution for this is to split based on the OS and use `taskkill` or similar on windows

### DIFF
--- a/lib/features/support/env.rb
+++ b/lib/features/support/env.rb
@@ -119,7 +119,7 @@ end
 def kill_script
   @pids.each {|p|
     begin
-    Process.kill("HUP", p)
+    Process.kill("KILL", p)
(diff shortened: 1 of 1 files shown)


Cawllec:
Would SIGTERM be better?  Then if it needs to be handled for graceful shutdown it can be.  Is it compatible with Windows?

martin308:
It is not compatible, only `KILL` 🔪. Presumably as windows doesn't actually use signals under the hood.

martin308:
We could revisit alternate signaling if/when someone has an issue with `KILL`?

PR #28, opened 2018-04-18
Changes to mock server to support Safari (on BrowserStack) and IE8/9

bengourley:
A couple of tweaks that will allow the mock api server to recieve requests from Safari on BrowserStack and IE8/9.

### DIFF
--- a/lib/features/support/env.rb
+++ b/lib/features/support/env.rb
@@ -3,7 +3,17 @@ require 'open3'
 require 'webrick'
 require 'json'
 
-MOCK_API_PORT = 19291
+# This port number is semi-arbitrary. It doesn't matter for the sake of
+# the application what it is, but there are some constraints due to some
+# of the environments that we know this will be used in – namely, driving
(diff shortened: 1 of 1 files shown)


lib/features/support/env.rb
```diff
@@ -167,14 +177,15 @@ def read_key_path hash, key_path
...
   def do_POST request, response
     case request['Content-Type']
-    when 'application/json'
-      stored_requests << {body: JSON.load(request.body()), request:request}
     when /^multipart\/form-data; boundary=([^;]+)/
       boundary = WEBrick::HTTPUtils::dequote($1)
       body = WEBrick::HTTPUtils.parse_form_data(request.body(), boundary)
       stored_requests << {body: body, request: request}
     else
-      stored_requests << {body: request.query, request:request}
```
kattrali: I'm fairly certain this logic isn't used anymore and was replaced by the multipart parser, though @fractalwrench may want to double check on the gradle plugin.

fractalwrench: lgtm

PR #29, opened 2018-04-24
Improve ENV variable setting

martin308:
This method of setting ENV variables does not work on windows. 

We can let ruby set the env variables as part of the command so that we don't need to worry about the formatting of them for cross platform compatibility in the same way that `run_script` does

### DIFF
--- a/lib/features/support/env.rb
+++ b/lib/features/support/env.rb
@@ -96,6 +96,8 @@ def run_command(*cmd, must_pass: true)
   STDOUT.puts stderr if ENV['VERBOSE']
 
   assert_true(status.success?) if must_pass
+
+  stdout.split("\n")
 end
 
 def set_script_env key, value
--- a/lib/features/support/env.rb
+++ b/lib/features/support/env.rb
@@ -89,25 +89,13 @@ def encode_query_params hash
   URI.encode_www_form hash
 end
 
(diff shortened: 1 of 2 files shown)


lib/features/support/env.rb
```diff
@@ -89,25 +89,13 @@ def encode_query_params hash
...
+  stdout, stderr, status = Open3.capture3(*cmd)
+  STDOUT.puts stdout if ENV['VERBOSE']
+  STDOUT.puts stderr if ENV['VERBOSE']
 
-    thread.join
-    command_status = thread.value
-  end
-  assert_true(command_status.success?) if must_pass
-  command_out
+  assert_true(status.success?) if must_pass
```
Cawllec: Not having stdout return is breaking some of the docker functionality atm, can we return `stdout.split("\n")`?

martin308: Sure we can return that. I only removed the return as it didn't seem like any of the methods were using the return value. What functionality broke?

Cawllec: The `find_default_docker_compose` method needs it to detect whether the compose file was valid, and the `test_service_running` method needs it to determine whether the instance is actually running.

martin308: cool, thanks for the info. Does what I added address that?

PR #30, opened 2018-04-26
feature(Logging): Add output_logs feature for logging after failures

Cawllec:
- Added output_logs method hook to post additional information after failure
- Added docker-specific support for recording which services were used during scenarios

Adds a hook that looks for a method called `output_logs`.  This function, which should be defined within the `env.rb` of the `features/support` folder of the project, should output additional information that will be required for proper debugging.

Added support for recording which docker containers are launched, and an example of an `output_logs` method within the `env.rb` of the test-fixture `docker-app`.


test/fixtures/docker-app/features/support/env.rb
```diff
@@ -0,0 +1,12 @@
...
+    logged_service = service[:service] == :all ? '' : service[:service]
+    command = "logs -t #{logged_service}"
+    begin
+      response = run_docker_compose_command(service[:file], command)
+    rescue => exception
+      response = "Couldn't retreive logs for #{service[:file]}:#{logged_service}"
+    end
+    STDOUT.puts response.is_a?(String) ? response : response.to_a
+  end
+end
```
fractalwrench: \n

lib/features/support/env.rb
```diff
@@ -21,6 +21,7 @@
 Before do
   stored_requests.clear
   find_default_docker_compose
+  clear_docker_services
```
fractalwrench: My understanding is that this will execute code which is specific to Docker on every mazerunner run, when lots of the platforms don't use Docker. Are there good reasons why we need to do it this way?

lib/features/support/env.rb
```diff
@@ -32,7 +33,10 @@
 After do |scenario|
   kill_script
 
-  write_failed_requests_to_disk(scenario) if scenario.failed?
+  if scenario.failed?
+    write_failed_requests_to_disk(scenario)
+    output_logs if defined? output_logs
```
fractalwrench: This is a very good idea :) I think this will be really handy for getting information on what the state of the system is like when the iOS simulator does weird things.

lib/features/support/env.rb
Cawllec: Good question - The answer is that we shouldn't.  Those functions should be in the env file of the project that uses them.  I'll remove both `clear_docker_services` and `find_default_docker_compose`.  @martin308 this may be a breaking change for you?

martin308: I think it should be ok. It looks like I'm going to need to write a whole bunch of custom stuff anyway due to the way docker works/doesn't work on windows

test/fixtures/docker-app/features/support/env.rb
Cawllec: This has been added, despite the lack of indication.

Cawllec:
@fractalwrench I've added a section in the readme, would you be able to give it a look?
@bugsnag/platforms In the interest of not-breaking specific setups, would people be able to run their maze-tests with this branch to verify it's not breaking anything?

fractalwrench:
@Cawllec the readme looks good.

PR #31, opened 2018-04-27
Fix typo in request error_reporting_steps.rb

bengourley:
I made this change backwards compatible so that it doesn't break anyone's tests.

Ruby didn't like this change though, until I added an argument for the regex capture, even though I knew I would never use it. I called it `_`  – is that idiomatic (or idotic 😂) Ruby? Or is there a better way in the regex to declare a sequence of characters as optional without capturing them?


Cawllec:
Yeah, that's fine.  Unused variables within a block should typically be prepended with `_`, so I guess if it's a nameless variable it makes sense

snmaynard:
Make it a non capturing group with `(?: a)?`

PR #32, opened 2018-05-02
Update expected version header

martin308:
As per https://github.com/bugsnag/bugsnag-dotnet/pull/94#pullrequestreview-116802884, 4 is preferred


martin308:
Updated to support both values

PR #34, opened 2018-05-09
Add steps for making assertions about query parameters

bengourley:
More IE8/9 requirements!

We need to make assertions about the query parameters instead of the headers.

### DIFF
--- a/lib/features/steps/request_assertion_steps.rb
+++ b/lib/features/steps/request_assertion_steps.rb
@@ -2,12 +2,22 @@ require 'test/unit'
 require 'minitest'
 require 'open-uri'
 require 'json'
+require 'cgi'
 
 include Test::Unit::Assertions
+
+timestamp_regex = /^\d{4}\-\d{2}\-\d{2}T\d{2}:\d{2}:[\d\.]+Z?$/
 
 def find_request(request_index)
   request_index ||= 0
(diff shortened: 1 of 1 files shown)


lib/features/steps/request_assertion_steps.rb
```diff
@@ -33,7 +43,20 @@ def find_request(request_index)
...
+  assert_match(timestamp_regex, header)
+end
+
+Then(/^the "(.+)" query parameter equals "(.+)"(?: for request (\d+))?$/) do |parameter_name, parameter_value, request_index|
+  assert_equal(parameter_value, parse_querystring(find_request(request_index))[parameter_name][0])
+end
+
+Then(/^the "(.+)" query parameter is not null(?: for request (\d+))?$/) do |parameter_name, request_index|
+  assert_not_nil(parse_querystring(find_request(request_index))[parameter_name][0], "The '#{parameter_name}' query parameter should not be null")
+end
```
fractalwrench: Will this be nil rather than an index exception (if that's a thing on Ruby)?

Cawllec: You could do an assert using `include?` to test if `parameter_name` exists in the query string.

bengourley: Seems fine to me:

```

PR #35, opened 2018-05-11
run_command needs assert_true so explicitly require it

bengourley:
These dependencies get loaded later on by the `request_assertion_steps.rb` but my code calls the `run_command` function before that is loaded, causing a the following error:

```
undefined method `assert_true' for main:Object (NoMethodError)
```

One question, I don't know whether this should then be removed from `request_assertion_steps.rb`?

### DIFF
--- a/lib/features/support/env.rb
+++ b/lib/features/support/env.rb
@@ -2,6 +2,8 @@ require 'rack'
 require 'open3'
 require 'webrick'
 require 'json'
+require 'test/unit'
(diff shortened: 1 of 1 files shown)


kattrali:
@bengourley Can you provide an example of how `run_command` is invoking assertions? Not that I doubt the fix works, just to be better understand what's going on.

kattrali:
Ah! Got it.

PR #38, opened 2018-08-08
Session payload request index assertions

bengourley:
In order to address https://github.com/bugsnag/bugsnag-js/pull/381#discussion_r208255754 I needed to make assertions against multiple session payloads in the same test. To do so, the session payload assertions needed the "for request n" suffix, so I added it to them all. It needed a little helper in the case were it wasn't passed in.

I also added this to the readme, as it wasn't mentioned there at all yet.

I used this branch with both the @bugsnag/node and @bugsnag/browser test suites, and they pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,8 +204,12 @@ This request will match:
 { "fruit": { "apple": "some nonsense" } }
 ```
 
+# On dealing with multiple requests
 
-
+Most payload assertion steps have optional suffix of `for request n` where `n` is a
+zero-based index into chronologically ordered collection of requests received by the
+mock server. If you leave off this suffix, the default behaviour is to run the step
+against request `0`, i.e. the first (and perhaps only) request.
 
 ## Running features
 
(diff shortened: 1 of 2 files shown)


fractalwrench:
I tested this out on Android and it seems to break at least one existing step:

```
And the session "user.id" equals "123"                                     # /Users/jamielynch/repos/maze-runner/lib/features/steps/session_tracking_steps.rb:23
      Undefined dynamic step: "the payload field "sessions.0.user.id" equals "123" for  for request 0" (Cucumber::UndefinedDynamicStep)
      /Users/jamielynch/repos/maze-runner/lib/features/steps/session_tracking_steps.rb:24:in `/^the session "(.+)" equals "(.+)"(?: for request (\d+))?$/'
      features/session_tracking.feature:41:in `And the session "user.id" equals "123"'
```

Unfortunately we don't really have a comprehensive test suite in one location for this repo at the moment, so I'd suggest trying it out on a couple of different repos to check for any other incompatible changes. I've already tested android-app, comparison, ios-app, and js-app in the `test/fixtures` directory.

PR #42, opened 2018-10-22
Server steps: Add required commands for Ruby & Laravel server tests

Cawllec:
- Adds in compound steps used in several of the Server maze-tests, such as Ruby, Laravel & Python.
- Adds in `current_ip` support, that gives the current server host for allowing docker-containers to communicate.

### DIFF
--- a/lib/features/steps/compound_steps.rb
+++ b/lib/features/steps/compound_steps.rb
@@ -1,0 +1,31 @@
+When(/^I configure the bugsnag endpoint$/) do
+  steps %Q{
+    When I set environment variable "BUGSNAG_ENDPOINT" to "http://#{current_ip}:#{MOCK_API_PORT}"
+  }
+end
+When(/^I navigate to the route "(.*)" on port "(\d*)"/) do |route, port|
+  steps %Q{
+    When I open the URL "http://localhost:#{port}#{route}"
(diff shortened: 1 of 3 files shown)


lib/features/steps/compound_steps.rb
```diff
@@ -0,0 +1,31 @@
...
+  }
+end
+Then(/^the request contained the api key "(.*)"$/) do |api_key|
+  steps %Q{
+    Then the "bugsnag-api-key" header equals "#{api_key}"
+    And the payload field "apiKey" equals "#{api_key}"
+  }
+end
+Then(/^the request used the "(.*)" notifier$/) do |notifier_name|
+  bugsnag_regex = /^http(s?):\/\/www.bugsnag.com/
```
kattrali: This shouldn't be asserted in a generic step since most notifier libraries use the GitHub URL rather than the homepage.

snmaynard: What should notifiers use here? Is it in the notifier spec? Alex can you make sure the notifier spec is clear once delisa has answered

Cawllec: I've updated the spec, and will replace the regex step with a `not-null` step.

PR #43, opened 2018-10-26
Enabled starting a docker service without building first

Cawllec:
## Goal
Adds support for starting a service without building it first.  While the steps aren't exactly idiomatic atm, there are multiple repositories relying on the current functionality, that would break if functionality was changed in the existing functions.

This is QoL, and isn't necessary for any docker-using maze tests, but would speed up tests where the docker build takes a long time, such as the upcoming Laravel tests.

### DIFF
--- a/lib/features/steps/docker_steps.rb
+++ b/lib/features/steps/docker_steps.rb
@@ -11,11 +11,15 @@ Given("I have built the service {string} from the stack {string}") do |service, 
 end
 
 When("I start the service {string}") do |service|
-  start_service service
+  start_service(service, true)
+end
+
+When("I start the service {string} without building") do |service|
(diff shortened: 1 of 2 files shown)


Cawllec:
When running Laravel containers, despite nothing changing between builds, the build is completed each time due to the `--build` flag against start.  This can take up to 5 minutes to run the composer install command, which makes it difficult to start adding thorough end-to-end tests.

snmaynard:
That means that the code has changed - the build is instant otherwise. This feels like the wrong solution. What has changed to invalidate the build?

Cawllec:
You're quite right, how the happens is quite different now to when I originally created this.  I don't think this is necessary anymore.

PR #47, opened 2018-11-07
Fix type in steps so other steps work

CodeHex:
There should be a space between request and request number


CodeHex:
I did a quick test and since its inside the whole optional ( for request...) if doesn't affect it

PR #48, opened 2018-11-13
Grab common steps and support from go maze tests

kinbiko:


lib/features/support/docker.rb
```diff
@@ -92,6 +92,32 @@ def run_docker_compose_command(file, command, must_pass=true)
...
+      true
+    rescue Errno::ECONNREFUSED, Errno::EHOSTUNREACH
+      false
+    end
+  end
+rescue Timeout::Error
+  false
+end
+
+def wait_for_port(port)
```
snmaynard: Where is this used?

lib/features/steps/request_assertion_steps.rb
```diff
@@ -216,3 +216,22 @@ def validate_error_reporting_thread(payload_key, payload_value, request_index)
   end
   assert_equal(1, count)
 end
+
+Then(/^I wait to receive a request$/) do
```
snmaynard: Whats this for?

lib/features/support/docker.rb
snmaynard: But why doesnt it have an accompanying step?

CodeHex: I think it just got missed, made a PR with the corresponding step https://github.com/bugsnag/maze-runner/pull/49

PR #53, opened 2019-04-09
Breadcrumbs: Add steps asserting specific breadcrumbs are not present

Cawllec:
## Goal

Allows assertions to be made that check specific breadcrumbs are not present in a response.

### DIFF
--- a/lib/features/steps/error_reporting_steps.rb
+++ b/lib/features/steps/error_reporting_steps.rb
@@ -51,45 +51,44 @@ Then("the event {string} matches the JSON fixture in {string}") do |field, fixtu
 end
 
 ## BREADCRUMB ASSERTIONS
+# Checks a breadcrumb with a type and name exists
+#
+# @param type [String] The type field
+# @param name [String] The name field
 Then("the event has a {string} breadcrumb named {string}") do |type, name|
   value = Server.current_request[:body]["events"].first["breadcrumbs"]
-  found = false
-  value.each do |crumb|
-    if crumb["type"] == type and crumb["name"] == name then
-      found = true
-    end
-  end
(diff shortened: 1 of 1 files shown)


lib/features/steps/error_reporting_steps.rb
```diff
@@ -61,6 +61,16 @@
   end
   fail("No breadcrumb matched: #{value}") unless found
 end
+Then("the event does not have a {string} breadcrumb named {string}") do |type, name|
+  value = Server.current_request[:body]["events"].first["breadcrumbs"]
+  found = false
+  value.each do |crumb|
```
snmaynard: We can use none? or any? (depending on what you want the check at the end to be) https://ruby-doc.org/core-2.6.2/Enumerable.html#method-i-none-3F

snmaynard:
I think it is worth looking into step documentation soonish as the number of steps we have is increasing and we want to make sure its easy to write new feature files -> https://codeforhire.com/2012/12/08/documenting-cucumber-steps/

Cawllec:
The use case here is to test that when a breadcrumb has been disabled through configuration (such as setting `ConsoleBreadcrumbsEnabled` to `false` in JS) the breadcrumb will not be present when the relevant event is presented.

lib/features/steps/error_reporting_steps.rb
Cawllec: I was thinking this could be simplified anyway, as the 4 events all require very similar logic.  I'll refactor before this is merged.

Cawllec:
Re documentation, it would be useful to output an html page for gitpages, rubydoc style.  I'll look into the cuke-steps lib.

Cawllec:
Closing this as unnecessary

PR #54, opened 2019-05-09
V2 automator driver

Cawllec:
## Goal

Adds support for browserstack app-automator, and base for browserstack web-automator


lib/features/support/driver.rb
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    at_exit do
+      @driver.quit unless @driver.nil?
+    end
+  end
+
+  def start_local_tunnel
+    status = nil
+    Open3.popen2("/BrowserStackLocal -d start --key #{@access_key} --local-identifier #{@local_id} --force-local") do |stdin, stdout, wait|
```
snmaynard: This isnt in the dockerfile?

lib/features/support/app-automate.rb
```diff
@@ -0,0 +1,63 @@
...
+    @capabilities.merge! devices[target_device]
+    @driver = Appium::Driver.new({
+      'caps' => @capabilities,
+      'appium_lib' => {
+        :server_url => "http://#{@username}:#{@access_key}@hub-cloud.browserstack.com/wd/hub"
+      }
+    }, false).start_driver
+  end
+
+  def stop_driver
```
snmaynard: What is this method called by?

lib/features/support/app-automate.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  attr_reader :device_type
+
+  APP_UPLOAD_URI = "https://api-cloud.browserstack.com/app-automate/upload"
+
+  def devices
+    Devices::DEVICE_HASH
+  end
+
+  def start_driver(target_device, app_location)
```
snmaynard: What is this method called by?

lib/features/support/app-automate.rb
```diff
@@ -0,0 +1,63 @@
+require 'appium_lib'
+require_relative './driver'
+
+class AppAutomateDriver < Driver
+
+  attr_reader :device_type
+
+  APP_UPLOAD_URI = "https://api-cloud.browserstack.com/app-automate/upload"
+
+  def devices
```
snmaynard: Should this be public or private?

lib/features/support/app-automate.rb
```diff
@@ -0,0 +1,63 @@
+require 'appium_lib'
```
snmaynard: I would expect this to be added to deps?

lib/features/support/driver.rb
Cawllec: The start command, or the installation?  The installation is already in the Dockerfile for maze-runner.

lib/features/support/app-automate.rb
Cawllec: This can be called by the local environment of whichever feature set is running.  There's a `@driver.quit` call added to `at_exit` in the `Driver` initialisation, but it may be required by certain feature sets (although it can probably be removed and re-added later if necessary).

lib/features/support/app-automate.rb
Cawllec: This is called by the environment of the feature set being tested.  It could be called from the driver initialisation, but I'd like the drivers to be reasonably "dumb" and controllable.

lib/features/support/app-automate.rb
Cawllec: It's already been added previously

lib/features/support/app-automate.rb
Cawllec: I'm not too worried about methods being private, as it's a potentially useful method which may be used in the future by other feature sets.

Cawllec:
I've added unit & integration tests against the app-automate driver.

lib/features/support/app-automate.rb
```diff
@@ -0,0 +1,121 @@
...
+  # Removes all application data and relaunches the app
+  def reset_app
+    @driver.reset unless @driver.nil?
+  end
+
+  private
+
+  def upload_app(app_location)
+    res = `curl -u "#{@username}:#{@access_key}" -X POST "#{APP_UPLOAD_URI}" -F "file=@#{app_location}"`
+    resData = JSON.parse(res)
```
tobyhs: Use underscore case in Ruby (`res_data`), although I would probably call this something like `response` instead

lib/features/support/fast-selenium.rb
```diff
@@ -0,0 +1,19 @@
+# copy pasta from https://www.browserstack.com/automate/ruby
```
tobyhs: Change the link to https://www.browserstack.com/automate/ruby#add-on as it isn't trivial to find the right section in the page

test/fixtures/browserstack-app/features/support/env.rb
```diff
@@ -0,0 +1,13 @@
+# Configure app environment
+require 'pp'
```
tobyhs: This appears to be unused now

lib/features/support/app-automate.rb
```diff
@@ -0,0 +1,121 @@
+require 'appium_lib'
```
tobyhs: It might be preferable to rename this file to app_automate_driver to maintain consistency with the class that this file defines.

lib/features/support/app-automate.rb
```diff
@@ -0,0 +1,121 @@
...
+      }
+    }, false).start_driver
+  end
+
+  # Checks for an element, waiting until it is present or the method times out
+  #
+  # @param element_id [String] the element to search for using the @locator strategy
+  # @param timeout [Integer] the maximum time to wait for an element to be present
+  def wait_for_element(element_id, timeout=15)
+    unless @driver.nil?
```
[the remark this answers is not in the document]
Cawllec: There are points at which the driver will be `nil` (before `start_driver` is called), so I was guarding against them

test/app_automate_test.rb
```diff
@@ -0,0 +1,279 @@
...
+    appium_mock.expect(:nil?, false)
+    driver.instance_variable_set(:@driver, appium_mock)
+
+    json_response = JSON.dump({
+      :app_url => "App_url"
+    })
+
+    command_call = Minitest::Mock.new
+    command_call.expect(:call, json_response) do |cmd|
+      cmd === "curl -u \"#{USERNAME}:#{ACCESS_KEY}\" -X POST \"https://api-cloud.browserstack.com/app-automate/upload\" -F \"file=@#{APP_LOCATION}\""
```
tobyhs: Use `%()` so you don't need to escape the double quotes (https://github.com/rubocop-hq/ruby-style-guide#percent-q-shorthand)

test/app_automate_test.rb
```diff
@@ -0,0 +1,279 @@
...
+  end
+
+  def test_click_element_locator
+    driver = AppAutomateDriver.new(USERNAME, ACCESS_KEY, LOCAL_ID, :accessibility_id)
+    appium_mock = Minitest::Mock.new
+    driver.instance_variable_set(:@driver, appium_mock)
+
+    appium_mock.expect(:nil?, false)
+    appium_mock.expect(:click, true)
+    appium_mock.expect(:find_element, appium_mock) do |locator, element_id|
```
tobyhs: `find_element` should be stubbed to return a different mock, not `appium_mock` itself

test/app_automate_test.rb
```diff
@@ -0,0 +1,279 @@
...
+class AppAutomateTest < Test::Unit::TestCase
+
+  USERNAME = "Username"
+  ACCESS_KEY = "Access_key"
+  LOCAL_ID = "Local_id"
+  APP_LOCATION = "App_location"
+
+  def test_default_assignments
+    driver = AppAutomateDriver.new(USERNAME, ACCESS_KEY, LOCAL_ID)
+    assert_equal(USERNAME, driver.instance_variable_get( :@username ))
```
tobyhs: Ideally, you would try to avoid using `instance_variable_get` and `send` if you can. Instead, you can test the `@username`, `@access_key`, etc. were used correctly in certain places (like `#start_driver`).

lib/features/support/app_automate_driver.rb
```diff
@@ -0,0 +1,122 @@
...
+      }
+    }, false).start_driver
+  end
+
+  # Checks for an element, waiting until it is present or the method times out
+  #
+  # @param element_id [String] the element to search for using the @locator strategy
+  # @param timeout [Integer] the maximum time to wait for an element to be present in seconds
+  def wait_for_element(element_id, timeout=15)
+    unless @driver.nil?
```
tobyhs: As followup on https://github.com/bugsnag/maze-runner/pull/54#discussion_r284053696 , this design feels awkward because almost every public method checks `@driver.nil?` and each method has a test case that stubs `#nil?` (which doesn't feel right either).

An alternative approach might be to have something separate (whether it is another class or multiple classes or methods) perform the initialization (uploading the APK, starting the tunnel, creating an `Appium::Driver`), and then have this driver wrapper class's `#initialize` method take the `Appium::Driver` as an argument.

(However, I don't really see the benefit of having this `AppAutomateDriver` wrapper yet)

test/fixtures/browserstack-app/app/src/main/java/com/bugsnag/android/mazerunner/MainActivity.kt
```diff
@@ -0,0 +1,34 @@
+package com.bugsnag.android.mazerunner
+
+import android.os.Build
+import android.os.Bundle
+import android.support.v7.app.AppCompatActivity
+import android.util.Log
+import android.widget.Button
+import com.bugsnag.android.*
```
tobyhs: I prefer to avoid wildcard/star imports so it is obvious where the classes/interfaces come from.

test/fixtures/browserstack-app/app/src/main/java/com/bugsnag/android/mazerunner/MainActivity.kt
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    override fun onResume() {
+        super.onResume()
+        initialiseBugsnag()
+    }
+
+    private fun initialiseBugsnag() {
+        val config = Configuration("12312312312312312312312312312312")
+        config.endpoint = "http://localhost:9339"
```
tobyhs: I thought these methods were deprecated and `setEndpoints` was supposed to be used instead.

lib/features/support/app_automate_driver.rb
Cawllec: I agree, but I can't think of a better solution right now.  Certain elements of this class will be abstracted out once the BrowserStack Browser-automation driver is extracted from Bugsnag-JS to here, and a proper design will likely be necessary to avoid too much complication.  Perhaps that'd be a good time to address these concerns?

lib/features/support/app_automate_driver.rb
```diff
@@ -0,0 +1,122 @@
...
+  end
+
+  # Initialises the BrowserStack app-automate connect by:
+  #   Uploading the app file
+  #   Starting the BrowserStackLocal tunnel
+  #   Creates and starts the Appium driver
+  #
+  # @param target_device [String] a key from the Devices array selecting which device capabilities to target
+  # @param app_location [String] the location of the test-app to upload
+  def start_driver(target_device, app_location)
```
snmaynard: Suggestions,

- Use inheritance rather than composition. We could inherit from `Appium::Driver` instead. When you construct one we can start the driver and even upload the app?
- This lets us drop the protections of calling methods on driver if its nil which feel strange.

There are other options that probably make sense, but I think at the moment the abstraction is a little wrong which is why you are having to call driver.quit inside the initializer and protect against driver being nil etc.

Cawllec:
I've made the driver use inheritance, and it seems to work well. I've wiped the unit tests temporarily, but I'm in the process of re-adding them.

Cawllec:
I'm actually not sure how to go about adding unit tests here: the appium lib itself doesn't have them.  Would a set of thorough integration tests suffice?

test/app_automate_driver_test.rb
```diff
@@ -0,0 +1,116 @@
...
+  APP_LOCATION = "App_location"
+  TEST_APP_URL = "Test_app_url"
+  TARGET_DEVICE = "ANDROID_9"
+  LOCAL_TUNNEL_COMMAND = "/BrowserStackLocal -d start --key #{ACCESS_KEY} --local-identifier #{LOCAL_ID} --force-local"
+
+  def test_defaults
+    AppAutomateDriver.any_instance.stubs(:upload_app).returns(TEST_APP_URL)
+    driver = AppAutomateDriver.new(USERNAME, ACCESS_KEY, LOCAL_ID, TARGET_DEVICE, APP_LOCATION)
+
+    assert_equal(ACCESS_KEY, driver.instance_variable_get(:@access_key))
```
tobyhs: Similar to https://github.com/bugsnag/maze-runner/pull/54#discussion_r284584573, shouldn't this be tested by other assertions instead? (in this particular case, checking `driver.server_url` and that username/access_key are used for uploading the app and starting the tunnel).

Similar comment applies to your next test method (`#test_overridden_location`)

test/app_automate_driver_test.rb
```diff
@@ -0,0 +1,116 @@
...
+      :error => "Error"
+    })
+    AppAutomateDriver.any_instance.stubs(:`).returns(json_response)
+    begin
+      driver = AppAutomateDriver.new(USERNAME, ACCESS_KEY, LOCAL_ID, TARGET_DEVICE, APP_LOCATION)
+    rescue Exception => exception
+      assert_equal("BrowserStack upload failed due to error: Error", exception.message)
+    end
+  end
+
```
tobyhs: Test(s) for `#wait_for_element` appear to be absent

test/app_automate_driver_test.rb
```diff
@@ -0,0 +1,116 @@
...
+  def test_click_element_defaults
+    AppAutomateDriver.any_instance.stubs(:upload_app).returns(TEST_APP_URL)
+    driver = AppAutomateDriver.new(USERNAME, ACCESS_KEY, LOCAL_ID, TARGET_DEVICE, APP_LOCATION)
+
+    mocked_element = mock('object')
```
tobyhs: I would call this something like 'element' instead of 'object' as object isn't very descriptive (same for line 98)

test/app_automate_driver_test.rb
```diff
@@ -0,0 +1,116 @@
...
+    Devices::DEVICE_HASH[TARGET_DEVICE].each do |key, value|
+      assert_equal(value, driver.caps[key.to_sym])
+    end
+  end
+
+  def test_upload_app_success
+    json_response = JSON.dump({
+      :app_url => TEST_APP_URL
+    })
+    AppAutomateDriver.any_instance.stubs(:`).returns(json_response)
```
tobyhs: If it isn't too difficult, consider adding an argument/parameter matcher.

test/app_automate_driver_test.rb
```diff
@@ -0,0 +1,148 @@
...
+    assert_raise(RuntimeError, "BrowserStack upload failed due to error: Error") do
+      AppAutomateDriver.new(USERNAME, ACCESS_KEY, LOCAL_ID, TARGET_DEVICE, APP_LOCATION)
+    end
+  end
+
+  def test_click_element_defaults
+    AppAutomateDriver.any_instance.stubs(:upload_app).returns(TEST_APP_URL)
+    driver = AppAutomateDriver.new(USERNAME, ACCESS_KEY, LOCAL_ID, TARGET_DEVICE, APP_LOCATION)
+
+    mocked_element = mock('object')
```
tobyhs: It looks like you changed other lines to use `mock('element')` but forgot to change this line.

test/app_automate_driver_test.rb
```diff
@@ -0,0 +1,148 @@
...
+
+    driver.send_keys_to_element("test_text_entry", "Test_text")
+  end
+
+  def test_start_driver
+    AppAutomateDriver.any_instance.stubs(:upload_app).returns(TEST_APP_URL)
+    driver = AppAutomateDriver.new(USERNAME, ACCESS_KEY, LOCAL_ID, TARGET_DEVICE, APP_LOCATION)
+
+    Appium::Driver.any_instance.expects(:start_driver)
+    Open3.expects(:popen2).with(LOCAL_TUNNEL_COMMAND)
```
tobyhs: https://github.com/bugsnag/maze-runner/pull/54/files#diff-4f9dfd7642654cddb10bc7b8249a9294R95 (`status = wait.value`) isn't covered. To exercise it, I think you can change this line to something like:
```ruby
    waiter = mock('Process::Waiter', value: mock('Process::Status'))
    Open3.expects(:popen2).with(LOCAL_TUNNEL_COMMAND).yields(mock('stdin'), mock('stdout'), waiter)
```

PR #55, opened 2019-05-22
Extend payloadVersion matching for different notifiers

Cawllec:
## Goal

It's been found that the `payloadVersion` location in the body of an error payload isn't consistent across notifiers (and isn't supposed to be).

In JS notifiers each event should contain its own `payloadVersion`, which is reflected in the existing check and Error reporting API document on APIary.

In other notifiers, the `payloadVersion` should be a top level key in the payload, which is what all the other notifiers currently do.

This change enables the testing of either of these positions in the same step, and updates the error reporting API validation step to work on the Android (and other) notifiers.

### DIFF
--- a/lib/features/steps/error_reporting_steps.rb
+++ b/lib/features/steps/error_reporting_steps.rb
@@ -1,9 +1,13 @@
+# Verifies that generic elements of the payload should be present
+#
+# @param payload_version [String] The payload version expected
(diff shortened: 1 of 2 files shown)


snmaynard:
Why is there a difference between the notifiers as to where payloadVersion goes? Where should it be cc @markbsb

PR #56, opened 2019-05-29
Fix variable typo

Cawllec:
## Goal
Fixes incorrect elements with expecting specific payload_version


test/fixtures/browserstack-app/features/support/env.rb
```diff
@@ -5,6 +5,8 @@
 bs_device = 'ANDROID_9'
 app_location = 'app/build/outputs/apk/release/app-release.apk'
 
+$api_key = "12312312312312312312312312312312"
```
kattrali: What is this for?

Cawllec: Currently maze generates a random API key that enables you to use it at runtime.  However as the test fixture is pre-built, it'll need a pre-determined API key.

PR #57, opened 2019-05-30
AppAutomateDriver: Better handling of Timeout errors

Cawllec:
## Goal

Provides easier extending of the `wait_for_element` app automate method by returning `true` or `false` when waiting for an element and allowing the caller to determine how to handle the response, rather than raising a TimeoutError.

### DIFF
--- a/test/app_automate_driver_test.rb
+++ b/test/app_automate_driver_test.rb
@@ -126,6 +126,61 @@ class AppAutomateDriverTest < Test::Unit::TestCase
     assert_false(response, "The driver must return false if it does not find an element")
   end
 
+  def test_wait_for_element_stale_error_retry
+    AppAutomateDriver.any_instance.stubs(:upload_app).returns(TEST_APP_URL)
+    driver = AppAutomateDriver.new(USERNAME, ACCESS_KEY, LOCAL_ID, TARGET_DEVICE, APP_LOCATION, :accessibility_id)
+
+    mocked_element = mock('element')
+    mocked_element.expects(:displayed?).at_least(2).at_most(2).raises(Selenium::WebDriver::Error::StaleElementReferenceError, 'Element is stale').then.returns(true)
+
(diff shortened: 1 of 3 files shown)


test/app_automate_driver_test.rb
```diff
@@ -92,23 +92,38 @@ def test_wait_for_element_defaults
     driver = AppAutomateDriver.new(USERNAME, ACCESS_KEY, LOCAL_ID, TARGET_DEVICE, APP_LOCATION)
 
     mocked_element = mock('element')
-    mocked_element.expects(:displayed?).returns(true)
+    mocked_element.expects(:displayed?).returns(true).at_least_once
```
snmaynard: Whats with all the at_least_once calls?

Cawllec: It only actually needs them in the one scenario, fixed that.  It's necessary in the failure scenario as the Selenium::Wait will result in the `find_element` method being called multiple times, with the `displayed?` method being called each time as well.

test/app_automate_driver_test.rb
```diff
@@ -126,6 +126,61 @@ def test_wait_for_element_failure
     assert_false(response, "The driver must return false if it does not find an element")
   end
 
+  def test_wait_for_element_stale_error_retry
+    AppAutomateDriver.any_instance.stubs(:upload_app).returns(TEST_APP_URL)
+    driver = AppAutomateDriver.new(USERNAME, ACCESS_KEY, LOCAL_ID, TARGET_DEVICE, APP_LOCATION, :accessibility_id)
+
+    mocked_element = mock('element')
+    mocked_element.expects(:displayed?).at_least(2).at_most(2).raises(Selenium::WebDriver::Error::StaleElementReferenceError, 'Element is stale').then.returns(true)
```
snmaynard: Either use `.twice` or `.times(2)`

PR #59, opened 2019-07-02
v2 - Add across-request comparators for session ids

Cawllec:
## Goal

Adds cross-request comparator steps for generic fields and session ids.  Allows the user to compare the value of a field in a current (i.e. first received) request to the value of a field in the next received request.

Contains steps to explicitly compare the session ids from the current request to the next request regardless whether the current request is a session payload or error payload.

## Tests

Added the `comparison tests` fixture set to the build pipeline, with additional features to cover the new comparisons.

### DIFF
--- a/lib/features/steps/request_assertion_steps.rb
+++ b/lib/features/steps/request_assertion_steps.rb
@@ -248,6 +248,6 @@ Then("the payload field {string} in the current request does not match the paylo
   assert(Server.stored_requests.length > 1, "The server does not have more than one request")
   current_value = read_key_path(Server.current_request[:body], current_field)
   next_value = read_key_path(Server.next_request[:body], next_field)
-  result = value_compare()
+  result = value_compare(current_value, next_value)
   assert_false(result.equal?, "The current value #{current_value} equaled the next value #{next_value}")
 end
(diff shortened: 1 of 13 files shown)


lib/features/steps/request_assertion_steps.rb
```diff
@@ -226,4 +226,28 @@ def read_key_path hash, key_path
     assert_not_nil(read_key_path(element, element_key_path),
            "Each element in '#{key_path}' must have '#{element_key_path}'")
   end
+end
+
+# Checks a given payload field for the current request is the same as a payload field in the next request
+#
+# @param current_field [String] the payload element from the current request
+# @param next_field [String] the payload element in the next request
+Then("the payload field {string} in the current request matches the payload field {string} in the next request") do |current_field, next_field|
```
snmaynard: could another approach be instead to read a payload field into a global variable or something? We could have a step like `'the payload field is stored in {string}'` which stores it in `@payload_fields[string] = value` and then in subsequent steps we can check fields are equal to that? Just seems a little more reusable to do something like that than to assume its the next payload.

Cawllec: We could do, I was reluctant to add that sort of complexity.  I guess so long as it's abstracted out enough it should be useable without being clunky.

Cawllec:
@snmaynard Would it be possible to get a re-review?

PR #60, opened 2019-07-10
Default server wait time to 10s if not running in Buildkite

Cawllec:
## Goal
Enable faster local test runs by using different wait times on Buildkite vs locally.

This is risky as reducing the time waiting for responses to arrive may cause test failures and flakiness downstream, even if this should only effect local runs.

The other option would be to keep the default as 30s and enable that to be overridden by an environment variable - which would be more flexible but require more thought when running locally.


Cawllec:
The motivation was from the tests taking too long to fail when running locally, but thinking about it this may wouldn't actually really help the problem.

A better way would be for the test runner to see that the tests were failing locally and look for a solution, or if a particular test output is desired run that test specifically.

PR #61, opened 2019-07-10
Enable slashes in branch names

Cawllec:
## Goal

Enables slashes in branch names to be used in maze-runner v2 using the newly added `BRANCH_NAME` environment variable.


Cawllec:
Ah yes, so `BRANCH_NAME` comes from the `.env` file in the root of our secrets bucket, and ensures that a slash free equivalent of the `BUILDKITE_BRANCH` is available on all our agents.

The logic is `export BRANCH_NAME=${BUILDKITE_BRANCH/\//-}`

PR #62, opened 2019-07-26
Check BrowserStack devices available

Cawllec:
## Goal

Calling `start_driver` now returns a boolean describing whether the driver is able to be started based on the amount of devices currently in use by BrowserStack.

- `start_driver` now returns `true` if a devices slot is available, `false` if it's unavailable
- `browser_stack_available` acquires the current state of BrowserStack, returning `true` if the amount of devices in use are less than the total, `false` if the opposite is true, and raises an Error if it was unable to get a status update.

Before this is merged, minor updates should be prepared for Bugsnag JS and Bugsnag Android, as they should handle the value passed back from `start_driver` and act appropriately.


Cawllec:
This is being closed due to a lack of need at the moment so long as concurrency groups are correctly formulated

PR #65, opened 2019-09-09
V2 unhandled helper steps

Cawllec:
## Goal

Adds simple checks that an error has the correct unhandled/handled flags, default or custom severity, and checks that unhandled/handled counts in sessions are correct.
Further details, such as checking severity reasons should be done on a per-test basis, as that can vary wildly in different notifiers and frameworks.

### DIFF
--- a/lib/features/steps/error_reporting_steps.rb
+++ b/lib/features/steps/error_reporting_steps.rb
@@ -43,8 +43,8 @@ end
 #
 # @param event [Int] The event to verify
 # @param severity [String] Optional. An expected severity different to the default "warning"
-Then(/^event (\d+) is unhandled(?: with the severity "(.+)")$/) do |event, severity|
-  step "the payload field \"events.#{event}.unhandled\" is true"
+Then(/^event (\d+) is handled(?: with the severity "(.+)")$/) do |event, severity|
+  step "the payload field \"events.#{event}.unhandled\" is false"
 end
 
 # Checks the payloadVersion is set correctly in the payload body in the Javascript or regular place
--- a/lib/features/steps/error_reporting_steps.rb
+++ b/lib/features/steps/error_reporting_steps.rb
@@ -31,7 +31,18 @@ end
 #
 # @param event [Int] The event to verify
 Then("event {int} is unhandled") do |event|
-  step "the payload field \"events.#{event}.unhandled\" is true"
+  if read_key_path(Server.current_request[:body], "events.#{event}.session").nil?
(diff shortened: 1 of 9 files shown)


test/fixtures/payload-helpers/Dockerfile
```diff
@@ -0,0 +1,7 @@
+ARG BUILDKITE_BRANCH
+FROM 855461928731.dkr.ecr.us-west-1.amazonaws.com/maze-runner:${BUILDKITE_BRANCH}-ci
```
fractalwrench: Should this (and a few other places) now be `$BRANCH_NAME` to fix the issue where branches with forward slashes don't pass?

lib/features/steps/error_reporting_steps.rb
```diff
@@ -23,6 +23,96 @@
...
+# Verifies that an event is correct for an unhandled error
+# This checks various elements of the payload matching an unhandled error including:
+#    The unhandled flag
+#    Any attached session information
+#    Severity
+#
+# @param event [Int] The event to verify
+# @param severity [String] An expected severity different to the default "error"
+Then("event {int} is unhandled with the severity {string}") do |event, severity|
+  if read_key_path(Server.current_request[:body], "events.#{event}.session").nil?
```
fractalwrench: It seems like these assertions are shared between all the added steps. Creating some private function and calling it from within each step would make it easier to make a change in one place, e.g.:

```
verify_error(boolean unhandled, string severity)
```

test/fixtures/payload-helpers/features/compare_requests_json_fixtures.feature
```diff
@@ -0,0 +1,47 @@
...
+        Then I wait to receive a request
+        And the request is valid for the session reporting API version "1.0" for the "Maze-runner" notifier
+
+    Scenario: The request body is correct for an unhandled payload
+        When I send a "unhandled"-type request
+        Then I wait to receive a request
+        And event 0 is unhandled
+
+    Scenario: The request body is correct for an unhandled payload with session data
+        When I send an "unhandled session"-type request
```
fractalwrench: Minor nit, but it may be worth tweaking the name as it sounds like the session has the unhandled state rather than the error

test/fixtures/payload-helpers/Dockerfile
Cawllec: Yes, I'll have to merge up.

PR #72, opened 2019-11-14
V2 ghpages fix

Cawllec:
## Goal

- Fix the markdown issues present in the current docs by using the Markdown and Redcarpet libraries
- Add a separate (and much quicker) docker image for the docs generation and publishing
- Fix the broken build & publish task to automate publishing in CI

## On git risks

The changed methodology of the preparation and publishing process of the docs to the `gh-pages` branch of this repo involves the creation of a `.git-credentials` file featuring a token to allow the branch to be pushed once updated.
A corresponding token has been created, and stored in this repos secret environment file.  It is scoped only to public repos, and shouldn't allow access to anything outside of general access.


README.md
```diff
@@ -121,19 +120,26 @@ After do |scenario|
     # Can be used to do specific cleanup if a scenario fails
   end
 end
- ```
+```
+
+### Step definition links
 
-### Step reference
+- [Request assertion steps](/maze-runner/requirements/step_transformers.html#step_definition64-stepdefinition)
```
tomlongridge: Are these links meant to work from the github.com/bugsnag/maze-runner page? I'm not sure they will 🤔

Also, are they likely to change if new steps are added?

README.md
```diff
@@ -301,7 +244,7 @@ By default the logger is set to report `WARN` level events or higher, but will l
 
 ## Contributing
 
-If steps would be useful for different projects running the maze, add the to
+If steps would be useful for different projects using maze-runner, add them to
```
tomlongridge:
```suggestion
If steps would be useful for different projects using maze-runner, to add them to
```

Rakefile
```diff
@@ -19,27 +36,30 @@ end
...
-      `git pull`
+      run_command "git init"
+      run_command "git remote add origin https://github.com/bugsnag/maze-runner.git"
+      run_command "git fetch"
+      run_command "git checkout gh-pages"
+      run_command "git pull"
+      run_command "echo https://#{ENV["DOCS_PUSH_TOKEN"]}:x-oauth-basic@github.com >> ~/.git-credentials"
+      run_command "git config credential.helper 'store'"
+      run_command "git config --global user.email 'notifiers@bugsnag.com'"
+      run_command "git config --global user.name 'Bugsnag notifiers'"
```
tomlongridge: Can this go in the Dockerfile.docs?

README.md
Cawllec: I think the current change is correct?  As in, "if you want to add steps to maze runner, add them to `file`"

tomlongridge: Yep, sorry I misparsed that!

README.md
Cawllec: For the first point, no, and I'll remedy that by separating the two (i.e. Readme for github and front-page for docs).

For the second point, yes, and without amending the plugin template to add groups there doesn't seem to be much we can do other than leave them out entirely.  It's not great.

Rakefile
Cawllec: It can, only potential issue would be the push token being uploaded in plain-text to our AWS ECR repository.  Probably not a big deal, but it's a potential risk.

Cawllec: As a note, AWS S3 already contains the token in plain text which is probably an equivalent risk, so this is likely a non-issue.

tomlongridge: I don't have a strong view either way, it just seemed like a one-time setup step rather than part of the publish steps.

README.md
tomlongridge: Sure - it'll do for now.

Rakefile
Cawllec: Realised this is a non issue as this image isn't uploaded to ECR.  The token doesn't leave CI.

PR #73, opened 2019-11-26
V2 breadcrumb step updates

Cawllec:
## Goal
This change moves the existing breadcrumb steps from error_reporting_steps.rb to their own file, breadcrumb_steps.rb.  In addition, it adds two new steps to enable easier testing of breadcrumbs:

- A step that tests that a breadcrumb with a particular type is not present within a payload, useful for testing both breadcrumb filtering and automatic breadcrumb configuration in notifiers such as JS
- A step that takes a JSON fixture and searches for a matching breadcrumb within a payload. This is useful for verifying the types of information and fields present in a breadcrumb without having to know the index of said breadcrumb.

In addition, this PR updates the step links to fit with the new docs generated by this PR, and adds some testing around the breadcrumb steps.

### DIFF
--- a/test/fixtures/payload-helpers/features/breadcrumb_json_fixtures.feature
+++ b/test/fixtures/payload-helpers/features/breadcrumb_json_fixtures.feature
@@ -13,7 +13,7 @@ Feature: Breadcrumb helper steps
     Scenario: The payload does not contain a type of breadcrumb
         When I send a "breadcrumbs"-type request
         Then I wait to receive a request
-        And the event does not have a "request" type breadcrumb
+        And the event does not have a "request" breadcrumb
 
     Scenario: The payload has a breadcrumb which matches a JSON fixture
         When I send a "breadcrumbs"-type request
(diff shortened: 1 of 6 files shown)


test/fixtures/payload-helpers/features/breadcrumb_json_fixtures.feature
```diff
@@ -0,0 +1,21 @@
...
+
+    Scenario: The payload does not contain a type of breadcrumb
+        When I send a "breadcrumbs"-type request
+        Then I wait to receive a request
+        And the event does not have a "request" breadcrumb
+
+    Scenario: The payload has a breadcrumb which matches a JSON fixture
+        When I send a "breadcrumbs"-type request
+        Then I wait to receive a request
+        And the event contains a breadcrumb matching the JSON fixture in "features/fixtures/breadcrumb_match.json"
```
tomlongridge: This should probably include some false-positive tests to make sure it doesn't match some scenarios too.

PR #75, opened 2020-01-28
Add BrowserStack project, builds, and names

Cawllec:
## Goal

This change is intended to enable developers to more easily find the BrowserStack instance connected to a specific buildkite step and job.

This accomplishes this using 3 previously unused BrowserStack appium capabilities:
- `project` which is searchable within BrowserStack.  This is most equivalent to a buildkite pipeline, and will be set to the pipeline name in buildkite, or "local" if buildkite is not available.
- `build` which is the top level unit sortable in BrowserStack.  Each one represents a buildkite job.  This will be set to a combination of the branch name and build number, ensuring each job has a unique build id, and when run locally will be assigned a random UUID.
- `name` which is the lowest level unit in BrowserStack, and equivalent to a step in buildkite.  Each one will have a name derived from a combination of the targeted device, the test step id, and the current retry value.  This should ensure each one is unique inside of the BrowserStack dashboard. When run locally, this will be assigned the device type.

In addition, each of these values will be logged at the `warn` level before the tests start.  This means that they will always be present at the start of the tests, and enable the user to quickly locate the BrowserStack instance without having to work out the specific build and name themselves.

## Caveats

This will require a small amount of work to allow the environment variables from Buildkite to enter into the maze-runner test container.  This will involve adding the following environment variables to the appropriate service definitions in the docker compose files:
- BUILDKITE_BRANCH
- BUILDKITE_PIPELINE_NAME
- BUILDKITE_RETRY_COUNT
- BUILDKITE
- BUILDKITE_BUILD_NUMBER
- BUILDKITE_STEP_KEY


test/app_automate_driver_test.rb
```diff
@@ -12,7 +12,27 @@ class AppAutomateDriverTest < Test::Unit::TestCase
   TARGET_DEVICE = "ANDROID_9"
   LOCAL_TUNNEL_COMMAND = "/BrowserStackLocal -d start --key #{ACCESS_KEY} --local-identifier #{LOCAL_ID} --force-local"
 
+  def setup
```
twometresteve: Do we really need unit tests for the maze runner code at all?  It feels like we're testing the tests and that time spent maintaining these tests is nugatory.  If we shipped Maze Runner as a product it might be different, but it's only used for Bugsnag CI purposes and to my mind the test harness is verified by virtue of the number of CI tests being run with it.

lib/features/support/app_automate_driver.rb
```diff
@@ -27,13 +28,27 @@ def initialize(username, access_key, local_id, target_device, app_location, loca
...
+
+    # Sets up identifiers for ease of connecting jobs
+    project_id = ENV['BUILDKITE'] ? ENV['BUILDKITE_PIPELINE_NAME'] : "local"
+    build_id = ENV['BUILDKITE'] ? "#{ENV['BUILDKITE_BRANCH']} #{ENV['BUILDKITE_BUILD_NUMBER']}" : SecureRandom.uuid
+    build_name = "#{@device_type} #{ENV['BUILDKITE_STEP_KEY']} #{ENV['BUILDKITE_RETRY_COUNT']}"
+
+    $logger.warn "Appium driver initialised for:"
+    $logger.warn "    project : #{project_id}"
+    $logger.warn "    build   : #{build_id}"
+    $logger.warn "    name    : #{build_name}"
```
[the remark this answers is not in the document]
Cawllec: Hmm, I was using it due to the fact our logger uses warn as it's base reporting level, however I could change that to notify and use that instead which would be more appropriate.  I didn't want to introduce too much new noise however.

lib/features/support/app_automate_driver.rb
```diff
@@ -27,13 +28,27 @@ def initialize(username, access_key, local_id, target_device, app_location, loca
     @access_key = access_key
     @local_id = local_id
     app_url = upload_app(username, access_key, app_location)
+
+    # Sets up identifiers for ease of connecting jobs
+    project_id = ENV['BUILDKITE'] ? ENV['BUILDKITE_PIPELINE_NAME'] : "local"
+    build_id = ENV['BUILDKITE'] ? "#{ENV['BUILDKITE_BRANCH']} #{ENV['BUILDKITE_BUILD_NUMBER']}" : SecureRandom.uuid
```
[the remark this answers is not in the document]
Cawllec: Yeah, that's a possibility.  This logic could be broken out into a separate method which could be a bit more discerning as to what's available.

test/app_automate_driver_test.rb
Cawllec: This is true to an extent, however especially when interacting with BrowserStack through Appium it's useful to be able to verify that everything is working as expected in the majority of circumstances.
At the moment our integration tests explicitly testing Maze-runner in use are fairly light, but we could move to focus on them rather than unit tests.

Cawllec: It's also good to rule out any unexpected behaviour that may occur only rarely in maze-runner use.

twometresteve: Ok, if we feel we're getting value out of them then that seems perfectly sensible and I don't have a problem with targeted tests.  I just wouldn't want to slavishly write unit tests for everything "because we do".

lib/features/support/app_automate_driver.rb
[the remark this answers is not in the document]
twometresteve: Ok, hadn't realised that warn was the base.    If it can configured to allow us to log at info without introducing noise from elsewhere then that would be great, but obviously not worth too much time investing in it.

test/app_automate_driver_test.rb
Cawllec: Agreed, I don't think unit testing all logic is necessary in most cases.  We probably only need "complete" unit testing around tricky bits of functionality which connects to external services, with light testing around the rest.

lib/features/support/app_automate_driver.rb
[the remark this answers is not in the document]
Cawllec: I think I'll address this in a quick PR setting the default to `INFO` but allowing a `quiet` option that sets it to `ERROR`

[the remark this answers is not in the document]
Cawllec: PR here: #77

lib/features/support/app_automate_driver.rb
```diff
@@ -89,6 +100,39 @@ def send_keys_to_element(element_id, text)
     find_element(@element_locator, element_id).send_keys(text)
   end
 
+  # Determines and returns sensible project, build, and name capabilities
+  #
+  # @param target_device [String] the device in the device list being targeted
+  #
+  # @return [Hash] A hash containing the 'project', 'build', and 'name' capabilities
+  def project_name_capabilities(target_device)
```
Cawllec: @twometresteve can you have a quick look at this method to make sure you're happy with how it's generating the build names?

twometresteve: Looks fine as a piece of code.  I just don't have good enough understand of how it will manifest itself within the build and if space separated strings will be an issue (or example), but happy for it to be merged and corrected later if necessary.

PR #78, opened 2020-02-06
Add retries and fast-failure to maze-runner

Cawllec:
## Goal

Adds 2 automatic retries when a step fails.  These retries work immediately without running to the end of the tests.
In addition, ensures the tests will exit on the first test failure.  These conditions work in conjunction, so the retries will occur immediately after a failing scenario, with the tests exiting if the failure occurs 3 times.

## Testing

I've verified that this works manually, but the nature of the `fail-fast` tests mean that a solid test will potentially be difficult.  As this is an in-built feature of cucumber, I don't believe it is necessary to add exhaustive tests.

I can however add an integration test for the retries using a stateful fixture that will only pass after a number of retries.

### DIFF
--- a/dockerfiles/Dockerfile.ci
+++ b/dockerfiles/Dockerfile.ci
@@ -33,4 +33,4 @@ COPY . ./
 RUN bundle install
 
 FROM ci as cli
-ENTRYPOINT ["bundle", "exec", "maze-runner"]
+ENTRYPOINT ["bundle", "exec", "maze-runner", "--fail-fast", "--retry", "2"]


snmaynard:
The one caveat of this is that we should be moving towards retries not being required. They can mask real issues.

twometresteve:
Peer review combined with cursory local verification that standard Cucumber options are passed in and used by Maze Runner is sufficient testing for this change.  We don't need to write tests for the behaviour of third party libraries that we use, per se.

Cawllec:
Regarding the issue with flakes, I suggest we trial this for a couple of weeks and then review the changes in a retro, to make sure we're happy with what has resulted.

PR #79, opened 2020-02-17
Add android 8.1 and 10

twometresteve:
## Goal

Provide the ability to run Maze Runner tests against Android 8.1 and 10.0.

## Design

Minor versions added to naming where it was was .0, as we wanted to provide support for both 8.0 and 8.1 (not so worried about 7.0).

### Added

ANDROID_4_4
ANDROID_7_1
ANDROID_8_0
ANDROID_8_1
ANDROID_10_0
IOS_13

### Changed

Some Rubocop violations ironed out, suppressing one (use of symbols for hashes) to avoid potentially breaking existing uses of the hash.

## Tests

Existing CI tests are sufficient in the main, but the Android Pipeline has been updated to use this and will be checked prior to merging this PR.


twometresteve:
I didn't mention is in the intro, but I also ironed out a few Rubocop violations as part of the deal, suppressing one to avoid potentially breaking existing uses.

Cawllec:
Is it possible to make this only an additive change?  This will cause CI to fail on Android and Expo pipelines.  Due to the messy nature of how changes are merged into those notifiers the fix may not be instantly available to all currently in-progress branches.

Once proper versioning is added to releases this will be a lot easier to merge without breaking pipelines.

twometresteve:
> Is it possible to make this only an additive change? This will cause CI to fail on Android and Expo pipelines. Due to the messy nature of how changes are merged into those notifiers the fix may not be instantly available to all currently in-progress branches.
> 
> Once proper versioning is added to releases this will be a lot easier to merge without breaking pipelines.

twometresteve:
> Is it possible to make this only an additive change? This will cause CI to fail on Android and Expo pipelines. Due to the messy nature of how changes are merged into those notifiers the fix may not be instantly available to all currently in-progress branches.
> 
> Once proper versioning is added to releases this will be a lot easier to merge without breaking pipelines.

Good thinking, I hadn't considered all the different branches in flight, or Expo at all.  I've now refactored devices.rb a little to retain all the old names without leading to lots of duplication.

lib/features/support/capabilities/devices.rb
```diff
@@ -1,82 +1,100 @@
...
-          'os_version' => '10'
-      }
+      'ANDROID_10_0' => ANDROID_10_0,
+      'ANDROID_9_0' => ANDROID_9_0,
+      'ANDROID_8_1' => ANDROID_8_1,
+      'ANDROID_8_0' => ANDROID_8_0,
+      'ANDROID_7_1' => ANDROID_7_1,
+      'ANDROID_6_0' => ANDROID_6_0,
+      'ANDROID_4_4' => ANDROID_4_4,
+      'IOS_12' => IOS_12,
```
Cawllec: Now that BrowserStack supports iOS 13 it may be worth adding at this point for future usage.

twometresteve: Agreed, done.

PR #69, opened 2019-10-15
Update Gemfile.lock

friederbluemle:
Update versions in Gemfile.lock to the latest.

PR #82, opened 2020-02-25
Delete Gemfile and Gemfile.lock

friederbluemle:
Follow up to #69


snmaynard:
The Gemfile is required, we just don't need to update the versions in it. Thanks for the contribution!

PR #88, opened 2020-03-10
Added check for number of breadcrumbs

tomlongridge:
## Goal

Added: "the event has `x` breadcrumbs"


Cawllec:
I'm not opposed to this, but an alternative would be to use:
https://github.com/bugsnag/maze-runner/blob/master/lib/features/steps/request_assertion_steps.rb#L154
as in:
```
the payload field "events.0.breadcrumbs" is an array with #NUM elements
```

tomlongridge:
I added the test locally to Android in the end so that it covers v1 and v2.

PR #91, opened 2020-03-30
V2 updates

snmaynard:
Two updates in this PR.
1) Update test-unit to prevent a warning being logged when running mazerunner
2) Remove the custom entrypoint that retries failures. If we want to do this it should be in the ci pipeline in the repo under test, as it is way more discoverable there. The merits of having it or not are a separate discussion that I believe has already happened, but having it baked in to the maze runner images means its hard to know that is why it is not running the full suite and its hard to disable it locally to opt out (you basically have to know you have to override the entrypoint).

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -11,12 +11,12 @@ PATH
       os (~> 1.0.0)
       rake (~> 12.3.3)
       selenium-webdriver (~> 3.11)
-      test-unit (~> 3.2.0)
+      test-unit (~> 3.3.0)
 
 GEM
   remote: https://rubygems.org/
   specs:
-    activesupport (6.0.2.1)
+    activesupport (6.0.2.2)
       concurrent-ruby (~> 1.0, >= 1.0.2)
       i18n (>= 0.7, < 2)
       minitest (~> 5.1)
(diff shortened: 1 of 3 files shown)


twometresteve:
I take your point about the retries being hidden, but it was discussed as a team, well understood by developers and providing real benefit as a short term approach to flaking tests.  Without putting the more architectural pleasing solution in place at the same time, I see this is a backward step.

snmaynard:
It is confusing for future travellers, being explicit is just as easy and is clearer so I dont see how this is a backwards step.

snmaynard:
I've raised the prs against the other repos to demonstrate how I think it should be done.

I use `docker run maze-runner` to have a local maze-runner cli and  having options hidden where I have to override the entrypoint is confusing. I even ran `docker run maze-runner --help` and saw the fail-fast option but theres no way to turn that off. This image should act like vanilla cucumber imo with some auto requires built in.

twometresteve:
> I've raised the prs against the other repos to demonstrate how I think it should be done.
> 
> I use `docker run maze-runner` to have a local maze-runner cli and having options hidden where I have to override the entrypoint is confusing. I even ran `docker run maze-runner --help` and saw the fail-fast option but theres no way to turn that off. This image should act like vanilla cucumber imo with some auto requires built in.

Ok, having now talked to you in person I can see the bigger picture now.  I'm happy with this in principle now and will approve/merge once PRs are in on other repos where appropriate so that the owners of notifiers that want the retry behaviour don't lose it and get grumpy.

PR #92, opened 2020-04-02
Add HTTP version step

Cawllec:
## Goal
Adds a step for interrogating the HTTP_VERSION of a given request

## Testing
Verified manually.
Added test fixture testing this step specifically.
To run:
`git checkout v1-add-http-versions`
`cd test/fixtures/request-test`
`bundle install`
`bundle exec maze-runner`


test/fixtures/request-test/features/steps/scripting_steps.rb
```diff
@@ -0,0 +1,7 @@
+When(/^I send an HTTP 1.1 request$/) do
```
tomlongridge, line 1: Suggest we should have a test for non-1.1 also?

PR #95, opened 2020-05-15
Ensure `nil?` is called before `empty?` on potentially `nil` variables

Cawllec:
## Goal

When testing some functionality that relies on the `BUILDKITE` environment variable it was discovered that the `empty?` and `nil?` guards on this code not only don't work correctly but cause an error to be thrown if the either of the variables in question are `nil`.

This change swaps the order of the `nil?` and `empty?` guards so that if the variable is `nil` the `empty?` call will not occur on the `nil` object, and the error won't be thrown.
It will allow us to define and test behaviour that depends on the presence of the `BUILDKITE` variable, although it should have minimal visible effect on the actual CI runs.


twometresteve:
Just remembered, please add an entry to `CHANGELOG.md`.

PR #97, opened 2020-05-26
Improve diagnostics

twometresteve:
## Goal

The main objective of this PR is to provide better diagnostics whenever a scenario fails, but logging out all requests that were received.  This is an improvement on what we currently have, where we often have no useful output to analyse. 

Some peripheral changes have also been made to resolve Rubucop warnings - expect this to happen across forthcoming PRs as we visit different areas.

## Design

This is intended to be a quick win, but it can make the log noisy.  The plan will be to move the logging into files that get attached as downloadable artifacts in Buildkite steps.

## Changeset

New support file to manage diagnostic output. Warnings resolved in a couple of other files in the same area.

## Tests

I've run the Cocoa tests against this branch locally and on CI to check behaviour and will encourage the use of `master` for ongoing work before we make a new release.  In general, though, we may want to look at enhancing the maze-runner pipeline to include more tests for specific notifiers so that we get sufficient confidence from just a successful CI run.

### DIFF
--- a/lib/features/support/diagnostics.rb
+++ b/lib/features/support/diagnostics.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+require 'json'
+
+After do |scenario|
+  if scenario.failed?
+    if Server.stored_requests.empty?
+      $logger.info 'No requests received'
+    else
+      $logger.info 'The following requests were received:'
+      Server.stored_requests.each_with_index do |request, number|
+        json = JSON.pretty_generate request
+        $logger.info "Request #{number}: \n#{json}"
+      end
+    end
+  end
+end
+
+
--- a/lib/features/support/logger.rb
+++ b/lib/features/support/logger.rb
@@ -1,9 +1,20 @@
+# frozen_string_literal: true
+
 require 'logger'
 
-if ENV['VERBOSE'] || ENV['DEBUG']
-  $logger = Logger.new(STDOUT, level: Logger::DEBUG)
-elsif ENV['QUIET']
-  $logger = Logger.new(STDOUT, level: Logger::ERROR)
-else
-  $logger = Logger.new(STDOUT, level: Logger::INFO)
+# A logger, with level configured according to the environment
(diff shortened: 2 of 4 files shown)


twometresteve:
Apologies - just force pushed having realised that I had left a messy commit history.

lib/features/support/diagnostics.rb
```diff
@@ -0,0 +1,19 @@
...
+
+After do |scenario|
+  if scenario.failed?
+    if Server.stored_requests.empty?
+      $logger.info 'No requests received'
+    else
+      $logger.info 'The following requests were received:'
+      Server.stored_requests.each_with_index do |request, number|
+        json = JSON.pretty_generate request
+        $logger.info "Request #{number}: \n#{json}"
```
imjoehaines, line 13: It doesn't really matter, but it might be nice for this to start at "Request 1" rather than "Request 0"

```suggestion
      Server.stored_requests.each.with_index(1) do |request, number|
        json = JSON.pretty_generate request
        $logger.info "Request #{number}: \n#{json}"
```

lib/features/support/logger.rb
```diff
@@ -1,9 +1,20 @@
...
+    elsif ENV['QUIET']
+      super(STDOUT, level: Logger::ERROR)
+    else
+      super(STDOUT, level: Logger::INFO)
+    end
+    self.datetime_format = '%Y-%m-%d %H:%M:%S'
+  end
+end
+
+# TODO Added for backward compatibility, but remove in time
```
Cawllec, line 19: If this global variable is going to be removed in time, is it worth showing the new way of doing it in the above `diagnostics.rb` addition as a reference for later changes?

twometresteve: That's a good idea and to be honest I hadn't yet worked out the right way to do it in Ruby.  A singleton seems reasonable, though, so have now update it.  I would have simply updated the logger across the board if it were not for a unit test that mocks it - I didn't want to stretch the scope of this PR too much.

PR #103, opened 2020-07-08
Add a step that matches different values for different BrowserStack OS's

Cawllec:
## Goal

This adds a verification step that checks for different values based on the value of the `AppAutomatorDriver`s `os` capability.

This has four outcomes:
1. `$driver` isn't present as this isn't a BrowserStack app automation test, the step will automatically fail
2. The `os` value isn't present in the table, the step will automatically fail
3. & 4. Pass and fail depending on if the expected value matches the payload value

## Usage

```
Then the event "exceptions.0.errorClass" matches the correct platform value:
  | android | Java.lang.RuntimeException |
  | ios         | NSException                           |
```

## Tests

Added steps to the BrowserStack app feature for a positive outcome, tested negative outcomes manually

### DIFF
--- a/lib/features/steps/app_automator_steps.rb
+++ b/lib/features/steps/app_automator_steps.rb
@@ -33,3 +33,22 @@ end
 When("I send the keys {string} to the element {string}") do |keys, element_id|
   $driver.send_keys_to_element(element_id, keys)
 end
+
+# Tests that the given payload value is correct for the target BrowserStack platform.
+# If the step is invoked when a remote BrowserStack device is not in use this step will fail.
+#
+# The DataTable used for this step should have `ios` and `android` in the same row as their expected value:
+#   | android | Java.lang.RuntimeException |
+#   | ios     | NSException                |
+#
+# @step_input field_path [String] The field to test
+# @step_input platform_values [DataTable] A table of acceptable values for each platform
+Then("the event {string} matches the correct platform value:") do |field_path, platform_values|
+  if !defined?($driver) || $driver.nil?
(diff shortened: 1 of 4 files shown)


lib/features/steps/app_automator_steps.rb
```diff
@@ -32,4 +32,23 @@
...
+#
+# The DataTable used for this step should have `ios` and `android` in the same row as their expected value:
+#   | android | Java.lang.RuntimeException |
+#   | ios     | NSException                |
+#
+# @step_input field_path [String] The field to test
+# @step_input platform_values [DataTable] A table of acceptable values for each platform
+Then("the event {string} matches the correct platform value:") do |field_path, platform_values|
+  if !defined?($driver) || $driver.nil?
+    fail("This step should only be used if the AppAutomateDriver is present")
```
twometresteve, line 48: Would be good to have a conversation about Rubocop warning at some point, which I've been trying to resolve generally.  A conversation for another day, though, as I imagine it would involved aligning dev environments to some extent (and what you've done is consistent with the existing code).

Cawllec:
Before merging this (and looking at the Rubocop issue), do we want to record step changes/additions in the changelog, or just functional changes?

twometresteve:
> Before merging this (and looking at the Rubocop issue), do we want to record step changes/additions in the changelog, or just functional changes?

Probably best to but a note in the changelog and reference the PR.  If you rebase on master now you'll see what I did for my recent changes.  Don't worry about the Rubocop warnings for now, it's a conversation for another day.

PR #105, opened 2020-07-09
Allow timeout to be set when waiting for an element to be present

twometresteve:
## Goal

Increase default timeout for an element to be aware to 30s, for the cases where an app may be slower to load.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Enhancements
 
 - Add ability to clear text from an element.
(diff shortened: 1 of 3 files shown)


twometresteve:
> Would it be possible to add a new step that would apply this instead of changing the default? We have some control over the feature/scenario order, so ideally we would only have a longer timeout on the first attempt.

You're right, changing it for everything is far from ideal.  I've added a new step now to provide more flexibility for the steps/scenarios that make use it.

PR #109, opened 2020-07-13
Update platform comparison step to allow step skipping via keywords

Cawllec:
## Goal

Adds the ability to skip a step on a platform (useful for parameters such as `appVersionCode` which are Android specific).
This is accomplished by using the `@skip` keyword/tag inside of the value table, e.g.:
```
  And the event "app.appVersionCode" matches the correct platform value:
    | android | 1.0       |
    | ios         | @skip |
```

Tested by adding the step to the BrowserStackApp test fixture.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# TBA
+
+## Enhancements
+
+- Update platform comparison step to allow skipping via keywords
+
 # 2.2.1 - 2020/07/10
 
 ## Fixes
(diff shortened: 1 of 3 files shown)


twometresteve:
Is the idea to explicitly skip checks, rather than simply omit the row altogether, to guard against accidentally missing a check?  Will steps fail currently is there is no row for the relevant platform?

Cawllec:
Currently they fail if the platform isn't present.  Like you say, this will allow us to explicitly skip a step, saying "it doesn't occur on this platform"

twometresteve:
> Currently they fail if the platform isn't present. Like you say, this will allow us to explicitly skip a step, saying "it doesn't occur on this platform"

Ok, I'm happy with that.

PR #110, opened 2020-07-15
Typed platform checks

Cawllec:
## Goal

Updates the previously added platform checks to specify whether the target value should be a String, Boolean, or Numeric.

This is required for:
1. If skipping certain checks such as `app.codeBundleId` on iOS the step needs to be able to test for any of the basic data types otherwise the Android checks will be invalidated.
2. The original step only used String checking, this allows us to check boolean and numeric values such as appVersionCodes or unhandled/handled error counts.

Alternatives:
The main alternative would be a more structured table and a single step.
This step/table would look something like:
```
Then the event "some variable thing" matches the correct platform value:
  | platform | type    | value                  |
  | android  | bool    | true                    |
  | ios          | @skip |                           |
```

Unfortunately this introduces complexities such as having to parse the type, and be more difficult to parse visually.

## Future work

It would be good to refactor this into a separate class (which can then have other platform specific comparisons added to it), which would allow us to streamline messaging and shared logic.

## Tests

Updated the metadata in the `Browserstack app` test fixture to run each variant of the different scenarios, including `@skip`s.

## When merging

Requires Squash and Merge

### DIFF
--- a/lib/features/steps/app_automator_steps.rb
+++ b/lib/features/steps/app_automator_steps.rb
@@ -53,17 +53,18 @@ When("I send the keys {string} to the element {string}") do |keys, element_id|
 end
 
 # Tests that the given payload value is correct for the target BrowserStack platform.
+# This step will assume the expected and payload values are strings.
 # If the step is invoked when a remote BrowserStack device is not in use this step will fail.
 #
 # The DataTable used for this step should have `ios` and `android` in the same row as their expected value:
 #   | android | Java.lang.RuntimeException |
 #   | ios     | NSException                |
 #
-# If the expected value is set to "skip", the check should be skipped.
+# If the expected value is set to "@skip", the check should be skipped.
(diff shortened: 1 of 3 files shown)


Cawllec:
I think those are both reasonable suggestions - shouldn't take too long to implement.  I'll add them in and update the cocoa test fixture.

Cawllec:
This could probably also do with Unit tests on the methods I added, I'll look at those ASAP

bengourley:
I just spotted one more thing by looking at the react native tests … I think this should read `number` not `numeric`:

```diff
- And the event "app.versionCode" equals the platform-dependent numeric
+ And the event "app.versionCode" equals the platform-dependent number
```

Cawllec:
Any particular reason? I preferred Numeric as it's a Ruby type, but it doesn't directly map to that

PR #111, opened 2020-07-20
Update timestamp regex to allow +0000 syntax

Cawllec:
## Goal

This allows us to support timestamps that don't use the `Z` notation for timezones, and use difference from UTC time, such as:
`2020-07-20T13:40:56+0000`

### DIFF
--- a/lib/features/support/env.rb
+++ b/lib/features/support/env.rb
@@ -4,4 +4,4 @@ require 'securerandom'
 $api_key = SecureRandom.hex(16).tr('+/=', 'xyz')
 
 # A regex providing the pattern expected from timestamps
-TIMESTAMP_REGEX = /^\d{4}\-\d{2}\-\d{2}T\d{2}:\d{2}:[\d\.]+Z?$/
+TIMESTAMP_REGEX = /^\d{4}\-\d{2}\-\d{2}T\d{2}:\d{2}:[\d\.]+(\+\d+)?Z?$/


lib/features/support/env.rb
```diff
@@ -4,4 +4,4 @@
 $api_key = SecureRandom.hex(16).tr('+/=', 'xyz')
 
 # A regex providing the pattern expected from timestamps
-TIMESTAMP_REGEX = /^\d{4}\-\d{2}\-\d{2}T\d{2}:\d{2}:[\d\.]+Z?$/
\ No newline at end of file
+TIMESTAMP_REGEX = /^\d{4}\-\d{2}\-\d{2}T\d{2}:\d{2}:[\d\.]+(\+\d+)?Z?$/
```
twometresteve, line 7: I thought the standard was one of:
- Z
- "+hh:mm" 
- "-hh:mm"  
I.e. with colons and allowing for +/-, but you're seeing other examples?
There's always a danger of over-engineering a regex, but who about something like the following (which Regulex does a great job of visualizing).
```
(?:(?:[\+\-]\d{2}:?\d{2})|Z)?
```

twometresteve, line 7: Adding some form of test here would be good and also remember to add a changelog entry.

Cawllec: Hmm, good point.  I guess this is a non-standard timestamp, it's originating from Cocoa but wasn't brought up during that UAT.

This is from the `device.time` parameter.  @fractalwrench do you know if we do anything to format timestamps such as these on Cocoa?

tomlongridge: Hah - in fact, it's all about to become easy! bugsnag/bugsnag-cocoa#758! It will then be Zulu time for all fields created - hence matching Android.

twometresteve: If that's the case then I would suggest leaving MR alone?

Cawllec: Yep, lets close this and I'll add a subtask to change the regex change I've made in the devices fixture back.

Cawllec:
A fix is coming in for Cocoa to Z notation all our timestamps, so this will no-longer be required.

PR #116, opened 2020-08-24
Reset driver

twometresteve:
## Goal

Restart the Appium driver if an exception is thrown when calling `reset` or `wait_for_element`.

## Design

It would have been nice to apply the restart behaviour more generically and in a centralized fashion, but that would have involved some up-front refactoring of the class.  I'm not opposed to do that longer term, but this is a relatively straightforward change with immediate benefit.

## Changeset

In addition to the core goal, I have also:
- Changed the tests we run to cover Android 6 and 10.  For the time they take, we could also choose to cover 7, 8 and 9 - for the purposes of checking that the selected Appium version and other capabilities basically work for each `DEVICE_TYPE`.
- Changed the pipeline so that when releasing, we push as `latest-v2-cli`, paving the way for a move to `v3` at some point (so users can choose to stay on the latest v2, not just the absolute latest).

## Tests

I've been running our Cocoa pipeline against this several time.  Stability of iOS 10 and 11 is much better and is allowing me to flush out other flakes in the pipeline not related to Appium.  Testing will continue against the master branch for a little while whilst continuing to investigate one or two other failures, before creating a new MazeRunner release.


Cawllec (approved):
I've updated a typo in the changelog, otherwise this all seems good.

PR #121, opened 2020-08-26
Consolidate Cucumber hooks into one file

twometresteve:
## Goal

Consolidate all Cucumber hooks into one file, making it much easier to see and maintain what actions are performed at the various points in the test lifecycle.

## Tests

The collapsing log output of our CI proves that the new file has been brought in and that the `Before`/`After` hooks are functioning as desired.


Cawllec (commented):
I'd suggest a name like `hooks` or `scenario_hooks` for the hooks file, rather than `scenario_context`, as that seems too ambiguous.

PR #126, opened 2020-09-03
Separate endpoints for error and session requests

twometresteve:
## Goal

Initial separation of the `session` and `notify` (error) endpoints in the `Server` class.

## Design

This change is definitely not complete (e.g. verification of sessions is currently completely unsupported), but I am trying to make reviews relatively manageable by working iteratively.  This iteration performs the initial separation of a single `requests` list into `sessions` and `errors` (for notify) in the `Server` class and `Servlet` is updated to be parameterised via a constructor, allowing separate instances to be mounted on `/notify` and `/sessions`.

Until I look more closely at the notifier repos, I'm unsure which Cucumber steps will be needed only for errors, which only apply to session and which might apply to both (and potentially others over time).  I've therefore tended to default to making them `error` specific, in the knowledge that I may need to rework/refactor them on a subsequent iteration.

My next iteration will come as a result of starting to run our Android/Cocoa/JavaScript pipelines against this.

## Changeset

Changes revolve around the `server` class, with the introduction of a new abstraction, `RequestList`, that will allow us to have as many different endpoints as we want, all operating in the same way.

I've started an upgrade guide, which of course will continue to evolve and changes are performed.

I have added some TODOs in some areas where I believe there is scope for separating steps into new and smaller files, which will make maintenance easier.  I'll present a PR in due course with just those steps moved for ease of review.

## Tests

Testing of this integration branch will continue over a series of PRs - most notably when notifier pipelines are updated (on branch) to start using it.

### DIFF
--- a/lib/features/support/server.rb
+++ b/lib/features/support/server.rb
@@ -5,6 +5,8 @@ require 'singleton'
 require 'webrick'
 require_relative './servlet'
 require_relative './logger'
+require_relative './request_list'
+require_relative './request_list'
 
 # Receives and stores requests through a WEBrick HTTPServer
 class Server
@@ -15,9 +17,12 @@ class Server
   #   80, 3000, 4000, 5000, 8000, 8080 or 9000-9999 [ from https://stackoverflow.com/a/28678652 ]
   PORT = 9339
 
+  attr_reader :errors
+  attr_reader :sessions
+
   def initialize
-    @all_notify_requests = []
-    @remaining_notify_requests = []
+    @errors = RequestList.new
+    @sessions = RequestList.new
   end
 
   # Whether the server thread is running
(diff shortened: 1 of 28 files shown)


lib/features/support/server.rb
```diff
@@ -1,86 +1,73 @@
...
 # Receives and stores requests through a WEBrick HTTPServer
 class Server
-  class << self
-    # Whether the server thread is running
-    #
-    # @return [Boolean] If the server is running
-    def running?
-      @thread&.alive?
-    end
+  include Singleton
```
Cawllec, line 12: I'm not sure if we need to move to a singleton model.  It doesn't seem to really provide us with anything, but makes accessing the server methods more complex.

twometresteve: Yes I think you're right now we've talk about it IRL.  I'll take another look.

twometresteve: Now reverted to use Eignenclass approach.  Much better.

Cawllec (approved):

PR #127, opened 2020-09-08
Add additional Android devices to choose from

twometresteve:
## Goal

Adds the full range of Android 6.0, 8.0 and 8.1 devices available on BrowaerStack.  The same could be done for other versions, it just hasn't been done at this stage.

## Design

I've created a new function to reduce the repetition in `devices.rb`.  The constant naming convention was agreed up-front and is the device name minus brand, upper-cased and spaces replaced with underscores.

## Changeset

Alternative devices added.

## Tests

The existing CI pipeline is sufficient to prove that I've not broken ANDROID_6_0 through ANDROID_10_0 (to only ones currently in use) and I've informally tested the new entries using a bespoke build of our Android notifier.

### DIFF
--- a/lib/features/support/capabilities/devices.rb
+++ b/lib/features/support/capabilities/devices.rb
@@ -2,68 +2,44 @@
 # noinspection RubyStringKeysInHashInspection
 class Devices
 
+  class << self
+
+    def make_android_hash(device, version)
+      {
+        'device' => device,
+        'os_version' => version,
+        'autoGrantPermissions' => 'true',
+        'platformName' => 'Android',
+        'os' => 'android'
+      }.freeze
+    end
+  end
+
   # The hash of device capabilities, accessible by simple names
-  ANDROID_10_0 = {
-    'autoGrantPermissions' => 'true',
-    'device' => 'Google Pixel 4',
-    'platformName' => 'Android',
-    'os' => 'android',
-    'os_version' => '10.0'
-  }.freeze
-  ANDROID_9_0 = {
-    'autoGrantPermissions' => 'true',
-    'device' => 'Google Pixel 3',
-    'platformName' => 'Android',
-    'os' => 'android',
-    'os_version' => '9.0'
-  }.freeze
-  ANDROID_8_1 = {
-    'autoGrantPermissions' => 'true',
-    'device' => 'Samsung Galaxy Note 9',
-    'platformName' => 'Android',
-    'os' => 'android',
-    'os_version' => '8.1'
-  }.freeze
-  ANDROID_8_0 = {
-    'autoGrantPermissions' => 'true',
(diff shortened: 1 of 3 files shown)


tomlongridge (approved):
Looks fine as-is, but it seems like it wouldn't take much work to repeat for the iOS devices and then just inline the calls to the `make_android/ios_hash` when declaring `DEVICE_HASH`?

twometresteve:
> Looks fine as-is, but it seems like it wouldn't take much work to repeat for the iOS devices and then just inline the calls to the `make_android/ios_hash` when declaring `DEVICE_HASH`?

Now mostly done for iOS devices, whilst also adding Android 11 and iOS 14.  I missed your suggestion of inlining the calls, but will do that next time I'm in the area.

lib/features/support/capabilities/devices.rb
```diff
@@ -2,95 +2,76 @@
...
+  ANDROID_6_0_NEXUS_6 = Devices.make_android_hash 'Google Nexus 6', '6.0'
+  ANDROID_6_0_GALAXY_S7 = Devices.make_android_hash 'Samsung Galaxy S7', '6.0'
+  ANDROID_6_0 = ANDROID_6_0_NEXUS_6
+  ANDROID_5_0 = Devices.make_android_hash 'Google Nexus 6', '5.0'
+  ANDROID_4_4 = Devices.make_android_hash 'Google Nexus 5', '4.4'
+  IOS_14 = make_ios_hash('iPhone 11', '14')
+  IOS_13 = make_ios_hash('iPhone 8', '13')
+  IOS_12 = make_ios_hash('iPhone 8', '12')
+  IOS_11 = make_ios_hash('iPhone 8', '11')
+  IOS_10 = make_ios_hash('iPhone 7', '10')
```
tomlongridge, line 52:
```suggestion
  IOS_14 = Devices.make_ios_hash('iPhone 11', '14')
  IOS_13 = Devices.make_ios_hash('iPhone 8', '13')
  IOS_12 = Devices.make_ios_hash('iPhone 8', '12')
  IOS_11 = Devices.make_ios_hash('iPhone 8', '11')
  IOS_10 = Devices.make_ios_hash('iPhone 7', '10')
```

tomlongridge (approved):

twometresteve:
> > Looks fine as-is, but it seems like it wouldn't take much work to repeat for the iOS devices and then just inline the calls to the `make_android/ios_hash` when declaring `DEVICE_HASH`?
> 
> Now mostly done for iOS devices, whilst also adding Android 11 and iOS 14. I missed your suggestion of inlining the calls, but will do that next time I'm in the area.

Further refactoring now performed to reduce repetition.

tomlongridge (approved):

PR #128, opened 2020-09-09
Add proxy functionality

twometresteve:
## Goal

Provides HTTP/S proxy functionality with and without basic authentication.

## Design

Having determined that the Ruby standard library provides support for HTTP/S proxies, it was the natural choice for implementing the functionality.  The proxy class is more than large enough to warrant its own file, although I have also deliberately avoided touching the existing `Server` class in the knowledge that it is being refactored as part of separating the Notify and Session endpoints.

## Changeset

New class for running a proxy on demand, supporting Cucumber steps and test scenarios.

## Tests

Tested using a new set of "proxy" Cucumber scenarios, each driven my a simple `curl` command.

### DIFF
--- a/lib/features/steps/proxy_steps.rb
+++ b/lib/features/steps/proxy_steps.rb
@@ -27,7 +27,7 @@ end
 # Test the proxy server handled a request for a host.
 #
 # @step_input host [String] Destination host to check
-Then('the proxy handles a request for {string}') do |host|
+Then('the proxy handled a request for {string}') do |host|
   Test::Unit::Assertions.assert_true(Proxy.instance.handled_host?(host),
                                      "The proxy did not handle a request for #{host}")
 end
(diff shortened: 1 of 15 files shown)


lib/features/steps/proxy_steps.rb
```diff
@@ -0,0 +1,35 @@
...
+# Starts an authenticated HTTPS proxy server.
+#
+Then('I start an authenticated https proxy') do
+  Proxy.instance.start :Https, true
+end
+
+# Test the proxy server handled a request for a host.
+#
+# @step_input host [String] Destination host to check
+Then('the proxy handles a request for {string}') do |host|
```
kattrali, line 30:
```suggestion
Then('the proxy handled a request for {string}') do |host|
```

kattrali (approved):
looks good!

lib/features/steps/proxy_steps.rb
twometresteve: That was my instinct too, actually - have now changed as suggested (but it affects other files too).

PR #129, opened 2020-09-10
Fix logging of requests

twometresteve:
## Goal

Centralise logging of requests, allowing non-JSON, non-ASCII or just large requests to be handled without error and more meaningfully.  For example, file upload requests can easily swamp the log output and this approach allow us to put some limits and truncation in place.

## Design

The `LogUtil` class added can be made as sophisticated as we want, but for now the intention is to make logging something that is useful and, in particular, does not cause errors during the execution of a test.

## Changeset

Request logging routines refactored.

## Tests

I've tested the branch using a currently failing Android Gradle Plugin test with DEBUG logging on.  This exercises both the logging of requests as they are received and at the end of failing scenario when Maze Runner outputs the requests that it received.  An extra unit test may be desirable, however, to back up the newly added class.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,20 @@
-# 2.5.0 - 2020/09/10
+# 2.5.0 - 2020/09/11
 
 ## Enhancements
 
-- Fix broken Build API step syntax
-  [#125](https://github.com/bugsnag/maze-runner/pull/125)
 - Range of additional Android 6, 8.0 and 8.1 device options provided, 
   together with Android 11 and iOS 14.
   [#127](https://github.com/bugsnag/maze-runner/pull/127)
 - Ability to run an HTTP/S proxy added
   [#128](https://github.com/bugsnag/maze-runner/pull/128)
+
+# Fixes
+
+- Fix broken Build API step syntax
+  [#125](https://github.com/bugsnag/maze-runner/pull/125)
+- Fix logging of non-JSON requests
+  [#XYZ](https://github.com/bugsnag/maze-runner/pull/xyz)
+
 
 # 2.4.0 - 2020/09/01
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,15 +1,21 @@
...
   [#127](https://github.com/bugsnag/maze-runner/pull/127)
 - Ability to run an HTTP/S proxy added
   [#128](https://github.com/bugsnag/maze-runner/pull/128)
 
+# Fixes
+
+- Fix broken Build API step syntax
+  [#125](https://github.com/bugsnag/maze-runner/pull/125)
+- Fix logging of non-JSON requests
+  [#XYZ](https://github.com/bugsnag/maze-runner/pull/xyz)
```
tomlongridge, line 16:
```suggestion
  [#129](https://github.com/bugsnag/maze-runner/pull/129)
```

lib/features/support/logger.rb
```diff
@@ -18,3 +18,44 @@ def initialize
...
 $logger = MazeLogger.instance
+
+# A collection of logging utilities
+class LogUtil
+  class << self
+    # Logs requests and other data, accounting for things file upload requests that are too big to log meaningfully.
+    #
+    # @param severity [Integer] A Logger::Severity
+    # @param data [Hash] The data to log (currently needs to be a Hash)
+    def log(severity, data)
```
tomlongridge, line 29: Potentially confusing name - I would have expected this to handle a string for normal logging.

```suggestion
    def log_hash(severity, data)
```

twometresteve: Interesting point - I initially called it `log_data` in the expectation that it will evolve to handle more than just hashes.  I changed it to `log`, I think because of my Java experience and method overloading - which I don't think applies in Ruby.  So on reflection I'll change it to `log_hash` as you suggest and change it again later if necessary.  It may only ever log hashes.

tomlongridge (approved):

PR #131, opened 2020-09-13
Cucumber table-based assertions against request sets

twometresteve:
## Goal

Provides the ability to ability to check a set of requests against expected values provided in a Cucumber table.  Expected values in tables can be:
- a simple string to match against
- the special string "null", which is interpreted as `nil`
- a regular expression, by surrounding the regexp string with `/`s - as you would in the Ruby language

## Design

We already had similar logic in the tests for our Cocoa notifier and started to build another similar solution for our Android Gradle Plugin.  This provides a common solution for all notifiers to make use of.

## Changeset

In addition to a new class and unit test, a couple of unit tests have been moved into their own file to avoid adding a `require` for a `steps` file from a unit test (which results in errors).   `helper.rb` is a bit of a dumping ground and not ideal, but we can refine that over time.

## Tests

Functionality verified by unit test.  General approach validated using the in-progress BAGP e2e tests.


Cawllec (approved):
Approving the changes to this point, but going to add a scenario to cover this in e2e tests as well, then request re-review.

fractalwrench (approved):
LGTM pending CI

PR #133, opened 2020-09-24
Separate appium sessions

twometresteve:
## Goal

Adds a new command line option that allows each Cucumber scenario to be run in its own Appium session.

## Design

This provides one option for dealing with underlying stability issues in Appium based test infrastructure.  It's not mandated and disabled by default, as there are potential drawbacks such as additional length of tests (which seems to vary depending on the device/OS version).

## Changeset

New command line option and associated handled.
New `Configuration` class to store options selected for later access.
New `MazeRunner` "binder" class, providing a central point for accessing features of MazeRunner - over time potentially removing the use of slightly messy global variables.

## Tests

I've enabled the option on a few of the pipeline steps, so that we have some coverage of it both on and off.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -50,13 +50,13 @@ steps:
         run: proxy-tests
     command: 'bundle exec bugsnag-maze-runner'
 
-  - label: 'Browserstack app-automate test - Android 6 (separate Appium sessions)'
+  - label: 'Browserstack app-automate test - Android 6'
     env:
       DEVICE_TYPE: ANDROID_6_0
     plugins:
       docker-compose#v3.3.0:
         run: browserstack-app-automate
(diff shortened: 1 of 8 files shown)


kattrali (approved):
The changes look good, though I'd imagine it would be unlikely to be enabled in most cases just because of the increased run times for each test case. Maybe that can be sped up somehow too in the future.

twometresteve:
> The changes look good, though I'd imagine it would be unlikely to be enabled in most cases just because of the increased run times for each test case. Maybe that can be sped up somehow too in the future.

Yes, I think it's one to keep in the back pocket for now.  It may prove more useful and viable in another test environment (and/or as older devices drifts off of support), but it would be a shame to throw it away having done the work.  Thanks for the review effort.

PR #134, opened 2020-09-28
Resilient Appium driver

twometresteve:
## Goal

Provides resilience to flaky Appium sessions by handling exceptions on each method and restarting the driver.

## Design

It didn't seem possible to have a central method for catching all Appium exception, so I've gone for the next best option of adding a class whose single concern is to catch Appium errors of each method and restart the session.

## Changeset

- New `ResilientAppiumDriver` class.
- Addition of `driver` to the central `MazeRunner` class to avoid the need to use global `$driver`.

## Tests

Our Cocoa notifier build suffers the most from flaking Appium session and I have shared internally the results of an extended soak test run against this.  It shows the exception handling code in action and no failed builds as a result.

As always with Maze Runner, tests will also be triggered automatically on our Cocoa, JavaScript and Android notifiers against the new master branch before we perform the next release.

### DIFF
--- a/lib/features/support/resilient_driver.rb
+++ b/lib/features/support/resilient_driver.rb
@@ -3,15 +3,20 @@ require_relative './fast_selenium'
 require_relative './logger'
 
 # Handles Appium driver restarts and retries in the case of failure
-class ResilientDriver
+class ResilientDriver < AppAutomateDriver
 
-  def initialize(driver)
-    @driver = driver
-  end
-
-  # Deliberately no error handling here
-  def start_driver
-    @driver.start_driver
+  # Creates the ResilientDriver
+  #
+  # @param username [String] the BrowserStack username
+  # @param access_key [String] the BrowserStack access key
+  # @param local_id [String] the identifier for the BrowserStackLocal tunnel
+  # @param target_device [String] a key from the Devices array selecting which device capabilities to target
+  # @param app_location [String] the location of the test-app to upload
+  # @param locator [Symbol] the primary locator strategy Appium should use to find elements
+  # @param additional_capabilities [Hash] a hash of additional capabilities to be used in this test run
+  def initialize(username, access_key, local_id, target_device, app_location, locator = :id, additional_capabilities = {})
+    super username, access_key, local_id, target_device, app_location, locator, additional_capabilities
+    MazeRunner.driver = self
   end
 
   # Checks for an element, waiting until it is present or the method times out
@@ -20,7 +25,7 @@ class ResilientDriver
   # @param timeout [Integer] the maximum time to wait for an element to be present in seconds
   # @param retry_if_stale [Boolean] enables the method to retry acquiring the element if a StaleObjectException occurs
   def wait_for_element(element_id, timeout = 15, retry_if_stale = true)
-    @driver.wait_for_element element_id, timeout, retry_if_stale
+    super.wait_for_element element_id, timeout, retry_if_stale
   rescue Selenium::WebDriver::Error::UnknownError, Selenium::WebDriver::Error::WebDriverError
     recover { wait_for_element element_id, timeout, retry_if_stale }
   end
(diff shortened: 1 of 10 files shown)


lib/features/support/resilient_driver.rb
```diff
@@ -0,0 +1,106 @@
+require 'appium_lib'
+require_relative './fast_selenium'
+require_relative './logger'
+
+# Handles Appium driver restarts and retries in the case of failure
+class ResilientAppiumDriver < AppAutomateDriver
```
imjoehaines, line 6: I'm not sure how the `AppAutomateDriver` is really used, but this feels like this is a place where `method_missing` would be useful. It would let us forward all calls onto the `AppAutomateDriver` without having to write replacements for each method individually. Looking at the docs, it seems like that approach would work fine and it'd mean we wouldn't have to add new methods to both `AppAutomateDriver` & `ResilientAppiumDriver`

e.g.

```ruby
class ResilientAppiumDriver
  def initialize
    @driver = AppAutomateDriver.new
  end

  def method_missing(method, *args, &block)
    raise "Method '#{method}' does not exist" unless @driver.respond_to?(method)

    begin
      @driver.send(method, *args, &block)
    rescue
      recover
      method_missing(method, *args, &block)
    end
  end

  def recover
    # ...
  end
end
```

I think it'd also be good to cap the number of retries to some arbitrary limit. I know timeouts will eventually kill it, but I'm not sure it makes sense to retry infinitely until then as it could delay knowing about actual problems — e.g. Travis' timeout is >1 hour, so it's not a trivial amount of time

```ruby
class ResilientAppiumDriver
  def initialize
    @driver = AppAutomateDriver.new
  end

  def method_missing(method, *args, &block)
    raise "Method '#{method}' does not exist" unless @driver.respond_to?(method)

    retries = 0

    until retries >= 5 # or some other arbitrary number
      begin
        @driver.send(method, *args, &block)

        return
      rescue => error
        retries += 1

        recover
      end
    end

    # re-raise the last error, so we don't hide the actual failure
    # it might be better to re-raise the _first_ error instead?
    # I'm not sure how likely it is that they will be different
    raise error
  end

  def recover
    # ...
  end
end
```

twometresteve: Nice, TIL.  And thanks for the code samples to give me a running start.  I'll give it a go and rerun my soak test for the Cocoa repo.
For info, `AppAutomateDriver` is instantiated in the notifier repos (`env.rb`) and opting into this just means changing the class that's used.

lib/features/support/resilient_driver.rb
```diff
@@ -0,0 +1,63 @@
...
+                                    additional_capabilities
+  end
+
+  def respond_to_missing?(method_name, include_private = false)
+    @driver.respond_to_missing? method_name
+  end
+
+  # rubocop: disable MethodMissingSuper
+  def method_missing(method, *args, &block)
+    raise "Method '#{method}' does not exist" unless @driver.respond_to?(method)
```
imjoehaines, line 32: I think something like this will let you get rid of the rubocop suppression; it looks a bit weird though because the `super` will immediately fail

```suggestion
    return super unless @driver.respond_to?(method)
```

lib/features/support/resilient_driver.rb
imjoehaines: Something I forgot about is the docs won't list any methods for `ResilientAppiumDriver` because of `method_missing`

I can't see a way to automatically include all of the methods from `AppAutomateDriver`, but you can [manually document each method if you want](https://rubydoc.info/gems/yard/file/docs/Tags.md#method), or just link to `AppAutomateDriver` with `@see AppAutomateDriver`

twometresteve: I vote `@see AppAutomateDriver`

lib/features/support/resilient_driver.rb
twometresteve: Yep, I like that - I'm comfortable with the wrapping class behaving just like the inner one when called incorrectly.

imjoehaines (approved):

PR #136, opened 2020-09-29
Allow app location to be given by url

twometresteve:
## Goal

Avoids having to upload the same app to BrowserStack more than once.

## Tests

Passing CI and unit tests sufficient to show that the standard case works.  Tested manually locally to verify that a BrowserStack URL can be specified.


Cawllec (approved):
LGTM, this'll be especially useful for reproducing issues.

PR #138, opened 2020-09-29
Fixes to ResilientDriver class

twometresteve:
## Goal

Fixes a few problems in the `ResilientAppiumDriver` class that crept in during initial review when moving to the `missing_method` approach.  

## Design

These problems went unnoticed until subsequent soak testing due to the inner class instance being assigned to `MazeRunner.driver` instead of the resilient wrapper.  Once this initial error became apparent the remaining problems were quickly realised during local testing and by forcing Appium exceptions to be raised randomly.

## Tests

Tested locally with a focus on resilience prior to creating this PR and will be further soak tested subsequently.


lib/features/support/resilient_driver.rb
```diff
@@ -22,6 +24,9 @@ def initialize(username, access_key, local_id, target_device, app_location, loca
                                     app_location,
                                     locator,
                                     additional_capabilities
+
+    # This must appear after creation of @driver as AppAutomateDriver also does this
+    MazeRunner.driver = self
```
imjoehaines, line 29: Is it worth a unit test to check the right driver is assigned when `ResilientAppiumDriver`/`AppAutomateDriver` get created?

imjoehaines (approved):

lib/features/support/resilient_driver.rb
twometresteve: Basic unit test now added - can also be enhanced over time.

PR #139, opened 2020-10-07
Better support for running tests on local devices

twometresteve:
## Goal

Genericize the Maze Runner tool to provide the ability to run Appium-based notifier tests on device farms other than BrowserStack, as well as locally held devices using Appium desktop.

## Design

This change is primarily a refactoring of existing functionality, with additional code written to allow parameters to be passed in on the command line and stored in a configuration class. 

Some improvements to the interface have also been realised, such as the addition of boilerplate Cucumber hooks for creating, starting and stopping the Appium driver no longer being required.

Whilst this change goes most of the way to provide support for running tests on locally held devices, there are some outstanding issues to resolve over the course of forthcoming PRs.  These are noted in code TODOs and more generally in an internal document.

## Changeset

- Refactored classes centred around the AppiumDriver (was AppAutomateDriver).
- Additional command line option processing
- Addition of Configuration class 
- Parameterization of paths etc to allow easy running as a Gem without building a Docker image

## Tests

Unit tests to be written/updated and CI settings changed.  As this is is a breaking change anyway, bumping the major version, we can afford to take a more relaxed approach to release than usual and use the upgrade of selected notifier pipelines to drive further, more targeted, changes.


lib/features/support/appium_driver.rb
```diff
@@ -0,0 +1,155 @@
+require 'appium_lib'
```
twometresteve, line 1: Please note this is a rename from `app_automate_driver.rb`

lib/features/support/browser_stack_utils.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
```
twometresteve, line 1: This class resulted from factoring code out of `app_automate_driver.rb`.

lib/features/support/capabilities/capabilities.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
```
twometresteve, line 1: Again, this largely resulted from factoring code out of `app_automate_driver.rb`, but with more additions.

test/appium_driver_test.rb
```diff
@@ -0,0 +1,247 @@
+# frozen_string_literal: true
```
twometresteve, line 1: Mostly a renaming from `app_automate_driver_test.rb`.

test/fixtures/android-app/build.gradle
```diff
@@ -1,14 +1,14 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
-    ext.kotlin_version = '1.2.21'
+    ext.kotlin_version = '1.2.51'
```
twometresteve, line 4: I could have presented this as a separate PR, to be honest - it allows the test fixture to be build locally on MacOS 10.15.

bin/bugsnag-maze-runner
```diff
@@ -75,12 +122,51 @@ class MazeRunnerEntry
...
+                                    else
+                                      raise "Unknown farm '#{farm}'"
+                                    end
+
+    # Farm specific options
+    if MazeRunner.configuration.farm == :bs
+      raise "option --#{OPTION_USERNAME} must be specified" if options[OPTION_USERNAME].nil?
+      raise "option --#{OPTION_ACCESS_KEY} must be specified" if options[OPTION_ACCESS_KEY].nil?
+      raise "Device type '#{device}' not known on BrowserStack.  Must be one of #{Devices::DEVICE_HASH.keys}" \
+        unless Devices::DEVICE_HASH.keys.include? device_type
```
kattrali, line 146: Rather than wrap the line, I'd break it up for readability.
```suggestion
      unless Devices::DEVICE_HASH.keys.include? device_type
        raise "Device type '#{device}' not known on BrowserStack.  Must be one of #{Devices::DEVICE_HASH.keys}"
      end
```

lib/features/hooks/hooks.rb
```diff
@@ -2,9 +2,51 @@
...
+                                                                   MazeRunner.configuration.apple_team_id,
+                                                                   MazeRunner.configuration.device_id
+  end
+  # Set app location (file or url) in capabilities
+  MazeRunner.configuration.capabilities['app'] = MazeRunner.configuration.app_location
+
+  # Create and start the drive
+  MazeRunner.driver = ResilientAppiumDriver.new(MazeRunner.configuration.appium_server_url,
+                                                MazeRunner.configuration.capabilities)
+  MazeRunner.driver.start_driver unless MazeRunner.configuration.appium_session_isolation
```
kattrali, line 39: pulling config into its own var makes this whole block more concise

```suggestion
  config = MazeRunner.configuration
  case config
  when :bs
    tunnel_id = SecureRandom.uuid
    config.capabilities = Capabilities.for_browser_stack(config.device_type, tunnel_id)

    config.app_location = BrowserStackUtils.upload_app(config.username, 
    												   config.access_key,
                                                       config.app_location)
    BrowserStackUtils.start_local_tunnel(config.bs_local, tunnel_id, config.access_key)
  when :local
    config.capabilities = Capabilities.for_local(config.device_type,
                                                 config.apple_team_id,
                                                 config.device_id)
  end
  # Set app location (file or url) in capabilities
  config.capabilities['app'] = config.app_location

  # Create and start the drive
  MazeRunner.driver = ResilientAppiumDriver.new(config.appium_server_url, config.capabilities)
  MazeRunner.driver.start_driver unless config.appium_session_isolation
```

bin/bugsnag-maze-runner
```diff
@@ -75,12 +122,51 @@ class MazeRunnerEntry
...
+    device_type = MazeRunner.configuration.device_type = options[OPTION_DEVICE]
+    MazeRunner.configuration.farm = if farm.nil?
+                                      :none
+                                    elsif farm == 'bs'
+                                      :bs
+                                    elsif farm == 'local'
+                                      :local
+                                    else
+                                      raise "Unknown farm '#{farm}'"
+                                    end
```
kattrali, line 139: Indenting the whole block instead of just the first line seems unusual style-wise. Could probably use a ruby second opinion. @imjoehaines ?

Some of the instances where many `elseif` statements are used could be made more compact with `case`/`when`

```suggestion
    MazeRunner.configuration.farm = case farm
    when nil then :none
	when 'bs' then :bs
	when 'local' then :local
	else
	  raise "Unknown farm '#{farm}'"
	end
```

bin/bugsnag-maze-runner
```diff
@@ -75,12 +122,51 @@ class MazeRunnerEntry
...
 
+    # Process CL options
     init if options[:init]
-    MazeRunner.configuration.appium_session_isolation = true if options[OPTION_SEPARATE_SESSIONS]
+    MazeRunner.configuration.appium_session_isolation = options[OPTION_SEPARATE_SESSIONS]
+    MazeRunner.configuration.app_location = options[OPTION_APP]
+    farm = options[OPTION_FARM]
+    device_type = MazeRunner.configuration.device_type = options[OPTION_DEVICE]
+    MazeRunner.configuration.farm = if farm.nil?
+                                      :none
```
kattrali, line 132: Should the default value of `farm` be set to `:none` in `Configuration.init`?

kattrali (approved):
👍 lgtm functionally, added a few general style comments for places I had to stop and reread a few times.

bin/bugsnag-maze-runner
imjoehaines: It's a recent-ish Rubocop rule. I'm not a fan personally but I think it's on by default

kattrali: ah, I wondered if it was something like that. thanks!

bin/bugsnag-maze-runner
```diff
@@ -23,13 +33,38 @@ class MazeRunnerEntry
       opt :init, 'Initialises a new Maze Runner project'
 
       opt :version, 'Display Maze Runner and Cucumber versions'
-      opt OPTION_SEPARATE_SESSIONS, 'Start a new Appium session for each scenario', short: :none
+      opt 'separate-sessions', 'Start a new Appium session for each scenario',
```
imjoehaines, line 36: Looks like this got changed from `OPTION_SEPARATE_SESSIONS`

```suggestion
      opt OPTION_SEPARATE_SESSIONS, 'Start a new Appium session for each scenario',
```

bin/bugsnag-maze-runner
```diff
@@ -23,13 +33,38 @@ class MazeRunnerEntry
...
+          short: :none,
+          type: :boolean,
+          default: false
+      opt OPTION_APP, 'The app to be installed and run against', short: :none, type: :string
+      opt OPTION_BS_LOCAL,
+          '(BS only) Path to the BrowserStackLocal binary',
+          short: :none,
+          type: :string,
+          default: '/BrowserStackLocal'
+      opt OPTION_FARM, 'Device farm to use: "bs" (BrowserStack) or "local"', short: :none, type: :string
```
imjoehaines, line 46: I wonder if it'd be simpler to have a switch for BrowserStack and a separate switch for local; not sure how many farms we want to support simultaneously so that might be a bad idea

e.g.

```sh
$ maze-runner --browserstack ...
$ maze-runner --local ...
$ maze-runner ... # equivalent to '--farm none'
```

bin/bugsnag-maze-runner
imjoehaines: There's also [a `has_key?` method](https://ruby-doc.org/core-2.7.1/Hash.html#method-i-has_key-3F) you can use instead of `.keys.include?`

imjoehaines (approved):

bin/bugsnag-maze-runner
twometresteve: Yeah, I just blindly followed what Rubocop told me to do on this one.  I imagine we'll have a discussion about landing on some kind of Rubocop config at some point and stickling to it (there are still plenty of violations at the moment that I'm gradually working through).

lib/features/hooks/hooks.rb
twometresteve: Yep, disappointed I didn't realise that myself.

bin/bugsnag-maze-runner
twometresteve: Yep, not sure what I did there.

bin/bugsnag-maze-runner
twometresteve: Hmm, a simpler command line and it's unlikely we will have more than a few different farms, but I think it would mean a more complex implementation?  Having to interrogate multiple options rather than a simple `case`, adding validation to reject mutually exclusive options being used together....

bin/bugsnag-maze-runner
twometresteve: Line separated.  I got used to Rubocop moaning that I should use favour the single line form and just did it, but apparently Rubocop is smart enough in this case to know that the increased line length would require wrapping.  

Rubocop has also told me to go further and just use `key?`.

bin/bugsnag-maze-runner
twometresteve: It saves a couple of lines, but a `!farm.nil?` has to guard the final `raise`, the reader loses the immediate knowledge that `Configuration.farm` ends up as `:none` and an initializer has to be added to`Configuration`.  
If there was already default configuration in play then I'd probably change it, but on balance I think this is clearer.  Just.

bin/bugsnag-maze-runner
twometresteve: Oh, I have changed it to a `case` now though, much better.

twometresteve:
Thanks for all the review comments and style suggestions.  I've implemented almost all of them and an about to merge based on the two approvals.  If there's anything I missed that you're even slightly passionate about just drop me a line and I'll follow up.

bin/bugsnag-maze-runner
imjoehaines: I was thinking of something along the lines of

```ruby
config.farm = :local if options[OPTION_LOCAL]
config.farm = :bs if options[OPTION_BROWSERSTACK]
```

but yeah, if you want to validate the options it's just more work than the current method :p

PR #140, opened 2020-10-16
Change base Docker image to Ubuntu

twometresteve:
## Goal

Moves the base Docker image to be Ubuntu rather than Alpine based.

## Design

This change was prompted by a few factors:
- The desire to resolve the following errors, logged several times when starting the BrowserStackLocal binary:
```
BrowserStackLocal: /usr/lib/libstdc++.so.6: no version information available (required by /BrowserStackLocal)
/BrowserStackLocal: Relink `/usr/lib/libstdc++.so.6' with `/usr/glibc-compat/lib/librt.so.1' for IFUNC symbol `clock_gettime'
```
- An untidy result from getting the binary running on Alpine in the first place
- [Anecdotal evidence](https://www.reddit.com/r/docker/comments/b6gk1x/why_use_ubuntu_as_base_image_when_alpine_exists/) that trying to use an Alpine image may introduce more problems than the benefit of a smaller image brings.

## Changeset

- Docker files.  
- `Gemfile.lock` saw some updates after rerunning `bundle install` after a conflicting rebase.

## Tests

The error messages have gone from the log and MazeRunner tests still run successfully.  I would be surprised if the change to Ubuntu makes a significant difference to our CI performance, given the compute resources already at our disposal, but we can monitor this as part of the ongoing testing for v3.


dockerfiles/Dockerfile.ci
```diff
@@ -1,31 +1,18 @@
-FROM docker/compose:1.24.0 as compose
-FROM ruby:2.7-alpine as ci
+FROM ubuntu:20.04 as ci
 
-ENV GLIBC_VERSION 2.28-r0
+RUN apt-get update
+RUN apt-get install -y ruby-full apt-utils docker-compose wget unzip bash bundler
```
Cawllec, line 4: Is there a reason to use this over composing the image from both `ubuntu` and `docker/compose` images?

Cawllec: To clarify - installing `docker-compose` this way

twometresteve: It just felt like the natural way of doing it - start from a base image and install what we need.  My assumption is that the original approach was just a workaround for there not being a package available for Alpine and I think this yields an overall cleaner result.

Cawllec (approved):

PR #143, opened 2020-10-19
[PLAT-4981] Add Interactive CLI steps

Cawllec:
## Goal

Adds a way to interact with a terminal, sending commands and verifying output.  This will enable testing of CLIs (such as the upcoming React-native CLI)

## Design

The `InteractiveCLI` instance will act as a single terminal run, storing the parsed STDOUT and STDERR streams to be tested against.
This will be accessible through the `Runner` class, ensuring only one terminal is running at any given time. While this terminal is running it won't be possible to start another one without explicitly stopping it.

## Tests

Covered by basic unit and e2e tests so far, but more will follow in subsequent PRs as part of the ongoing work.

### DIFF
--- a/lib/features/steps/runner_steps.rb
+++ b/lib/features/steps/runner_steps.rb
@@ -63,64 +63,70 @@ When('I wait for {int} second(s)') do |seconds|
   sleep(seconds)
 end
 
-# Starts an interactive terminal
-When('I start a new terminal') do
+# Starts an interactive shell
+When('I start a new shell') do
   Runner.stop_interactive_session
   Runner.get_interactive_session
 end
 
-# Stops currently running interactive terminal
-When('I stop the current terminal') do
+# Stops currently running interactive shell
+When('I stop the current shell') do
   Runner.stop_interactive_session
 end
 
-# Run a command on the terminal
+# Run a command on the shell
 #
-# @step_input command [String] The command to run on the terminal
+# @step_input command [String] The command to run on the shell
 When('I input {string} interactively') do |command|
-  current_terminal = Runner.get_interactive_session
-  success = current_terminal.run_command(command)
+  current_shell = Runner.get_interactive_session
+  success = current_shell.run_command(command)
   assert(success, 'The terminal had already closed')
 end
 
-# Get the current buffer values in the terminal
+# Get the current buffer values in the shell
 #
 # @step_input expected_chars [String] The chars present in current buffer
-Then('the terminal is outputting {string}') do |expected_chars|
-  current_terminal = Runner.get_interactive_session
-  assert_equal(expected_chars, current_terminal.current_buffer)
+Then('the current stdout line is {string}') do |expected_chars|
+  current_shell = Runner.get_interactive_session
+  assert_equal(expected_chars, current_shell.current_buffer)
 end
 
 # Verify a string appears in the stdout logs
 #
 # @step_input expected_line [String] The string present in stdout logs
-Then('the terminal has output {string}') do |expected_line|
-  current_terminal = Runner.get_interactive_session
-  match = current_terminal.parsed_output.any? { |line| line == expected_line }
-  assert(match, "No output lines matched #{expected_line}")
+Then('the shell has output {string} to stdout') do |expected_line|
+  current_shell = Runner.get_interactive_session
+  match = current_shell.stdout_lines.any? { |line| line == expected_line }
+  assert(match, "No output lines from #{current_shell.stdout_lines} matched #{expected_line}")
 end
 
 # Verify a string appears in the stderr logs
 #
 # @step_input expected_err [String] The string present in stderr logs
-Then('the terminal has the error message {string}') do |expected_err|
-  current_terminal = Runner.get_interactive_session
-  match = current_terminal.parsed_errors.any? { |line| line == expected_err }
-  assert(match, "No output lines matched #{expected_err}")
+Then('the shell has output {string} to stderr') do |expected_err|
+  current_shell = Runner.get_interactive_session
+  match = current_shell.stderr_lines.any? { |line| line == expected_err }
+  assert(match, "No output lines from #{current_shell.stderr_lines} matched #{expected_err}")
 end
 
-# Verify the exit code of the terminal
-#
-# @step_input exit_code [Integer] The expected exit code
-Then('the terminal exit code equals {int}') do |exit_code|
-  current_terminal = Runner.get_interactive_session
-  assert_equal(exit_code, current_terminal.last_exit_code)
+# Verify the shell exited successfully (assuming a 0 is a success)
+Then('the shell exited successfully') do
+  current_shell = Runner.get_interactive_session
+  assert_equal(0, current_shell.last_exit_code)
 end
 
-# Assert that the terminal exited with an error code
+# Verify the exit code of the shell
 #
 # @step_input exit_code [Integer] The expected exit code
-Then('the terminal exited with an error code') do
-  current_terminal = Runner.get_interactive_session
-  assert(current_terminal.last_exit_code != 0, "Terminal exited with code 0")
+Then('the shell exit code is {int}') do |exit_code|
+  current_shell = Runner.get_interactive_session
+  assert_equal(exit_code, current_shell.last_exit_code)
+end
+
+# Assert that the shell exited with an error code (assuming non-0 is an error)
+#
+# @step_input exit_code [Integer] The expected exit code
+Then('the shell exited with an error code') do
+  current_shell = Runner.get_interactive_session
+  assert(current_shell.last_exit_code != 0, "Shell exited with code 0")
 end
--- a/lib/features/steps/runner_steps.rb
+++ b/lib/features/steps/runner_steps.rb
@@ -83,6 +83,14 @@ When('I run {string} on the current terminal') do |command|
   assert(success, 'The terminal had already closed')
 end
 
+# Get the current buffer values in the terminal
+#
+# @step_input expected_chars [String] The chars present in current buffer
+Then('the terminal is outputting {string}') do |expected_chars|
+  current_terminal = Runner.get_interactive_session
+  assert_equal(expected_chars, current_terminal.current_buffer)
+end
+
 # Verify a string appears in the stdout logs
 #
 # @step_input expected_line [String] The string present in stdout logs
--- a/lib/features/steps/runner_steps.rb
+++ b/lib/features/steps/runner_steps.rb
@@ -62,3 +62,49 @@ When('I wait for {int} second(s)') do |seconds|
   $logger.warn 'Sleep was used! Please avoid using sleep in tests!'
   sleep(seconds)
 end
+
+# Starts an interactive terminal
+When('I start a new terminal') do
+  Runner.stop_interactive_session
+  Runner.get_interactive_session
+end
+
+# Stops currently running interactive terminal
+When('I stop the current terminal') do
+  Runner.stop_interactive_session
+end
+
+# Run a command on the terminal
+#
+# @step_input command [String] The command to run on the terminal
+When('I run {string} on the current terminal') do |command|
+  current_terminal = Runner.get_interactive_session
+  success = current_terminal.run_command(command)
+  assert(success, 'The terminal had already closed')
+end
+
+# Verify a string appears in the stdout logs
+#
+# @step_input expected_line [String] The string present in stdout logs
+Then('the terminal has output {string}') do |expected_line|
+  current_terminal = Runner.get_interactive_session
+  match = current_terminal.parsed_output.any? { |line| line == expected_line }
+  assert(match, "No output lines matched #{expected_line}")
+end
+
+# Verify a string appears in the stderr logs
+#
+# @step_input expected_err [String] The string present in stderr logs
+Then('the terminal has the error message {string}') do |expected_err|
+  current_terminal = Runner.get_interactive_session
+  match = current_terminal.parsed_errors.any? { |line| line == expected_err }
+  assert(match, "No output lines matched #{expected_err}")
+end
+
+# Verify the exit code of the terminal
+#
+# @step_input exit_code [Integer] The expected exit code
+Then('the terminal exit code equals {int}') do |exit_code|
+  current_terminal = Runner.get_interactive_session
+  assert_equal(exit_code, current_terminal.last_exit_code)
+end
--- a/lib/features/steps/runner_steps.rb
+++ b/lib/features/steps/runner_steps.rb
@@ -77,7 +77,7 @@ end
 # Run a command on the terminal
 #
 # @step_input command [String] The command to run on the terminal
-When('I run {string} on the current terminal') do |command|
+When('I input {string} interactively') do |command|
   current_terminal = Runner.get_interactive_session
   success = current_terminal.run_command(command)
   assert(success, 'The terminal had already closed')
@@ -115,4 +115,12 @@ end
 Then('the terminal exit code equals {int}') do |exit_code|
   current_terminal = Runner.get_interactive_session
   assert_equal(exit_code, current_terminal.last_exit_code)
+end
+
+# Assert that the terminal exited with an error code
+#
+# @step_input exit_code [Integer] The expected exit code
+Then('the terminal exited with an error code') do
+  current_terminal = Runner.get_interactive_session
+  assert(current_terminal.last_exit_code != 0, "Terminal exited with code 0")
 end
--- a/lib/features/support/interactive_cli.rb
+++ b/lib/features/support/interactive_cli.rb
@@ -18,7 +18,10 @@ class InteractiveCLI
   #   @return [Number] The PID of the running terminal
   attr_reader :pid
 
+  # @!attribute [r] current_buffer
+  #   @return [String] A string representation of the current output present in the terminal
   attr_reader :current_buffer
+
   # Creates an InteractiveCLI instance
   #
   # @param environment [Hash] A hash of environment variables
@@ -30,9 +33,11 @@ class InteractiveCLI
     @parsed_errors = []
     @current_buffer = ''
     @last_exit_code = nil
+    @in_stream = nil
+    @pid = nil
     @running = false
     @stop_command = stop_command
-    start_terminal(terminal)
+    start_threaded_terminal(terminal)
   end
 
   # Attempts to stop the terminal using the preset command
@@ -46,7 +51,7 @@ class InteractiveCLI
   #
   # @return [Boolean] Whether the stream is currently running
   def running
-    @in_stream && !@in_stream.closed?
+    !(@in_stream.nil? || @in_stream.closed?)
   end
 
   # Runs the given command if the terminal is running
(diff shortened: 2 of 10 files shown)


lib/features/support/interactive_cli.rb
```diff
@@ -0,0 +1,117 @@
...
+
+  # @!attribute [r] current_buffer
+  #   @return [String] A string representation of the current output present in the terminal
+  attr_reader :current_buffer
+
+  # Creates an InteractiveCLI instance
+  #
+  # @param environment [Hash] A hash of environment variables
+  # @param terminal [String] A path to the terminal to run, defaults to `/bin/sh`
+  # @param terminal [String] The stop command, defaults to `exit`
```
imjoehaines, line 29:
```suggestion
  # @param stop_command [String] The stop command, defaults to `exit`
```

lib/features/support/interactive_cli.rb
```diff
@@ -0,0 +1,117 @@
...
+  attr_reader :pid
+
+  # @!attribute [r] current_buffer
+  #   @return [String] A string representation of the current output present in the terminal
+  attr_reader :current_buffer
+
+  # Creates an InteractiveCLI instance
+  #
+  # @param environment [Hash] A hash of environment variables
+  # @param terminal [String] A path to the terminal to run, defaults to `/bin/sh`
```
imjoehaines, line 28: I think this would be more accurately a `shell` if the intention is only to execute sh, bash, zsh etc... or maybe just `command` if it can be anything that `Open3` can handle

lib/features/support/interactive_cli.rb
```diff
@@ -0,0 +1,117 @@
...
+    @env = environment
+    @parsed_output = []
+    @parsed_errors = []
+    @current_buffer = ''
+    @last_exit_code = nil
+    @in_stream = nil
+    @pid = nil
+    @running = false
+    @stop_command = stop_command
+    start_threaded_terminal(terminal)
```
imjoehaines, line 40: Generally I prefer avoiding doing work in a constructor; could we provide a public `start` method to be called on a new instance instead?

lib/features/support/interactive_cli.rb
```diff
@@ -0,0 +1,117 @@
...
+    @in_stream.puts(command)
+    true
+  end
+
+  private
+
+  # Starts a terminal on another thread
+  #
+  # @param terminal [String] A path to the terminal to run
+  def start_threaded_terminal(terminal)
```
imjoehaines, line 73: Does this not work properly without being run in a thread? I'm wondering if we need this method, as callers could start it in a new thread if they needed to

If it always has to be run in a thread, is it worth inlining `start_terminal` here so that it doesn't look like there are two ways to start it?

lib/features/support/interactive_cli.rb
```diff
@@ -0,0 +1,117 @@
+require 'open3'
+
+# Encapsulates a terminal session, retaining state and input streams for interactive tests
+class InteractiveCLI
+  # @!attribute [r] parsed_output
+  #   @return [Array] An array of output strings received from the terminals STDOUT pipe
+  attr_reader :parsed_output
```
imjoehaines, line 7: I'd be tempted to call this `stdout_lines` for clarity, as there's no really any parsing going on

lib/features/support/interactive_cli.rb
```diff
@@ -0,0 +1,117 @@
...
+
+  # @!attribute [r] parsed_error
+  #   @return [Array] An array of error strings received from the terminals STDERR pipe
+  attr_reader :parsed_errors
```
imjoehaines, line 11: As above, I think `stdderr_lines` might be clearer

lib/features/support/interactive_cli.rb
```diff
@@ -0,0 +1,117 @@
...
+    @running = false
+    @stop_command = stop_command
+    start_threaded_terminal(terminal)
+  end
+
+  # Attempts to stop the terminal using the preset command
+  #
+  # @return [Boolean] true if successful, false otherwise
+  def stop
+    run_command(@stop_command)
```
imjoehaines, line 47: Does the Open3 thread exit after this is run? If so it might be worth waiting for that to happen and also the thread that's created by `start_threaded_terminal`, e.g.

```suggestion
    run_command(@stop_command)
    @thread.join(10)
```

test/cli_test.rb
```diff
@@ -0,0 +1,76 @@
...
+class InteractiveCLITest < Test::Unit::TestCase
+  def setup
+    @capabilities = { key: 'value' }
+
+    ENV.delete('BRANCH_NAME')
+    ENV.delete('BUILDKITE')
+    ENV.delete('BUILDKITE_PIPELINE_NAME')
+    ENV.delete('BUILDKITE_BUILD_NUMBER')
+    ENV.delete('BUILDKITE_RETRY_COUNT')
+    ENV.delete('BUILDKITE_STEP_KEY')
```
imjoehaines, line 15: Does this need to happen? It doesn't look like `ENV` is used anywhere else in this PR

test/fixtures/cli/Gemfile
```diff
@@ -0,0 +1,4 @@
+source "https://rubygems.org"
```
imjoehaines, line 1: Is this file needed for anything?

lib/features/steps/runner_steps.rb
```diff
@@ -62,3 +62,65 @@
   $logger.warn 'Sleep was used! Please avoid using sleep in tests!'
   sleep(seconds)
 end
+
+# Starts an interactive terminal
+When('I start a new terminal') do
+  Runner.stop_interactive_session
+  Runner.get_interactive_session
```
imjoehaines, line 69: I think this can wait for it to be ready for input by sleeping until it has a `pid`, e.g.

```suggestion
  cli = Runner.get_interactive_session
  attempts = 0

  until cli.pid
    raise "did not start in time" if attempts > 10 # or some number that makes more sense

    sleep(0.1) # or whatever
    attempts += 1
  end
```

lib/features/support/interactive_cli.rb
```diff
@@ -0,0 +1,117 @@
...
+  # @!attribute [r] parsed_error
+  #   @return [Array] An array of error strings received from the terminals STDERR pipe
+  attr_reader :parsed_errors
+
+  # @!attribute [r] last_exit_code
+  #   @return [Number] The exit code of the last terminal command
+  attr_reader :last_exit_code
+
+  # @!attribute [r] pid
+  #   @return [Number] The PID of the running terminal
```
imjoehaines, line 18: This can be `nil` if it hasn't started yet

```suggestion
  #   @return [Number, nil] The PID of the running terminal
```

lib/features/support/interactive_cli.rb
```diff
@@ -0,0 +1,117 @@
...
+  # @!attribute [r] parsed_output
+  #   @return [Array] An array of output strings received from the terminals STDOUT pipe
+  attr_reader :parsed_output
+
+  # @!attribute [r] parsed_error
+  #   @return [Array] An array of error strings received from the terminals STDERR pipe
+  attr_reader :parsed_errors
+
+  # @!attribute [r] last_exit_code
+  #   @return [Number] The exit code of the last terminal command
```
imjoehaines, line 14: This will be `nil` if a command hasn't been run

```suggestion
  #   @return [Number, nil] The exit code of the last terminal command
```

lib/features/support/interactive_cli.rb
```diff
@@ -0,0 +1,117 @@
...
+  # Starts a terminal
+  #
+  # @param terminal [String] A path to the terminal to run
+  def start_terminal(terminal)
+    Open3.popen3(@env, terminal) do |stdin, stdout, stderr, wait_thr|
+      @pid = wait_thr.pid
+      $logger.debug(pid) { 'Starting terminal session' }
+
+      @in_stream = stdin
+      @out_stream = stdout
```
imjoehaines, line 90: It looks like this is the only place it's used, so I think it can deleted

```suggestion
```

imjoehaines (changes requested):
Left a lot of comments but most are about naming

I think the important things are:
- wait for the shell to be ready for input before moving on from the "I start a new terminal" step
- adding steps that wait for output, so you don't need to `sleep 2` in tests
- I think most/all places where "terminal" is used would be more accurate as "shell"

It might also be a good idea to add stricter assertions so that false positives are less likely, e.g.:

- `the last line of stdout is {string}`
- `the last line of stderr is {string}`

lib/features/steps/runner_steps.rb
Cawllec: Yes, this was something I was planning to do (to remove all the waits).

lib/features/support/interactive_cli.rb
Cawllec: Yep, that was for debugging, no longer required

lib/features/steps/runner_steps.rb
```diff
@@ -62,3 +62,65 @@
...
+When('I input {string} interactively') do |command|
+  current_terminal = Runner.get_interactive_session
+  success = current_terminal.run_command(command)
+  assert(success, 'The terminal had already closed')
+end
+
+# Get the current buffer values in the terminal
+#
+# @step_input expected_chars [String] The chars present in current buffer
+Then('the terminal is outputting {string}') do |expected_chars|
```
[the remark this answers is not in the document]
Cawllec: That's true.  I think I tend towards the first `the current stdout line is {string}`

lib/features/support/interactive_cli.rb
twometresteve: Interesting - I wasn't aware of the ability/convention to explicitly not the possibility of `nil`s.

lib/features/steps/runner_steps.rb
```diff
@@ -62,3 +62,65 @@
...
+  assert_equal(expected_chars, current_terminal.current_buffer)
+end
+
+# Verify a string appears in the stdout logs
+#
+# @step_input expected_line [String] The string present in stdout logs
+Then('the terminal has output {string}') do |expected_line|
+  current_terminal = Runner.get_interactive_session
+  match = current_terminal.parsed_output.any? { |line| line == expected_line }
+  assert(match, "No output lines matched #{expected_line}")
```
twometresteve, line 100: I think we should always output the actual lines received as well to make debugging possible.  It might mean a lot of output, but as long as it's formatted in a friendly manner (e.g. multiline and clear delimiters at start/end) I would rather have lots of output than none at all.

lib/features/steps/runner_steps.rb
```diff
@@ -62,3 +62,65 @@
...
+Then('the terminal has the error message {string}') do |expected_err|
+  current_terminal = Runner.get_interactive_session
+  match = current_terminal.parsed_errors.any? { |line| line == expected_err }
+  assert(match, "No output lines matched #{expected_err}")
+end
+
+# Verify the exit code of the terminal
+#
+# @step_input exit_code [Integer] The expected exit code
+Then('the terminal exit code equals {int}') do |exit_code|
```
twometresteve, line 115: "terminal exit code" sounds to me like the terminal has finished and this is the exit code for the whole process.  But multiple commands can be run in the same terminal, right?  And this is just the exit code of the last command?  
Assuming I understand that correctly then maybe just reword slightly?
```suggestion
Then('the last terminal exit code is {int}') do |exit_code|
```
(Changed `equals` to `is` for more consistency, although there is probably a whole piece there for us across all the existing steps!).

lib/features/support/interactive_cli.rb
twometresteve: I agree, above doing processing beyond basic initialization in constructors - it can be surprising for clients and make testing more difficult.

test/fixtures/cli/Gemfile
twometresteve: Is it so we can run `bundle exec maze-runner` in the local directory?

test/fixtures/cli/features/CLI.feature
```diff
@@ -0,0 +1,24 @@
+Feature: Interactive CLI support
+
+  Scenario: Supports a node script
+    Given I start a new terminal
+    And I wait for 2 seconds
```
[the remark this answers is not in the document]
twometresteve: I agree, but it's a good candidate for a follow-on PR to be balanced with everything else.

twometresteve:
This looks a nicely forming piece of work with some great comments coming in from @imjoehaines .

To take this forward, please can you...
- Create a new `integration/interactive-cli` branch from master and retarget this onto it.
- Address just the straightforward comments on this PR and seek approval to merge
- Raise further, bite-size, PRs to address any comments that will take more thought/time to implement. (And as part of that consider if we want to address it straight away or not).

That way you get the bulk of the work accepted, reviewing further changes becomes much easier and more people can come in to help (if needed).  It also means the project(s) that need this can get running against the integration branch much sooner and start providing feedback, without us having to commit to a release.

lib/features/support/interactive_cli.rb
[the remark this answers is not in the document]
Cawllec: I think that's reasonable, in the sense it's merely the output from the `run_command`.  I'll make it match up with the comment below.

test/fixtures/cli/features/CLI.feature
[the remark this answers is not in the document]
Cawllec: This should be removed by waiting for the pid to be present, as recommended in other comments

lib/features/support/interactive_cli.rb
Cawllec: I think `shell` or `terminal` is appropriate as the intent is to present an interactive Terminal.  Making it command could be more accurate, but less descriptive.

lib/features/support/interactive_cli.rb
Cawllec: I'm not necessarily sure I agree.  

The purpose of the object is to represent a single-run shell or terminal which can be interacted with.  Adding a single-use `start` method, which must then determine whether the instance has previously been started and handle being called multiple times on a single instance, seems like an unnecessary addition.

Reading the name and purpose of the class I would assume that, by creating an instance of the class, I would be able to interact with a shell, with no further required setup.  You can argue that this can be managed by it being started before it is returned by the `Runner` class, but I'm not sure that actually adds anything other than an extra step.

It may help with unit testing, but even that's easily worked around by stubbing and mocking class methods.

test/cli_test.rb
Cawllec: Nope, just a leftover from a copied test setup :)

lib/features/support/interactive_cli.rb
twometresteve: My comment was based mainly on experience of similar setups in the past and me having developed a general feeling about it over the years, rather than having a specific concern for this particularly instance.  Certainly opinion and style more than fact in that respect. :-). Also, to be fair, it seems much easier to mock things when unit testing in Ruby than other languages I have dealt with... so if you're happy that you can unit test it effectively without spinning up external processes (for instance), I can many deal with it staying as it is.

imjoehaines: Yeah I've been bitten by this in the past, but I don't think there's anything especially egregious about this particular case 🙂 — happy for it to stay as-is

lib/features/steps/runner_steps.rb
```diff
@@ -62,3 +62,65 @@
...
+  current_terminal = Runner.get_interactive_session
+  assert_equal(exit_code, current_terminal.last_exit_code)
+end
+
+# Assert that the terminal exited with an error code
+#
+# @step_input exit_code [Integer] The expected exit code
+Then('the terminal exited with an error code') do
```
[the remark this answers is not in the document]
twometresteve: Should we have either of those?  An exit code of 0 for success is only a convention (albeit very common) and there may be other codes that are considered "success".  I think people could make the wrong assumption in certain contexts and it would be safer just to go on explicit exit codes.  Have a step `the exit code is not {int}` to deal with anything non-zero, if that is wanted (which I can see it would be in most practical cases).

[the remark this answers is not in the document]
imjoehaines: I think a specific exit code should always have specific text detailing the error too, which is much more understandable to assert against IMO

The only time a specific exit code might be important to check is if we're testing a tool that we wrote, but even then the output is what humans read so seems more useful to check that

I don't think any tool uses 0 for failure or >0 for success as those have been the understood meanings of exit codes for decades (and they'd fail every CI build 😀)

lib/features/steps/runner_steps.rb
twometresteve: (Happy for this to be handled in a separate PR, of course).

lib/features/support/interactive_cli.rb
Cawllec: This is one of the requirements for unit testing the class, as mocha will not function on objects running in other threads.

lib/features/steps/runner_steps.rb
Cawllec: This should be being output via the CLI instance, whenever it receives a newline char it will store the current line, `$logger.debug` the line (against the shell PID), and then clear for the next line

lib/features/steps/runner_steps.rb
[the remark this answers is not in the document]
Cawllec: I think making an assumption that 0 is a success is reasonable, so long as that's documented and it's possible to look for other codes.

lib/features/steps/runner_steps.rb
Cawllec: I don't believe we'll get an exit code through Open3 other than the shell exit code, so i believe this is valid.

Cawllec:
While running through the reviews I believe I've actually addressed everything.  If I could get a re-review I'll separate adding more unit and e2e tests into a different task/PR.

lib/features/support/runner.rb
```diff
@@ -69,8 +70,32 @@ def run_script(script_name, blocking: false, success_codes: [0])
...
+    # @return [InteractiveCLI] The interactiveCLI instance
+    def get_interactive_session
+      @interactive_session ||= InteractiveCLI.new(environment)
+      attempts = 0
+      until @interactive_session.pid
+        raise "Shell session would not start within 3 seconds" if attempts > 10 # allows 3 seconds of opening
+        attempts += 1
+        sleep(0.3)
+      end
+      @interactive_session
```
imjoehaines, line 84: This works but I find it a little confusing as it runs through the `until` even though it's already started; I think something like this is clearer (but doesn't have to be in this PR)

```suggestion
      return @interactive_session unless @interactive_session.nil?

      interactive_session = InteractiveCLI.new(environment)
      attempts = 0
      until @interactive_session.pid
        raise "Shell session would not start within 3 seconds" if attempts > 10 # allows 3 seconds of opening
        attempts += 1
        sleep(0.3)
      end

      @interactive_session = interactive_session
```

i.e. if there's a session already then return it straight away, otherwise make a new one

imjoehaines (approved):

lib/features/steps/runner_steps.rb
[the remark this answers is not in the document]
twometresteve: 👍

lib/features/support/runner.rb
twometresteve: Thanks, I've taken a note to address this in a follow-on PR.

twometresteve (approved):
LGTM, subject to a few points to come in follow-on PRs.

PR #145, opened 2020-10-19
Provide intended OS version (targeting master)

twometresteve:
## Goal

This is a representation of #142 , which was a chained PR that got merged prematurely.  

The commits and diff are precisely the same.


lib/features/support/maze_runner.rb
```diff
@@ -7,8 +7,8 @@
 class MazeRunner
   class << self
     attr_accessor :driver
-    def configuration
-      @configuration ||= Configuration.new
+    def config
```
Cawllec, line 10: Not concerned at all, just wondering the reason behind this change?

twometresteve: Just for brevity, that's all.  Some areas of the code are a bit noisy currently and it was easy to make this change now that there are no notifiers currently making use of it.

Cawllec (approved):

PR #157, opened 2020-11-03
[PLAT-4982] Server response code steps

Cawllec:
## Goal

Adds steps to allow setting the HTTP status code response of the servlet to any given POST or OPTIONS request.

This will work in one of two ways:
- Setting an HTTP status code that will be used for all subsequent requests
- Setting an HTTP status code that will be used for the next received request, after which the code will return to `200`

## Tests

- Added maze test fixture exercising changes.


twometresteve (approved):
Very tidy.

PR #159, opened 2020-11-04
Add assertions for and auto-checking of Bugsnag-Integrity header

fractalwrench:
## Goal

Adds `the Bugsnag-Integrity header is valid` as a step to validate the `Bugsnag-Integrity` header. This should be a valid SHA-1 or simple digest.

The validation logic is covered by new unit tests. This has also been run against the Android CI here: https://github.com/bugsnag/bugsnag-android/pull/987

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Enhancements
 
-- Add assertions to verify Bugsnag-Integrity header is correct
+- Add assertions for and automatic verification of Bugsnag-Integrity headers
   [#159](https://github.com/bugsnag/maze-runner/pull/159)
 
 # 3.3.0 - 2020/11/05
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# TBD
+
+## Enhancements
+
+- Add assertions to verify Bugsnag-Integrity header is correct
+  [#159](https://github.com/bugsnag/maze-runner/pull/159)
+
 # 3.3.0 - 2020/11/05
 
 ## Enhancements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## Enhancements
 
-- Add assertions for and automatic verification of Bugsnag-Integrity headers
+- Add Cucumber steps for checking Bugsnag-Integrity headers, in addition to automatically 
+  verifying digests on all received requests when the header is present.
   [#159](https://github.com/bugsnag/maze-runner/pull/159)
 
 # 3.3.0 - 2020/11/05
--- a/lib/features/support/header.rb
+++ b/lib/features/support/header.rb
@@ -26,7 +26,7 @@ def valid_simple_digest_header(request)
   header = request[:request]['Bugsnag-Integrity']
   assert_not_nil(request[:body], 'No request body present')
   body = JSON.generate(request[:body])
-  computed_digest = "simple #{body.to_s.size}"
+  computed_digest = "simple #{body.to_s.bytesize}"
   assert_equal(header, computed_digest, "SHA1 digest does not match request payload. Computed #{computed_digest}.")
   header == computed_digest
 end
--- a/lib/features/support/header.rb
+++ b/lib/features/support/header.rb
@@ -3,30 +3,13 @@ require 'json'
 
 def valid_bugsnag_integrity_header(request)
   header = request[:request]['Bugsnag-Integrity']
-
+  digests = request[:digests]
   if header.start_with?('sha1')
-    assert_true(valid_sha_digest_header(request), 'Bugsnag-Integrity header does not match SHA1 digest.')
+    computed_digest = "sha1 #{digests[:sha1]}"
   elsif header.start_with?('simple')
-    assert_true(valid_simple_digest_header(request), 'Bugsnag-Integrity header does not match simple digest.')
+    computed_digest = "simple #{digests[:simple]}"
   else
-    assert_true(false, 'Bugsnag-Integrity header does not match SHA1 or Request Length digest.')
+    return false
   end
-end
-
-def valid_sha_digest_header(request)
-  header = request[:request]['Bugsnag-Integrity']
-  assert_not_nil(request[:body], 'No request body present')
-  body = JSON.generate(request[:body])
-  computed_digest = 'sha1 ' + Digest::SHA1.hexdigest(body.to_s)
-  assert_equal(header, computed_digest, "SHA1 digest does not match request payload. Computed #{computed_digest}.")
-  header == computed_digest
-end
-
-def valid_simple_digest_header(request)
-  header = request[:request]['Bugsnag-Integrity']
-  assert_not_nil(request[:body], 'No request body present')
-  body = JSON.generate(request[:body])
-  computed_digest = "simple #{body.to_s.bytesize}"
-  assert_equal(header, computed_digest, "SHA1 digest does not match request payload. Computed #{computed_digest}.")
   header == computed_digest
 end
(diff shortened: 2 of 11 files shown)


lib/features/support/header.rb
```diff
@@ -0,0 +1,25 @@
...
+
+def valid_bugsnag_integrity_header(request)
+  valid = valid_sha_digest_header(request) || valid_simple_digest_header(request)
+  assert_true(valid, "Bugsnag-Integrity header does not match SHA1 or Request Length digest.")
+end
+
+def valid_sha_digest_header(request)
+  header = request[:request]["Bugsnag-Integrity"]
+  assert_not_nil(request[:body], "No request body present")
+  body = JSON.generate(request[:body])
```
twometresteve, line 12: Computing the digest after Maze Runner has parsed JSON text into a Ruby hash, and this has generated it back into JSON, isn't going to work.  Any differences in whitespace will lead to a difference in the digest and it needs to be computed "at source".

Perhaps if you can action the other comments, then assign the ticket to me and I'll put that in though.

lib/features/support/header.rb
```diff
@@ -0,0 +1,25 @@
...
+  computed_digest = "sha1 " + Digest::SHA1.hexdigest(body.to_s)
+  assert_equal(header, computed_digest, "SHA1 digest does not match request payload. Computed #{computed_digest} from #{body}.")
```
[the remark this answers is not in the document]
fractalwrench: Removed logging of body

lib/features/support/header.rb
```diff
@@ -0,0 +1,25 @@
+require 'digest/sha1'
+require 'json'
+
+def valid_bugsnag_integrity_header(request)
+  valid = valid_sha_digest_header(request) || valid_simple_digest_header(request)
```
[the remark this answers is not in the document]
fractalwrench: Updated to check the digest type.

fractalwrench:
@twometresteve addressed the simpler review comments, assigning the rest over to you 🎉

twometresteve:
@imjoehaines - I performed the initial review here for @fractalwrench, but subsequently added some additional changes for the automatic checking of the`Bugsnag-Integrity` header.  If present, it must now be valid.

To test this I built the test fixture locally against the in-development Android notifier that now provides the header (hence migrating to the latest notifier version) and ran the tests with DEBUG on.  I was able to inspect the logs to verify that the headers were present and that the computed digests matched.

Note that some status checks are currently outstanding.  I have changed the Github branch protection rules after refactoring the pipeline tests on #164 .  It will come out in the wash once #164 is merged and this PR rebased.

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# TBD
+
+## Enhancements
+
+- Add assertions for and automatic verification of Bugsnag-Integrity headers
```
fractalwrench, line 5:
```suggestion
- Add assertions for automatic verification of Bugsnag-Integrity headers
```

twometresteve: Ah, what I really meant was `Add assertions for, and automatic verification of, Bugsnag-Integrity headers`(with commas), because there are both opt-in Cucumber steps and the no-option, check every request I get feature that have been added.  I'll reword to make that clearer.

lib/features/support/header.rb
```diff
@@ -0,0 +1,32 @@
...
+  computed_digest = 'sha1 ' + Digest::SHA1.hexdigest(body.to_s)
+  assert_equal(header, computed_digest, "SHA1 digest does not match request payload. Computed #{computed_digest}.")
+  header == computed_digest
+end
+
+def valid_simple_digest_header(request)
+  header = request[:request]['Bugsnag-Integrity']
+  assert_not_nil(request[:body], 'No request body present')
+  body = JSON.generate(request[:body])
+  computed_digest = "simple #{body.to_s.size}"
```
imjoehaines, line 29: This should use [`bytesize`](https://ruby-doc.org/core-2.7.2/String.html#method-i-bytesize), to handle multi-byte characters

```suggestion
  computed_digest = "simple #{body.to_s.bytesize}"
```

lib/features/support/servlet.rb
```diff
@@ -67,6 +68,35 @@ def log_request(request)
...
+    header = request['Bugsnag-Integrity']
+    return if header.nil?
+
+    # Header must have type and digest
+    parts = header.split ' '
+    raise "Invalid Bugsnag-Integrity header: #{header}" unless parts.size == 2
+
+    # Both digest types are stored whatever
+    sha1 = Digest::SHA1.hexdigest(request.body)
+    simple = request.body.length
```
imjoehaines, line 82: This also needs to be `bytesize`

```suggestion
    simple = request.body.bytesize
```

lib/features/support/servlet.rb
```diff
@@ -67,6 +68,35 @@ def log_request(request)
       $logger.debug "BODY: #{JSON.pretty_generate(JSON.parse(request.body))}"
     end
   end
+
+  def check_digest(request)
+    header = request['Bugsnag-Integrity']
+    return if header.nil?
```
imjoehaines, line 74: I wonder if we should have a flag that can be set in an after configuration hook, which would tell Maze Runner that a missing integrity header is an error (or even whether it should check the header at all)

The implicit checking is nice, but if it silently does nothing when the header isn't present then it feels a bit dangerous

e.g.

```ruby
AfterConfiguration do
  MazeRunner.config.check_integrity_header = true
end
```

```diff
+    return unless MazeRunner.config.check_integrity_header
+
     header = request['Bugsnag-Integrity']
-    return if header.nil?
+    raise 'No Bugsnag-Integrity header found' if header.nil?
```

imjoehaines (changes requested):

lib/features/support/servlet.rb
twometresteve: I'm not against that, but if we did that then I'd prefer it got passed in as a command line option.  The `config` construction is more a by-product of that and more for internal MR use that for clients to interact with.

Also, in terms of checking that a header is there once implemented by the notifier, I saw that as where the notifier should have its own scenarios to explicitly check for it (as Android has).

twometresteve: ...that is, the intention of the implicit checking was an additional safeguard on top of what notifier should be checking for themselves, as opposed to it being our only safety net.

lib/features/support/header.rb
imjoehaines: > Computing the digest after Maze Runner has parsed JSON text into a Ruby hash, and this has generated it back into JSON, isn't going to work. Any differences in whitespace will lead to a difference in the digest and it needs to be computed "at source".

I think we still need to do this before merging

lib/features/support/servlet.rb
imjoehaines: Yeah I get it, but it's a bit of a shame to have to add assertions to the notifier to know that it's being checked. This definitely doesn't block merging, but I think it'd be nice to opt in to a stricter mode eventually

About config vs command line options; I'd prefer leaning towards configuring MR in ruby if it's something that is universal to all MR tests, e.g. if we added a command line option to the Ruby pipeline we'd need to edit 71 different steps — not hard to do, but it's easy to miss some or forget to add it to new steps in the future

twometresteve: OK, I've raised a fresh internal ticket (PLAT-5419) to look at that.  Good point about accessing config directly at the language level if it works for a given notifier, I just wouldn't want to force other platforms down that route.  So we can add an option and the entry in `MazeRunner.config` will happen naturally.

lib/features/support/header.rb
twometresteve: Well spotted, that almost slipped through the net.  I've made some changes to the steps and unit tests, but will rerun the Android test that relies on it as well to ensure that it's what is required.

imjoehaines (approved):

PR #163, opened 2020-11-06
[PLAT-4983, PLAT-5294] Refactor and update Multipart request steps

Cawllec:
## Goal
Refactors Multipart steps out of general request assertions.
Adds additional steps including:
- Validating a header with a regex
- Verifying the request payload is a Multipart/form-data type
- Deprecating old multipart steps for shorter, more concise wording (old steps are still present for Android)
- Adding multipart JSON fixture comparison steps to avoid type conflicts with current methods
- Adds a basic self assertion to the comparison results (note, specific output is preferred)
- Adds last_exit_code checking for docker commands, and appropriate steps for testing 

## Tests

- Payload_helpers test feature added to cover the multipart steps and how to use the new steps.
- New docker steps added to docker test scenarios

## Notes
- This PR shouldn't be merged until subsequent PR updating [bugsnag-source-maps](https://github.com/bugsnag/bugsnag-source-maps) tests for these changes has been created and tested.

- This hasn't expanded on the Multipart comparison functionality beyond what's stated, so we still don't support integer or boolean comparison for individual multipart fields.  This will be a quick add if desired.


lib/features/steps/request_assertion_steps.rb
```diff
@@ -55,6 +55,16 @@
   assert_equal(header_value, Server.current_request[:request][header_name])
 end
 
+# Tests that a header matches a regex
+#
+# @step_input header_name [String] The header to test
+# @step_input regex_string [String] The regex to match with
+Then('the {string} header matches the regex {string}') do |header_name, regex_string|
+  regex = Regexp.new(regex_string)
+  value = Server.current_request[:request][header_name]
```
twometresteve, line 64: Is there a case for asserting that the header exists first, just for a friendlier message if it fails?

lib/features/steps/runner_steps.rb
```diff
@@ -64,6 +64,23 @@
...
 end
 
+# Allows validation of the last exit code of the last run docker-compose command.
+# Will fail if no commands have been run.
+#
+# @step_input expected_code [Integer] The expected exit code
+Then('the exit code of the last docker command was {int}') do |expected_code|
+  exit_code = Docker.last_exit_code
+  assert_not_nil(exit_code, 'No docker exit code available to verify')
+  asset_equal(exit_code, expected_code)
```
twometresteve, line 74: Missing `r` in assert, but also the expected code should be the first argument.  See also L81.

test/fixtures/docker-app/features/run_docker_services.feature
```diff
@@ -12,14 +12,17 @@ Feature: Running docker services
         Given I have built the service "test_1"
         When I start the service "test_1"
         Then the service "test_1" should be running
+        And the last run docker command exited successfully
```
twometresteve, line 15: I don't think these features are being run currently.  I believe there would need to be an entry in `docker-compose.yml` and a separate build step in the pipeline to run them.

lib/features/steps/multipart_request_steps.rb
```diff
@@ -0,0 +1,126 @@
...
+  steps %(
+    Then the "Content-Type" header matches the regex "^multipart\\/form-data; boundary=[\\h-]+$"
+    And the multipart request has a non-empty body
+  )
+end
+
+# Tests the number of fields a multipart request contains.
+#
+# @step_input part_count [Integer] The number of expected fields
+Then('the multipart request has {int} fields') do |part_count|
```
imjoehaines, line 21: I think all of these steps that check `Server.current_request[:body]` aren't specific to multipart requests, so I'm not sure they're needed

My understanding from the source maps tests is that `Server.current_request[:body]` is the parsed request body, so it doesn't matter if it's a JSON or multipart request the parsed body would be the same

So I think these are unnecessary as the existing `the payload field "blah" equals "blah"` steps work for multipart requests already

lib/features/steps/multipart_request_steps.rb
```diff
@@ -0,0 +1,126 @@
...
+# @step_input fixture_path [String] Path to a JSON fixture
+Then('the multipart body does not match the JSON fixture in {string}') do |fixture_path|
+  raw_payload_value = Server.current_request[:body]
+  payload_value = {}
+  raw_payload_value.each_with_object(payload_value) do |(k, v), out|
+    out[k] = JSON.parse(v.to_s)
+  rescue JSON::ParserError
+    out[k] = v.to_s
+  end
+  expected_value = JSON.parse(open(fixture_path, &:read))
```
imjoehaines, line 94: I think (my memory is terrible, sorry!) part of the reason I wrote new steps was because I couldn't figure out where this was relative to, so I made mine use an absolute path instead — it'd be good to have docs for where the JSON file should be/what the CWD is (and possible be able to change that)

IMO it's also nice for these files not to be in `features/fixtures`, because some repos will have a lot of different fixtures, so it's easier to see when they are separate (I put them in `features/expected`)

lib/features/steps/multipart_request_steps.rb
```diff
@@ -0,0 +1,126 @@
...
+Then('the multipart body matches the JSON fixture in {string}') do |fixture_path|
+  raw_payload_value = Server.current_request[:body]
+  payload_value = {}
+  raw_payload_value.each_with_object(payload_value) do |(k, v), out|
+    out[k] = JSON.parse(v.to_s)
+  rescue JSON::ParserError
+    out[k] = v.to_s
+  end
+  expected_value = JSON.parse(open(fixture_path, &:read))
+  result = value_compare(expected_value, payload_value)
```
imjoehaines, line 112: It might be worth extracting this to a function that this & the above step can share, if only the assertion is different

lib/features/support/docker.rb
```diff
@@ -8,6 +8,16 @@ class << self
...
 
+    # @!attribute [w] last_exit_code allows reset of the last_exit_code in hooks
+    attr_writer :last_exit_code
+
+    # Stores the last exit code of any docker-compose commands for later comparison
+    #
+    # @returns [Integer, Nil] The last exit code, nil if no command has been run
+    def last_exit_code
+      @last_exit_code ||= nil
+    end
```
imjoehaines, line 19: This can just be an `attr_accessor` as the default is nil if it's not been set/doesn't exist yet

```suggestion
    attr_accessor :last_exit_code
```

e.g.

```ruby
irb(main):001:0> class Abc
irb(main):002:1>   attr_accessor :xyz
irb(main):003:1> end
=> nil
irb(main):004:0> a = Abc.new
=> #<Abc:0x007fbbd99e4a48>
irb(main):005:0> a.xyz
=> nil
irb(main):006:0> a.xyz = 123
=> 123
irb(main):007:0> a.xyz
=> 123
```

lib/features/support/docker.rb
```diff
@@ -55,7 +65,7 @@ def compose_project_name
     def run_docker_compose_command(command, compose_file: COMPOSE_FILENAME, success_codes: nil)
       project_name = compose_project_name.nil? ? '' : "-p #{compose_project_name}"
       command = "docker-compose #{project_name} -f #{compose_file} #{command}"
-      Runner.run_command(command, success_codes: success_codes)
+      @last_exit_code = Runner.run_command(command, success_codes: success_codes)
```
imjoehaines, line 68: `Runner.run_command` returns an array of exit code & output. I've ended up using both for the source map work (for checking error messages), so this could be:

```suggestion
      @last_exit_code, @output = Runner.run_command(command, success_codes: success_codes)
```

with `output` added as an `attr_accessor` & a step for reading it, e.g. https://github.com/bugsnag/bugsnag-source-maps/blob/9ab060ab28d4dd3f13a6fbb9fdbfd6d3c6cdca2b/features/steps/steps.rb#L88-L97

lib/features/steps/runner_steps.rb
```diff
@@ -64,6 +64,23 @@
   Docker.start_service(service, command: one_line_cmd)
 end
 
+# Allows validation of the last exit code of the last run docker-compose command.
+# Will fail if no commands have been run.
+#
+# @step_input expected_code [Integer] The expected exit code
+Then('the exit code of the last docker command was {int}') do |expected_code|
```
imjoehaines, line 71: I think it'd be nice to have an unsuccessful version as well; I don't think the exact exit code matters very often, but it's useful to check it's not 0

lib/features/steps/multipart_request_steps.rb
```diff
@@ -0,0 +1,126 @@
...
+require 'open-uri'
+require 'json'
+require 'cgi'
+
+include Test::Unit::Assertions
+
+# @!group Multipart request assertion steps
+
+# Verifies that the request contains multipart form-data
+Then('the request contains multipart form-data and corresponding headers') do
```
imjoehaines, line 11: How about something like this:

```suggestion
Then('the request is valid multipart form-data') do
```

lib/features/steps/multipart_request_steps.rb
```diff
@@ -0,0 +1,126 @@
...
+# @step_input part_key [String] The key to the multipart element
+Then('the field {string} for multipart request is null') do |part_key|
+  step %(the multipart field "#{part_key}" is null)
+end
+
+# Tests that the multipart payload body does not match a JSON fixture.
+# JSON formatted multipart fields will be parsed into hashes.
+#
+# @step_input fixture_path [String] Path to a JSON fixture
+Then('the multipart body does not match the JSON fixture in {string}') do |fixture_path|
```
imjoehaines, line 86: A check for the file existing would be good, so there's a better error if the path is wrong

```suggestion
Then('the multipart body does not match the JSON fixture in {string}') do |fixture_path|
  assert_true(File.exists?(fixture_path, "'#{fixture_path}' does not exist")
```

It might be even nicer to expand this to an absolute path, so it's super clear what's going on:

```suggestion
Then('the multipart body does not match the JSON fixture in {string}') do |fixture_path|
  full_path = File.realpath(fixture_path)
  assert_true(File.exists?(full_path, "'#{full_path}' does not exist")
```

test/fixtures/docker-app/features/run_docker_services.feature
Cawllec: I'll double check this, they _should_ be, but I can imagine they may not be

lib/features/steps/multipart_request_steps.rb
Cawllec: From some quick testing the `fixture_path` doesn't need to be in `features/fixtures`.  I'll change the variable name and doc string to resolve the issue.

lib/features/steps/multipart_request_steps.rb
Cawllec: Looking into it a bit further I believe you're correct.  I'll leave the older steps in to ensure backwards compatibility, but remove the newer ones as unnecessary.

lib/features/support/docker.rb
Cawllec: Ah, didn't realise you found both valuable, but that would be a good addition!

lib/features/steps/runner_steps.rb
```diff
@@ -64,6 +64,47 @@
...
+  exit_code = Docker.last_exit_code
+  assert_not_nil(exit_code, 'No docker exit code available to verify')
+  assert_not_equal(exit_code, 0)
+end
+
+# Allows testing a docker command output a specific string
+# Will fail if no commands have been run
+#
+# @step_input expected_string [String] The string expected in a single log line
+Then('the last run docker command output {string} to stdout') do |expected_string|
```
imjoehaines, line 97: The command is run with `popen2e` so there's no way to know if this was stdout or stderr as they are merged

```suggestion
Then('the last run docker command output {string}') do |expected_string|
```

lib/features/steps/multipart_request_steps.rb
```diff
@@ -0,0 +1,130 @@
...
+end
+
+# Tests the multipart request has at least one field.
+Then('the multipart request has a non-empty body') do
+  parts = Server.current_request[:body]
+  assert(parts.size.positive?, "Multipart request payload contained #{parts.size} fields")
+end
+
+# (Deprecated) Retained for backwards compatibility.
+# Use "the payload field {string} is not null"
```
imjoehaines, line 46: YARD has an `@deprecated` tag that should put this in the generated docs so it's more obvious

imjoehaines (approved):

lib/features/steps/multipart_request_steps.rb
twometresteve: Ok, we'll seek to update this in a future PR.

PR #164, opened 2020-11-09
Reinstate environment clearing between scenarios

twometresteve:
## Goal

Ensures that the environment is clear at the start of each scenario, save any explicitly set by the test itself.

## Design

This change allows clients to register blocks to be invoked at the following Cucumber lifecycle stages:
- AfterConfiguration
- Before
- After

Clients can use the above hooks directly, but Cucumber Ruby invokes them in the order in which there are declared, which can be problematic when test specific logic needs to be performed after framework logic (or before the logic in the case of After).

Having this mechanism in place allows clients to run logic in a predictable and intuitive way.

## Changeset

- Environment cleared at the start of each scenario
- New mechanism for registering callbacks, hanging off of `MazeRunner.hooks`
- Pipeline updated for new tests

## Tests

New framework test to verify the general mechanism and specific use case that triggered the change.


.buildkite/pipeline.yml
```diff
@@ -35,23 +35,19 @@ steps:
         run: unit-test
     command: 'bundle exec rake'
 
-  - label: 'Comparison tests'
+  - label: 'No-device tests'
```
twometresteve, line 38: I've refactored this into a single steps to avoid asking for several AWS instances, for a brief period of time (which will then take time to scale down).

lib/features/hooks/hooks.rb
```diff
@@ -54,25 +54,31 @@
...
-  #   thing, but a number of our pipelines rely on certain variables existing throughout (e.g. Ruby).
-  # Runner.environment.clear
+  Runner.environment.clear
   Server.stored_requests.clear
   Store.values.clear
 
-  next if MazeRunner.config.farm == :none
+  unless MazeRunner.config.farm == :none
+    MazeRunner.driver.start_driver if MazeRunner.config.appium_session_isolation
+  end
```
imjoehaines, line 70:
```suggestion
  if MazeRunner.config.farm != :none && MazeRunner.config.appium_session_isolation
    MazeRunner.driver.start_driver
  end
```

lib/features/steps/runner_steps.rb
```diff
@@ -8,6 +8,21 @@
   Runner.environment[key] = value
 end
 
+# Verifies that an environment variable is set to a given value.
+#
+# @step_input key [String] The environment variable
+# @step_input value [String] The intended value of the environment variable
+Then('the environment variable {string} equals {string}') do |key, expected_value|
+  assert_equal(expected_value, Runner.environment[key])
+end
```
imjoehaines, line 17: I don't think this should be checking `Runner.environment` as it doesn't necessarily represent the actual environment variables

```suggestion
Then('the environment variable {string} equals {string}') do |key, expected_value|
  assert_equal(expected_value, ENV[key])
end
```

lib/features/steps/runner_steps.rb
```diff
@@ -8,6 +8,21 @@
...
+
+# Verifies that an environment variable is not set.
+#
+# @step_input key [String] The environment variable
+Then('the environment variable {string} is null') do |key|
+  assert_nil(Runner.environment[key])
+end
```
imjoehaines, line 24:
```suggestion
Then('the environment variable {string} is null') do |key|
  assert_nil(ENV[key])
end
```

or if this should check that it's not set then maybe this would be better:

```suggestion
Then('the environment variable {string} is not set') do |key|
  assert_false(ENV.key?(key))
end
```

test/fixtures/framework/features/support/env.rb
```diff
@@ -0,0 +1,4 @@
+MazeRunner.hooks.before do |scenario|
```
imjoehaines, line 1: I think this would be more useful if it checked the different kind of hooks, when they are called and the order that two of the same type of hook are called

imjoehaines (commented):
It'd be good to remove `Before`/`After` etc... from [DOCS.md](https://github.com/bugsnag/maze-runner/blob/master/DOCS.md#writing-features) and replace it with this new way to do things

lib/features/hooks/hooks.rb
twometresteve: Yep, and actually Rubocop then wants it to be a 1-liner.

lib/features/hooks/hooks.rb
```diff
@@ -54,25 +54,31 @@
...
   end
   MazeRunner.driver.start_driver unless config.appium_session_isolation
+
+  # Call any blocks registered by the client
+  MazeRunner.hooks.call_after_configuration config
 end
 
 # Before each scenario
 Before do |scenario|
-  STDOUT.puts "--- Scenario: #{scenario.name}"
```
twometresteve, line 61: I shouldn't have removed this - reinstating.

lib/features/steps/runner_steps.rb
twometresteve: Discussed IRL.  New steps (which are only intended for the verification of MazeRunner functionality) have been renamed and pushed into the fixture directory.

twometresteve:
> It'd be good to remove `Before`/`After` etc... from [DOCS.md](https://github.com/bugsnag/maze-runner/blob/master/DOCS.md#writing-features) and replace it with this new way to do things

Just realised I missed this - will do.

twometresteve:
> > It'd be good to remove `Before`/`After` etc... from [DOCS.md](https://github.com/bugsnag/maze-runner/blob/master/DOCS.md#writing-features) and replace it with this new way to do things
> 
> Just realised I missed this - will do.

Have now given `DOCS.md` a general refresh.

test/fixtures/framework/features/support/env.rb
twometresteve: Yes, I agree - I'll do this as a follow-up PR straight afterwards, if that's ok, just so that we can get a release out.

imjoehaines (approved):

PR #167, opened 2020-11-12
Allow the Bugsnag-Integrity header

imjoehaines:
## Goal

The current `Access-Control-Allow-Headers` list doesn't include Bugsnag-Integrity, so requests using it fail in web browsers

> Request header field Bugsnag-Integrity is not allowed by Access-Control-Allow-Headers in preflight response.

## Tests

This is/will be tested by bugsnag-js; there aren't really any MR specific tests for this


twometresteve (approved):
LGTM - please just add an entry to the change log.

PR #170, opened 2020-11-13
Fix always false comparison for simple digest

imjoehaines:
## Goal

`parts[1]` is always a string because it's split from the header,
so will never equal the expected bytesize

Added `inspect` calls because this is the error without them:

> Given simple digest 1519 does not match the computed 1519

With them the error would have been:

> Given simple digest "1519" does not match the computed 1519


twometresteve (approved):
LGTM

PR #171, opened 2020-11-13
[v2] Backport allow bugsnag integrity header

imjoehaines:
## Goal

This is PR https://github.com/bugsnag/maze-runner/pull/167 backported to v2 so that JS can make use of it before it is migrated to v3


twometresteve (approved):
LGTM

PR #173, opened 2020-11-13
Reword request count assertion error message

fractalwrench:
## Goal

Rewords the error message when the number of requests does not match what is expected, as this has caused confusion in the past.

An example of what this looks like:

```
When I run "CXXAutoContextScenario"                                 # features/steps/android_steps.rb:1
And I wait to receive a request                                     # /maze-runner/lib/features/steps/request_assertion_steps.rb:12
Expected 1 requests but received 0 within the 30s timeout. This could indicate that Bugsnag crashed with a fatal error, or that it hasn’t made the requests that it should have done. Please check the device logs to confirm. (RuntimeError)
```

### DIFF
--- a/lib/features/steps/request_assertion_steps.rb
+++ b/lib/features/steps/request_assertion_steps.rb
@@ -25,7 +25,11 @@ Then('I wait to receive {int} request(s)') do |request_count|
     received = (Server.stored_requests.size >= request_count)
     sleep 0.1
   end
-  raise "Expected #{request_count} requests but received #{Server.stored_requests.size} within the 30s timeout. This could indicate that Bugsnag crashed with a fatal error, or that it hasn’t made the requests that it should have done. Please check the device logs to confirm." unless received
+  unless received
+    raise "Expected #{request_count} requests but received #{Server.stored_requests.size} within the 30s timeout. " \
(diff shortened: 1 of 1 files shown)


lib/features/steps/request_assertion_steps.rb
```diff
@@ -25,7 +25,8 @@
     received = (Server.stored_requests.size >= request_count)
     sleep 0.1
   end
-  raise "Requests not received in 30s (received #{Server.stored_requests.size})" unless received
+  raise "Expected #{request_count} requests but received #{Server.stored_requests.size} within the 30s timeout. This could indicate that Bugsnag crashed with a fatal error, or that it hasn’t made the requests that it should have done. Please check the device logs to confirm." unless received
```
twometresteve, line 28: To eliminate Rubocop violations:
```suggestion
  unless received
    raise "Expected #{request_count} requests but received #{Server.stored_requests.size} within the 30s timeout. " \
      'This could indicate that Bugsnag crashed with a fatal error, or that it hasn’t made the requests that it ' \
      'should have done. Please check the device logs to confirm.'
  end
```

twometresteve (changes requested):
Just Rubocop to fret about...

twometresteve (approved):
LGTM

PR #174, opened 2020-11-17
Allow more than one valid ISO8601 timezone format

imjoehaines:
## Goal

ISO8601 timezones can either be 'Z' for UTC or a timezone offset, e.g. '+01:00' or '-12:34'

The PHP notifier uses timezone offsets rather than forcing everything to UTC, so the current assertions fail — only the Laravel notifier has MR tests so this wasn't noticed at the time

## Tests

One of the tests now uses Ruby's `iso8601` method, which returns '+00:00' for UTC

### DIFF
--- a/lib/features/support/env.rb
+++ b/lib/features/support/env.rb
@@ -4,4 +4,4 @@ require 'securerandom'
 $api_key = SecureRandom.hex(16).tr('+/=', 'xyz')
 
 # A regex providing the pattern expected from timestamps
-TIMESTAMP_REGEX = /^\d{4}\-\d{2}\-\d{2}T\d{2}:\d{2}:[\d\.]+Z?$/
+TIMESTAMP_REGEX = /^\d{4}\-\d{2}\-\d{2}T\d{2}:\d{2}:[\d\.]+(Z|[+-]\d{2}:\d{2})?$/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# TBD
+
+## Enhancements
+
+- Allow timezone offsets in timestamps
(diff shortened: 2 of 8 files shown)


lib/features/support/env.rb
```diff
@@ -4,4 +4,4 @@
 $api_key = SecureRandom.hex(16).tr('+/=', 'xyz')
 
 # A regex providing the pattern expected from timestamps
-TIMESTAMP_REGEX = /^\d{4}\-\d{2}\-\d{2}T\d{2}:\d{2}:[\d\.]+Z?$/
+TIMESTAMP_REGEX = /^\d{4}\-\d{2}\-\d{2}T\d{2}:\d{2}:[\d\.]+(Z|[+-]\d{2}:\d{2})?$/
```
twometresteve, line 7:
```suggestion
TIMESTAMP_REGEX = /^\d{4}\-\d{2}\-\d{2}T\d{2}:\d{2}:[\d\.]+(Z|[+-]\d{2}(:?\d{2})?)?$/
```
Assuming we can trust [Wikipedia](https://en.wikipedia.org/wiki/ISO_8601), it appears that both the colon and minutes offset are optional (but obviously not colon with no minutes).

twometresteve: (Just noticed that Rubocop wants it frozen as well)

imjoehaines: They're not optional in [RFC3339](https://www.ietf.org/rfc/rfc3339.txt) (which is usually what people mean when they say ISO8601) so I'd be tempted to require them

twometresteve (approved):

twometresteve (approved):
LGTM

PR #176, opened 2020-11-18
Refactor parsing and processing of command line options

twometresteve:
## Goal

Split the existing `bugsnag-maze-runner` entry point into separate files to reduce its size and make unit testing of each aspect easier.

## Design

I've been as friendly as I can be, by performing the refactoring in three distinct commits - each should be reviewable in its own right with no overlap.

## Changeset

This is a refactor only, there should be no change in functionality.

## Tests

Should be covered by passing CI, but I have also performed some light local testing including use of the `--help` and `--version` options.  Some further testing with notifier clients before release would also seem prudent on this occasion.

### DIFF
--- a/bin/bugsnag-maze-runner
+++ b/bin/bugsnag-maze-runner
@@ -5,6 +5,7 @@ require 'cucumber/cli/main'
 require_relative '../lib/features/support/maze_runner'
 require_relative '../lib/features/support/option'
 require_relative '../lib/features/support/option_parser'
+require_relative '../lib/features/support/option_processor'
 require_relative '../lib/features/support/capabilities/capabilities'
 require_relative '../lib/features/support/capabilities/devices'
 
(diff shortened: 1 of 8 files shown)


bin/bugsnag-maze-runner
```diff
@@ -165,79 +68,18 @@ class MazeRunnerEntry
   def start(args)
     # Parse args, processing any Maze Runner specific options
     @args = args.dup
-    options = @parser.parse args
+    options = Maze::OptionParser.parse args
+    return unless options
```
imjoehaines, line 72: If this is just for the help/version cases then I think it's easier to `exit` there; this makes it seem like not getting any options is expected

lib/features/support/option.rb
```diff
@@ -0,0 +1,25 @@
...
+  USERNAME ||= 'username'
+  ACCESS_KEY ||= 'access-key'
+  BS_APPIUM_VERSION ||= 'appium-version'
+
+  # Local-only options
+  OS ||= 'os'
+  OS_VERSION ||= 'os-version'
+  APPIUM_SERVER ||= 'appium-server'
+  APPLE_TEAM_ID ||= 'apple-team-id'
+  UDID ||= 'udid'
```
imjoehaines, line 24: I don't think these need to be `||=`

```suggestion
  # Common options
  SEPARATE_SESSIONS = 'separate-sessions'
  FARM = 'farm'
  APP = 'app'
  A11Y_LOCATOR = 'a11y-locator'
  RESILIENT = 'resilient'

  # BrowserStack-only options
  BS_LOCAL = 'bs-local'
  BS_DEVICE = 'device'
  USERNAME = 'username'
  ACCESS_KEY = 'access-key'
  BS_APPIUM_VERSION = 'appium-version'

  # Local-only options
  OS = 'os'
  OS_VERSION = 'os-version'
  APPIUM_SERVER = 'appium-server'
  APPLE_TEAM_ID = 'apple-team-id'
  UDID = 'udid'
```

lib/features/support/option_processor.rb
```diff
@@ -0,0 +1,72 @@
...
+          raise "option --#{Option::OS} must be specified" if options[Option::OS].nil?
+          raise "option --#{Option::OS_VERSION} must be specified" if options[Option::OS_VERSION].nil?
+          # Ensure OS version is a valid float so that notifier tests can perform numeric checks, e.g:
+          # 'MazeRunner.config.os_version > 7'
+          unless /^[1-9][0-9]*(\.[0-9])?/.match? options[Option::OS_VERSION]
+            raise "option --#{Option::OS_VERSION} must be a valid OS version matching '/^[1-9][0-9]*(\\.[0-9])?/'"
+          end
+
+          os = MazeRunner.config.os = options[Option::OS].downcase
+          MazeRunner.config.os_version = options[Option::OS_VERSION].to_f
```
imjoehaines, line 53: This was like this before, but should probably just reference `config`

```suggestion
          os = config.os = options[Option::OS].downcase
          config.os_version = options[Option::OS_VERSION].to_f
```

lib/features/support/option.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+# Provides the set of Maze Runner command line options
+class Option
```
imjoehaines, line 4: Should this be in the `Maze` module?

imjoehaines (approved):

lib/features/support/option.rb
twometresteve: Without them you get loads of `warning: already initialized constant` messages comes out, because of the way that Ruby files are loaded into Cucumber (using the `-r`option). I started to look at ways of doing that differently, but it didn't seem to gain traction.

imjoehaines: Is that because cucumber is loading all the files in `lib/features/...`? So would putting this in `lib/maze/option.rb` fix it?

I think it'd be better to avoid having cucumber load files that it doesn't need to

lib/features/support/option_processor.rb
twometresteve: Good spot - I've fixed on the next branch, where it will start to matter more as I've performed further refactoring.

lib/features/support/option.rb
twometresteve: Yes, I realised that after creating the PR - it's fixed on my next branch.

lib/features/support/option.rb
twometresteve: Yes, that makes sense.  I'll do this on the next branch as well, as I have some further moves coming and it would be good to avoid awkward rebases for now.

bin/bugsnag-maze-runner
twometresteve: Changed as suggested.

imjoehaines: Sorry, I meant exit in the rescues in `option_parser`; then you always get `options` here

twometresteve: Got you, yes that would be cleaner.

PR #178, opened 2020-11-20
macOS automation support

twometresteve:
## Goal

Provides the ability to run tests against macOS apps (using `--farm=local` and `--os=macos`) using Appium and [AppiumForMac](https://github.com/appium/appium-for-mac).

## Design

Follows the thrust of the recent prototype developed by @nickdowell, updated to align with the current codebase.

## Changeset

- Addition of `macos` as a valid `--os`
- Setting of Appium capabilities for MacOS
- Hooks to launch/kill MacOS apps
- New Cucumber steps to support MacOS testing

## Tests

Tested successfully locally with the macOS test fixture and scenarios currently in development in `bugsnag-cocoa`.

### DIFF
--- a/lib/features/hooks/hooks.rb
+++ b/lib/features/hooks/hooks.rb
@@ -22,9 +22,9 @@ AfterConfiguration do |config|
                                                          config.appium_version,
                                                          config.capabilities_option
 
-    config.app_location = BrowserStackUtils.upload_app config.username,
-                                                       config.access_key,
-                                                       config.app_location
+    config.app = BrowserStackUtils.upload_app config.username,
+                                              config.access_key,
+                                              config.app
     BrowserStackUtils.start_local_tunnel config.bs_local,
                                          tunnel_id,
                                          config.access_key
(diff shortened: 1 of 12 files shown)


lib/features/hooks/hooks.rb
```diff
@@ -122,6 +126,9 @@
 
   if MazeRunner.config.appium_session_isolation
     MazeRunner.driver.driver_quit
+  elsif MazeRunner.config.os == 'macos'
+    # Close the app - without the sleep, launching the app for the next scenario intermittently fails
+    system("killall #{MazeRunner.config.app_location} && sleep 1")
```
nickdowell, line 131: It could be confusing that `app_location` does not contain the app location for macOS apps... maybe `app` would be more appropriate?

nickdowell (approved):

lib/features/hooks/hooks.rb
twometresteve: Yeah, seems reasonable.

PR #177, opened 2020-11-20
Add --capabilities option (with CL processing refactor)

twometresteve:
## Goal

Adds the `--capabilities` command line option to allow clients to specific an arbitrary set of additional Appium capabilities in tests.

## Design

This PR also continues the refactoring effort to:
- Separate the concerns of command line option parsing, validation and processing
- Move classes, where appropriate, into the `Maze` module.

## Changeset

- Add `--capabilities` options
- Refactor CL processing
- Opportunistic fix for logging of BrowserStack link when run locally
- Leftover comments from #176 actioned:
  - Avoid use of `||=` in `options.rb`
  - Move `Options`

## Tests

Unit tests added for CL parsing/validation/processing and CI provides reasonable coverage - but further testing against notifiers will be performed before making the next release.

PR #179, opened 2020-11-24
Detect invalid requests

twometresteve:
## Goal

Introduces default behaviour to enforce the presence of the Bugsnag-Integrity header on all JSON HTTP requests.

It also addresses a deficiency in the `I should receive no requests` step, which should wait for a given time in order that errant requests are detected. 

## Design

This change introduces the concept of invalid requests, whether that is due to the header missing or because the request could not be parsed as JSON.  This latter class of request would cause an exception to be output in the logs, but not cause scenarios to fail.  With this change in place, scenarios will now explicitly fail, providing details of the invalid requests detected.

## Changeset

- Improved error handled in the `Servlet`
- New `invalid_requests` array added to the `Server`
- New `Configuration` entries added to provide default wait times on receiving requests (and receiving no requests)
- Opportunistic resolution of Rubocop violations

## Tests

The interesting test cases for this change are where scenarios fail, which unfortunately is awkward to include in CI.  I've therefore performed a variety of tests locally against notifier scenarios.  Further, extended, tests will also be performed before the next release - not least of all because this change uncovers a few failures in notifier pipelines due to one or more notifier bugs.


lib/features/steps/request_assertion_steps.rb
```diff
@@ -13,29 +15,35 @@
...
   end
   unless received
     raise "Expected #{request_count} requests but received #{Server.stored_requests.size} within the 30s timeout. " \
-      'This could indicate that Bugsnag crashed with a fatal error, or that it hasn’t made the requests that it ' \
-      'should have done. Please check the device logs to confirm.'
+      "This could indicate that:\n" \
+      " - Bugsnag crashed with a fatal error\n" \
+      " - Bugsnag did not make the requests that it should have done.\n" \
+      " - The requests were made, but not deemed to be valid (e.g. missing integrity header)\n" \
+      'Please check the Maze Runner and device logs to confirm.'
```
kattrali, line 38: There's multiline strings:

```rb
    raise %Q(Expected #{request_count} requests but received #{Server.stored_requests.size} within the 30s timeout.
      This could indicate that:
      - Bugsnag crashed with a fatal error
      - Bugsnag did not make the requests that it should have done.
      - The requests were made, but not deemed to be valid (e.g. missing integrity header)
      Please check the Maze Runner and device logs to confirm.
    )
```

You can also use [heredoc](https://en.wikipedia.org/wiki/Here_document):

```rb
    raise <<-MESSAGE
    Expected #{request_count} requests but received #{Server.stored_requests.size} within the 30s timeout.
      This could indicate that:
      - Bugsnag crashed with a fatal error
      - Bugsnag did not make the requests that it should have done.
      - The requests were made, but not deemed to be valid (e.g. missing integrity header)
      Please check the Maze Runner and device logs to confirm.
    MESSAGE
```

[More examples in Ruby](https://www.rubyguides.com/2018/11/ruby-heredoc/)

lib/features/support/servlet.rb
```diff
@@ -40,6 +39,19 @@ def do_POST(request, response)
     end
     response.header['Access-Control-Allow-Origin'] = '*'
     response.status = Server.status_code
+  rescue JSON::ParserError => e
+    msg = "Unable to parse request as JSON: #{e.message}"
+    $logger.error msg
+    Server.invalid_requests << {
```
kattrali, line 45: Does this have implications for failed requests from mapping file uploads, etc? Those requests aren't JSON.

CHANGELOG.md
```diff
@@ -3,9 +3,11 @@
 ## Enhancements
 
 - Add `--capabilities` option
-  [#176](https://github.com/bugsnag/maze-runner/pull/176)
```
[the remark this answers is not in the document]
twometresteve: Yeah, the diff here throws you off the scent, but the overall change was intentional and I believe correct now.  In an earlier PR I had entered the wrong PR number and subsequent PRs continued to get it wrong by adding one to the last changelog entry.  For the change here, #176 was a refactor and not of changelog interest, so I removed it in preference to #177 and 177,178 became 178,179.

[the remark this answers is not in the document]
kattrali: ahhh. makes sense.

lib/features/steps/request_assertion_steps.rb
twometresteve: Thanks.  I hate myself for having done it that way.

lib/features/support/servlet.rb
twometresteve: It shouldn't do.  The `JSON::ParserError` rescue block is redundant for mapping file uploads, as that arm of the `case` won't try to parse it as JSON.  The `StandardError` block seemed like a good thing to put in place for everything, so I thought the code was neatest by having both handlers together, even if one isn't strictly necessary for some of the logic.

kattrali: Ah! I see it now. lgtm.

kattrali (approved):

lib/features/support/servlet.rb
```diff
@@ -84,8 +92,14 @@ def log_request(request)
...
   end
 
+  # Checks the Bugsnag-Integrity header, if present, against the request and based on configuration.
+  # If the header is present, if the digest must be correct.  However, the header need only be present
+  # if configuration says so.
   def check_digest(request)
     header = request['Bugsnag-Integrity']
+    if MazeRunner.config.enforce_bugsnag_integrity
+      raise 'Bugsnag-Integrity header must be present according to MazeRunner.config.enforce_bugsnag_integrity'
+    end
```
imjoehaines, line 102: Won't this always raise if `enforce_bugsnag_integrity` is true?

twometresteve: You're right, good spot.  This PR is merged now, but I'll create a fresh one once you've had a chance to make any further comments.

imjoehaines: That was my only comment 😀

PR #181, opened 2020-11-25
Fix documented type in test_unhandled_state

imjoehaines:
The documented type for this is currently a Hash but it's actually an Integer
https://bugsnag.github.io/maze-runner/top-level-namespace.html#test_unhandled_state-instance_method

### DIFF
--- a/lib/features/steps/error_reporting_steps.rb
+++ b/lib/features/steps/error_reporting_steps.rb
@@ -288,7 +288,7 @@ end
 
 # Tests whether an event has the correct attributes we'd expect for un/handled events
 #
-# @param event [Hash] The body of the event
+# @param event [Integer] The index of the event
 # @param unhandled [Boolean] Whether the event is unhandled or handled
 # @param severity [String] Optional. An overwritten severity to look for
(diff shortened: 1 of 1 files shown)


twometresteve (approved):
There are lots of Rubocop violations in the file that could be resolved while we're there (if you have a friendly IDE), but happy to accept this as is!

imjoehaines:
> There are lots of Rubocop violations in the file that could be resolved while we're there (if you have a friendly IDE), but happy to accept this as is!

I fixed all the errors on the latest version (1.4.2); it'd be good to add Rubocop to the gemfile so everyone runs the same version

twometresteve (approved):
LGTM

twometresteve:
> > There are lots of Rubocop violations in the file that could be resolved while we're there (if you have a friendly IDE), but happy to accept this as is!
> 
> I fixed all the errors on the latest version (1.4.2); it'd be good to add Rubocop to the gemfile so everyone runs the same version

Nice!  And yes, agreed - to date I've just ben ironing out violations as I go, there were so many.  But it would be good to tighten the screw at some point and even build compliance into CI (without just having a whole load of TODOs).

PR #183, opened 2020-11-26
InteractiveCLI fixes for RN CLI

imjoehaines:
## Goal

Fixes some problems with the Interactive CLI / its tests, so that it's ready to use in anger

The tests were failing for two reasons:
- the docker image didn't have Node installed, so they couldn't run the scripts
- the command didn't exit until _after_ the stderr assertion, so stderr was never populated in time

If "the shell exited successfully" was added to the first step it would also fail due to the second reason above — the shell only exited _after_ the test finished. This has also been fixed, so tests can assert that they worked by waiting for the shell to exit

It's a bit unfortunate to have to explicitly let the shell exit, but I think this is unavoidable as the code doesn't know what the command should do, so it can't make assumptions about when it should be done

## Design

A new "I wait for the current shell to exit" step has been added, which closes STDIN and then waits for the command to stop, timing out after 15 seconds (if it times out then the step will fail)

I've also added "I wait for the shell to output {string} to stdout/stderr" steps, which wait for output rather than using "I wait X seconds". I did a bit of refactoring here to avoid having the same wait code in 3 places (the 2 new steps + "I wait to receive X requests")

## Tests

The current MR tests for this now pass

### DIFF
--- a/lib/features/steps/runner_steps.rb
+++ b/lib/features/steps/runner_steps.rb
@@ -124,6 +124,17 @@ When('I stop the current shell') do
   Runner.stop_interactive_session
 end
 
+# Attempts to wait for the currently running interactive shell to exit
+When('I wait for the current shell to exit') do
+  shell = Runner.get_interactive_session
+  result = shell.wait_for_exit
+
+  # The result should be the Thread object if it successfully stopped; if it
+  # timed out then 'nil' is returned
+  assert_false(result.nil?, 'The shell is still running when it should have exited')
+  assert_false(shell.running?, 'The shell is still running when it should have exited')
+end
+
 # Run a command on the shell
 #
 # @step_input command [String] The command to run on the shell
--- a/lib/features/steps/runner_steps.rb
+++ b/lib/features/steps/runner_steps.rb
@@ -163,7 +163,7 @@ Then('the shell has output {string} to stdout') do |expected_line|
   assert(match, "No output lines from #{current_shell.stdout_lines} matched #{expected_line}")
 end
 
-# Wait for a string to appear in the stdout logs
+# Wait for a string to appear in the stdout logs, timing out after MazeRunner.config.receive_requests_wait seconds.
 #
 # @step_input expected_line [String] The string present in stdout logs
 Then('I wait for the shell to output {string} to stdout') do |expected_line|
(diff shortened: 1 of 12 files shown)


lib/features/steps/runner_steps.rb
```diff
@@ -150,6 +163,20 @@
...
 
+# Wait for a string to appear in the stdout logs
+#
+# @step_input expected_line [String] The string present in stdout logs
+Then('I wait for the shell to output {string} to stdout') do |expected_line|
+  wait = Maze::Wait.new(timeout: MazeRunner.config.receive_requests_wait)
+  current_shell = Runner.get_interactive_session
+
+  success = wait.until do
+    current_shell.stdout_lines.any? { |line| line == expected_line }
```
twometresteve, line 174: I think it is fine for now, given there is a timeout, but we might need to introduce a separate step one day to give a specific timeout for longer running processes.

twometresteve: ...something along the lines of `I wait for up to {int} seconds for the shell...`.

lib/features/steps/runner_steps.rb
```diff
@@ -150,6 +163,20 @@
   assert(match, "No output lines from #{current_shell.stdout_lines} matched #{expected_line}")
 end
 
+# Wait for a string to appear in the stdout logs
```
twometresteve, line 166:
```suggestion
# Wait for a string to appear in the stdout logs, timing out after MazeRunner.config.receive_requests_wait seconds.
```

lib/features/steps/runner_steps.rb
imjoehaines: Yeah, if we need to build a whole RN app that'll probably be necessary. I might leave it out for now as it should be super quick to add and we might not end up doing that in the end

lib/wait.rb
```diff
@@ -0,0 +1,31 @@
...
+      @max_attempts = interval / timeout
+    end
+
+    # Wait until the given block succeeds (returns a truthy value) or the
+    # timeout is exceeded
+    #
+    # @return [Object] The last value returned by the block
+    def until(&block)
+      success = false
+      attempts = 0
```
twometresteve, line 20: Need to increment `attempts`

imjoehaines: Yeah this class was just busted, fixed in https://github.com/bugsnag/maze-runner/pull/183/commits/40ec240405e39715ca49bd3eccca69d3c89f9e51

PR #184, opened 2020-11-27
Allow running docker commands interactively

imjoehaines:
This PR adds support for running docker commands via the InteractiveCLI. This allows running scripts that need user input within tests

While working on this it turned out that using Open3 for this doesn't quite work — the resulting shell doesn't count as a TTY and so some programs make assumptions about how it behaves which breaks expectations. Ruby has a `PTY` module which fixes this problem, but needed some work to transition to

I've also changed how the Runner works with regards to interactive sessions — previously the method to get the session and interact with it would start a new session if one was running. This breaks when a command crashes or exits before a step tries to assert against it, because the step will spin up a new session. Now a session is only created by `Runner.start_interactive_session` which must be called before any other step that interacts with the CLI (via the `I start a new shell` step)


test/fixtures/cli/features/CLI.feature
```diff
@@ -4,18 +4,16 @@ Feature: Interactive CLI support
     Given I start a new shell
     When I input "./features/fixtures/node_script" interactively
     And I wait for the shell to output "Starting node script" to stdout
-    Then the current stdout line is "Repeat 5 times "
+    Then the current stdout line contains "Repeat 5 times "
```
imjoehaines, line 7: These steps have to use "contains" because the readline library uses ANSI escape codes to manipulate the cursor now that it thinks we're in a real TTY

i.e. this step could also be `the current stdout line is "\e[1G\e[0JRepeat 5 times \e[16G"` but that seemed worse than doing a substring match

twometresteve (approved):
LGTM

PR #185, opened 2020-11-27
Add steps for interactive shells

twometresteve:
## Goal

Adds a collection of steps for working with interactive shells.

## Design

This is an integration branch.  The only change on top of those already reviewed is that for the entry in `CHANGELOG.md`.


lib/features/support/interactive_cli.rb
```diff
@@ -0,0 +1,162 @@
...
+      end
+    end
+  ensure
+    stderr_reader.close unless stderr_reader.closed?
+    stderr_writer.close unless stderr_writer.closed?
+  end
+
+  # Strips whitespace from shell lines, can be used to sanitize further if required
+  def format_line(line)
+    line.strip
```
imjoehaines, line 160: Some commands are expected to output with whitespace so I'm not 100% sure this is a good idea; changing it might make things more annoying though

e.g. `git status --porcelain` should output like this:

```
 M some/modified/file.rb
?? a/new/file.rb
```

i.e. modified files are ` M` rather than just `M`, but our assertions have to omit the space or they don't match

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
 - Enforce presence of `Bugsnag-Integrity` based on MazeRunner.config.enforce_bugsnag_integrity
   [#179](https://github.com/bugsnag/maze-runner/pull/179)
   [#180](https://github.com/bugsnag/maze-runner/pull/180)
+- Add steps for running interaction shells
```
imjoehaines, line 12:
```suggestion
- Add steps for running interactive shells
```

imjoehaines (approved):

PR #187, opened 2020-12-01
Allow checking the exit code of multiple commands

imjoehaines:
## Goal

Currently our matching on stdout is done against any line — this is a good thing because otherwise there are race conditions and the potential for ordering changes to cause tests to flake

However, we also use the output of `echo $?` (and similar) to check the exit codes of commands. This means that any command that exits successfully will then allow all other commands to pass the "exited successfully" step, because the success output is already in the stdout logs

This PR is a fix specifically for that problem by adding a UUID to the exit code output, so each time the step runs it expects something different & unique

We also have an issue with output generally where any line of text in stdout will pass, even if another command is run afterwards. We may want to clear stdout before running a command, but maybe that could also cause issues? A manual "I clear stdout" command might be more robust, but a lot more error prone & annoying to use. This PR unblocks my immediate problem and I haven't had an issue with conflicting stdout otherwise, so I haven't attempted to fix that yet

### DIFF
--- a/lib/features/steps/runner_steps.rb
+++ b/lib/features/steps/runner_steps.rb
@@ -1,3 +1,4 @@
+require 'securerandom'
 require_relative '../../wait'
 
 # @!group Runner steps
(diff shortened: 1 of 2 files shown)


twometresteve (approved):
LGTM

PR #188, opened 2020-12-08
[PLAT-5475] Allow environment variables be used to configure specific options

Cawllec:
## Goal

Enables environment variables to be used when configuring:
- Device farm username -> `MAZE_DEVICE_FARM_USERNAME`
- Device farm access key -> `MAZE_DEVICE_FARM_ACCESS_KEY`
- BrowserStack local path -> `MAZE_BS_LOCAL`
- Appium server URL -> `MAZE_APPIUM_SERVER`
- Apple team ID -> `MAZE_APPLE_TEAM_ID`
- Apple UDID -> `MAZE_UDID`

These variables will be lower priority than actual configuration flags, which will take precedence.

In addition minor formatting changes have been made for readability.

## Tests

Unit tests for the option parser in general have been added, verifying:
- Defaults
- Flag configuration
- Short-flag configuration
- Environment variable configuration
- Configuration priority

### DIFF
--- a/lib/options/option_parser.rb
+++ b/lib/options/option_parser.rb
@@ -54,21 +54,21 @@ module Maze
 
           # BrowserStack-only options
           opt Option::BS_LOCAL,
-              '(BS only) Path to the BrowserStackLocal binary',
+              '(BS only) Path to the BrowserStackLocal binary. MAZE_BS_LOCAL env var or "/BrowserStackLocal" by default',
               short: :none,
               type: :string,
-              default: '/BrowserStackLocal'
+              default: ENV['MAZE_BS_LOCAL'] || '/BrowserStackLocal'
           opt Option::BS_DEVICE,
               'BrowserStack device to use (a key of Devices.DEVICE_HASH)',
               short: :none,
               type: :string
           opt Option::USERNAME,
-              'Device farm username',
+              'Device farm username. MAZE_DEVICE_FARM_USERNAME env var by default',
               short: '-u',
               type: :string,
               default: ENV['MAZE_DEVICE_FARM_USERNAME']
           opt Option::ACCESS_KEY,
-              'Device farm access key',
+              'Device farm access key. MAZE_DEVICE_FARM_ACCESS_KEY env var by default',
               short: '-p',
               type: :string,
               default: ENV['MAZE_DEVICE_FARM_ACCESS_KEY']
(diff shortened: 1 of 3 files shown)


twometresteve:
Looks good and just testing out now.... a few little comments/ideas to start with:
- Don't for the CHANGELOG.md
- Could we add an environment option for `--bs-local`?  I find it quite annoying having to set that when running locally.
- There's also `--appium-server`, although the default for that is typically ok.
- Could we update the pipeline so that one of the maze runner steps passes in the username/access key as an environment variable rather than command line option?  Just to give a bit more confidence in it all fitting in place as we expect.

lib/options/option_parser.rb
```diff
@@ -54,8 +62,16 @@ def parse(args)
               'BrowserStack device to use (a key of Devices.DEVICE_HASH)',
               short: :none,
               type: :string
-          opt Option::USERNAME, 'Device farm username', short: '-u', type: :string
-          opt Option::ACCESS_KEY, 'Device farm access key', short: '-p', type: :string
+          opt Option::USERNAME,
+              'Device farm username',
+              short: '-u',
+              type: :string,
+              default: ENV['MAZE_DEVICE_FARM_USERNAME']
```
twometresteve, line 69: This does work, but the help doesn't tell the user that they can use the environment variable.  We may need to explicitly change the descriptions to something like "Device farm username (defaults to MAZE_DEVICE_FARM_USERNAME)".

Cawllec:
Re: Changelog, has 3.6.0 been released?  If it has, I've got the changelog ready to go.

Also, preliminary docs [here](https://bugsnag.atlassian.net/wiki/spaces/EN/pages/1792049160/Maze+Runner+Configuration)

lib/options/option_parser.rb
Cawllec: Added some extra detail to the help strings.

twometresteve:
> here

No 3.6.0 isn't released yet, but I'm happy to roll this in with it - so you're safe to add here.  

In fact, if you put a release date of 2020/12/09 in at the same time it would save me a PR.  🙂

twometresteve (approved):
LGTM, if you can just add the changelog entry and set the release date for me.

twometresteve (approved):
LGTM

PR #190, opened 2020-12-16
Clear request arrays after each scenario

twometresteve:
## Goal

Clear the request arrays that we assert against after each scenario completes, rather than beforehand.

## Design

Until now our device-based test fixtures have not started Bugsnag until a button is actively pressed.  That has changed with the birth of the RN CLI, such that session requests are sent as soon as the app is first opened - which occurs before the scenario itself actually starts.  

## Changeset

Clearing of arrays moved to the `After`hook and clearly comments as to why.

I've also taken the opportunity to add a new Cucumber step for checking session requests for React Native, which could have come from either the Cocoa or Android notifier.

## Tests

Regression is mitigated by Maze Runner CI and I have tested the new step locally with a RN app on both Android and iOS.

### DIFF
--- a/lib/features/hooks/hooks.rb
+++ b/lib/features/hooks/hooks.rb
@@ -70,11 +70,6 @@ end
 Before do |scenario|
   STDOUT.puts "--- Scenario: #{scenario.name}"
 
-  Runner.environment.clear
-  Store.values.clear
-  # Request arrays are cleared after the scenario as some test fixtures
-  # will send requests on startup - before this hook takes place.
-
   MazeRunner.driver.start_driver if MazeRunner.config.farm != :none && MazeRunner.config.appium_session_isolation
 
   # Launch the app on macOS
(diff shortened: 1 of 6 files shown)


tomlongridge (approved):

lib/features/hooks/hooks.rb
```diff
@@ -148,6 +148,11 @@
     msg = "#{Server.invalid_requests.length} invalid request(s) received during scenario"
     scenario.fail msg
   end
+
+  # Request arrays are cleared after the scenario to allow requests to be registered
```
imjoehaines, line 152: Should we just clear everything here so we don't need this comment?

imjoehaines (approved):

lib/features/hooks/hooks.rb
twometresteve: I think I was being overly averse to change risk and it would make a lot more sense for it all to be together.  In fact, having looked again I've realised that I put it in the wrong `After` hook, which would cause problems if the scenario was forced to fail.

imjoehaines (approved):

PR #191, opened 2020-12-16
Display version at startup

twometresteve:
## Goal

Display the Maze Runner version when the mock server starts.

## Design

Our CI instances include localized Docker registries, so that just after a release of Maze Runner we can often we left in doubt as to whether the old `latest` image is still being used or if something has gone horribly wrong.  This should avoid that doubt.

## Changeset

Additional log in server class.

## Tests

Covered by CI and basic local testing that options such as `--help` and `--version` are not adversely affected.

### DIFF
--- a/lib/version.rb
+++ b/lib/version.rb
@@ -1,3 +1,3 @@
-module BugsnagMazeRunner
+module Maze
   VERSION = '3.6.1'.freeze
 end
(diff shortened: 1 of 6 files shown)


lib/version.rb
```diff
@@ -1,3 +1,3 @@
-module BugsnagMazeRunner
+module Maze
```
twometresteve, line 1: This just bring it in line with an evolving convention elsewhere in the codebase.

fractalwrench (approved):
LGTM

PR #193, opened 2021-01-04
Endpoint separation for JS-Browser and add Selenium web driver

twometresteve:
## Goal

Updates the Cucumber steps in order that the JS-Browser tests will run against separate notify/sessions endpoints.  I've also taken the opportunity to add the Selenium web driver code, currently present in the bugsnag-js repo.

## Design

The changes here follow the existing patterns for both the separation of endpoints and addition of a new `--browser` option to the Maze Runner command line.

Due to the increased use of Maze Runner outside of Docker (and the increased chance of clashes in the default namespace), I have also moved classes into their the `Maze` namespace where appropriate.

## Changeset

- Addition of Selenium webdriver code currently present in bugsnag-js
- New `--browser` option
- Updated/added Cucumber steps to support the separation of endpoints.

## Tests

Largely covered by CI and having successfully developed the necessary changes to JS-Browser on a branch.  However, it would be a fair review comment to ask for at least one e2e test of the new Browser capability to be added.

### DIFF
--- a/lib/features/hooks/hooks.rb
+++ b/lib/features/hooks/hooks.rb
@@ -144,7 +144,7 @@ After do |scenario|
     end
   end
 
-  next if MazeRunner.config.farm == :none
+  next if MazeRunner.config.farm == :none || !MazeRunner.config.bs_browser.nil?
 
   if MazeRunner.config.appium_session_isolation
     MazeRunner.driver.driver_quit
--- a/lib/features/hooks/hooks.rb
+++ b/lib/features/hooks/hooks.rb
@@ -3,11 +3,12 @@
 require 'cucumber'
 require 'json'
 require 'securerandom'
+require 'selenium-webdriver'
 
 AfterConfiguration do |_cucumber_config|
 
   # Start mock server
-  Server.start
+  Maze::Server.start
   config = MazeRunner.config
   next if config.farm == :none
 
@@ -18,6 +19,7 @@ AfterConfiguration do |_cucumber_config|
   if config.farm == :bs
     tunnel_id = SecureRandom.uuid
     if config.bs_device
+      # BrowserStack device
       config.capabilities = Capabilities.for_browser_stack_device config.bs_device,
                                                                   tunnel_id,
                                                                   config.appium_version,
@@ -26,7 +28,9 @@ AfterConfiguration do |_cucumber_config|
       config.app = BrowserStackUtils.upload_app config.username,
                                                 config.access_key,
                                                 config.app
+      config.capabilities['app'] = config.app
     else
+      # BrowserStack browser
       config.capabilities = Capabilities.for_browser_stack_browser config.bs_browser,
                                                                    tunnel_id,
                                                                    config.capabilities_option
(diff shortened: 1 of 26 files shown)


lib/features/hooks/hooks.rb
```diff
@@ -108,28 +123,28 @@
...
+      $logger.info "#{Maze::Server.errors.size} errors were received:"
+      Maze::Server.errors.all.each.with_index(1) do |request, number|
         $logger.info "Request #{number}:"
         LogUtil.log_hash(Logger::Severity::INFO, request)
       end
     end
   end
 
-  next if MazeRunner.config.farm == :none
+  next if MazeRunner.config.farm == :none || !MazeRunner.config.bs_browser.nil?
```
imjoehaines, line 147: Would removing this line & changing the `else` on 154 to `elsif MazeRunner.config.farm != :none` be clearer?

lib/features/support/selenium_driver.rb
```diff
@@ -0,0 +1,48 @@
...
+    @driver.execute_script('return true')
+  rescue Selenium::WebDriver::Error::UnsupportedOperationError
+    false
+  end
+
+  # check if the browser supports local storage, e.g. safari 10 on browserstack
+  # does not have working local storage
+  def local_storage?
+    # Assume we can use local storage if we aren't able to verify by running JavaScript
+    return true unless MazeRunner.driver.javascript?
```
imjoehaines, line 35:
```suggestion
    return true unless javascript?
```

lib/options/option_validator.rb
```diff
@@ -32,25 +33,34 @@ def validate(options)
...
 
       # Device
+      bs_browser = options[Maze::Option::BS_BROWSER]
       bs_device = options[Maze::Option::BS_DEVICE]
-      if bs_device.nil?
-        errors << "--#{Maze::Option::BS_DEVICE} must be specified"
-      else
+      if bs_browser.nil? && bs_device.nil?
+        errors << "Either --#{Maze::Option::BS_BROWSER} or --#{Maze::Option::BS_DEVICE} must be specified"
+      elsif !bs_browser.nil?
```
imjoehaines, line 45: It might be a bit clearer to test the values themselves; only `nil` & `false` are falsy

```suggestion
      elsif bs_browser
```

lib/options/option_validator.rb
```diff
@@ -32,25 +33,34 @@ def validate(options)
...
+
+        browsers = YAML.safe_load(File.read("#{__dir__}/../features/support/capabilities/browsers.yml"))
+
+        unless browsers.include? bs_browser
+          browser_list = browsers.keys.join ', '
+          errors << "Browser type '#{bs_browser}' unknown on BrowserStack.  Must be one of: #{browser_list}."
+        end
+      elsif !bs_device.nil?
```
imjoehaines, line 53:
```suggestion
      elsif bs_device
```

lib/features/hooks/hooks.rb
```diff
@@ -60,7 +76,6 @@
       $logger.info "BrowserStack session(s): #{url}"
     end
   end
-  MazeRunner.driver.start_driver unless config.appium_session_isolation
 
   # Call any blocks registered by the client
   MazeRunner.hooks.call_after_configuration config
```
imjoehaines, line 81: This isn't related to the PR but I think the `next if config.farm == :none` on line 13 will mean these `after_configuration` never get called

lib/features/steps/browser_steps.rb
```diff
@@ -0,0 +1,105 @@
...
+
+When('I navigate to the URL {string}') do |path|
+  $logger.debug "Navigating to: #{path}"
+  MazeRunner.driver.navigate.to path
+end
+
+When('the test should run in this browser') do
+  wait = Selenium::WebDriver::Wait.new(timeout: 10)
+  wait.until {
+    MazeRunner.driver.find_element(id: 'bugsnag-test-should-run') &&
```
imjoehaines, line 11: Do we have some docs about these IDs? I don't think I've seen this before

imjoehaines (approved):

lib/features/steps/browser_steps.rb
twometresteve: Hmm, I think I need to take another look at this.  I've dropped a number of steps in from the JS repo quite blindly, but a few are quite JS browser specific and I think should move back.  The `bugsnag-test-should-run` element is present in the JS test fixtures, so should make more sense once the step is back in the same repo.

twometresteve: Two steps now moved back to the JS repo.  I think there's some further refinement that can be done in time to abstract the JS repo away from Selenium, but I think it's ok for the time being.

lib/features/hooks/hooks.rb
twometresteve: Unfortunately that won't work, because the farm for Browsers is still `:bs` and the selenium driver doesn't have the `reset_with_timeout` method.  But it can certainly be clearer and I _think_ a similar change can be made if we test `MazeRunner.config.bs_device`on the `elsif` instead.

lib/options/option_validator.rb
twometresteve: Changed as suggested.

lib/options/option_validator.rb
twometresteve: Changed.

lib/features/support/selenium_driver.rb
twometresteve: Yep, changed.

lib/features/hooks/hooks.rb
twometresteve: PLAT-5666 raised for this.

PR #197, opened 2021-01-08
Strip ansi escape codes from CLI stdout/stderr

imjoehaines:
## Goal

Strip ansi escape codes from CLI stdout/stderr, which seems to help reduce flakes in the RN CLI pipeline


twometresteve (approved):
LGTM

snmaynard:
The boring gem has no license. Do we need it?

PR #201, opened 2021-01-17
Step corrections for separation of endpoints

twometresteve:
## Goal

Corrects some errors that came to light as part of updating the Android pipeline to use what is now a candidate for MazeRunner v4.

## Changeset

Cucumber step corrections and refactoring.  Also updated the Android notifier that the end-to-end tests use in order that the `Bugnsag-Integrity` header step can be brought in.

## Tests

Covered by CI combined with Android PR https://github.com/bugsnag/bugsnag-android/pull/1080, which uses it.

### DIFF
--- a/lib/features/steps/error_reporting_steps.rb
+++ b/lib/features/steps/error_reporting_steps.rb
@@ -7,7 +7,7 @@
 #
 # @step_input version [String] The payload version expected
 # @step_input name [String] The expected name of the notifier
-Then('the request is valid for the error reporting API version {string} for the {string} notifier') do |version, name|
+Then('the error is valid for the error reporting API version {string} for the {string} notifier') do |version, name|
   steps %(
     Then the error "Bugsnag-Api-Key" header equals "#{$api_key}"
     And the error payload field "apiKey" equals "#{$api_key}"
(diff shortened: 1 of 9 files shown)


lib/features/steps/error_reporting_steps.rb
```diff
@@ -169,6 +182,19 @@
   step "the error payload field \"events.0.#{field}\" matches the JSON fixture in \"#{fixture_path}\""
 end
 
+Then('the event {string} string is empty') do |keypath|
```
imjoehaines, line 185: Is this supposed to be `{string} string` or just `{string}`?

lib/features/steps/request_assertion_steps.rb
```diff
@@ -63,8 +56,14 @@ def assert_received_requests(request_count, list, list_name)
 # @step_input request_type [String] The type of request (error, session, etc)
```
imjoehaines, line 56: I think it's worth calling out that you can also use "requests" in the docs as it's the only step that works this way. Maybe even have it as a separate step entirely?

```suggestion
# @step_input request_type [String] The type of request (error, session, etc) or "requests" to assert that the test Server hasn't received any requests at all
```

imjoehaines (approved):

lib/features/steps/error_reporting_steps.rb
twometresteve: I think the wording could possibly be improved, but it is deliberate and what Android has always used (this change absorbs the existing step into MazeRunner).

lib/features/steps/request_assertion_steps.rb
twometresteve: Ah, this is an interesting one.  I originally had it as a separate step and Cucumber barfed at it being ambiguous, which makes sense.  I'll add a note to the docs.

PR #202, opened 2021-01-18
[PLAT-5750] Add appium class: LocalAppiumServer

Cawllec:
## Goal

Adds a basic class to control the usage of a local appium server.

This is added to the hooks section so an appium server is attempted to be automatically started if `--farm=local` is present in maze-runners kwargs.  This has a few conditions:
- If the appium server has already been successfully started any further calls of `start` will result in a quiet exit and no further action.
- The `appium` node module is intended to be used.  If this isn't present a warning will be logged and the appium server won't be started, but the run will continue (potentially resulting in an error in the appium driver)
- The default address and port built in are `0.0.0.0` and `4723`, the same as the default appium address and port.  These will be automatically overwritten by the values parsed from the `APPIUM_SERVER` option
- If something is blocking the desired port a warning will be logged, but the run will continue to allow an externally control appium server to be used.  This may result in an error when the appium driver starts if this external appium server isn't accepting connections.
- An `at_exit` block is automatically added after the server thread is started to interrupt the appium process with `SIGINT` and wait for the associated thread to close before exiting.

TODO:
- ~Add check that `appium` is present on the command line~
- ~Add separate logging to file~ - Added as a later task/ticket
- ~Comments/DocStrings~
- ~Add e2e test for MacOS~ - Added as a later task/ticket.  Using the bugsnag-cocoa `tests/macos-test-pipeline` branch for testing currently
- ~Add further unit tests (when the above are added)~


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 4.4.0 - TBD
+
+## Enhancements
+
+- Adds in-built appium server control for local testing [#202](https://github.com/bugsnag/maze-runner/pull/202)
+
```
twometresteve, line 6: I've got a changed lined up too, which I was looking at a 4.3.1 for, but looks like we might be able to roll out together.

twometresteve: In fact I've just released 4.3.1 for the benefit of Android - just rebase onto that.

lib/maze/local_appium_server.rb
```diff
@@ -0,0 +1,95 @@
...
+      # Checks whether the server is running, as indicated by the @pid presence
+      #
+      # @return [Boolean] Whether the local appium server is running
+      def running
+        @pid ? true : false
+      end
+
+      # Stops the appium server, if running, using SIGINT for correct shutdown
+      def stop
+        return unless running
```
twometresteve, line 65: nit: Rubocop wants a blank line after this

lib/maze/local_appium_server.rb
```diff
@@ -0,0 +1,95 @@
+# frozen_string_literal: true
+
+require 'pty'
+require 'logger'
+
+module Maze
+  # Basic shell that runs an Appium server on a separate thread
+  class LocalAppiumServer
```
twometresteve, line 8: think we could lose "local" from the class and file name.

lib/maze/local_appium_server.rb
```diff
@@ -0,0 +1,95 @@
...
+        false
+      end
+
+      # Checks if the given port is already in use
+      #
+      # @param port [String] The port that should be available
+      #
+      # @return [Boolean] Whether something is running on the given port
+      def appium_port_available?(port)
+        `netstat -vanp tcp | grep #{port}`.empty?
```
twometresteve, line 91: A lot of information comes out on that `netstat` command - the `grep` could easily yield a false match.

lib/maze/local_appium_server.rb
```diff
@@ -0,0 +1,95 @@
...
+            @pid = pid
+            $logger.debug("Appium:#{@pid}") { 'Appium server started' }
+            stdout.each do |line|
+              $logger.debug("Appium:#{@pid}") { line }
+            end
+          end
+        end
+        at_exit do
+          stop
+        end
```
twometresteve, line 50: Wouldn't this be better in `hooks.rb` now?

lib/maze/local_appium_server.rb
twometresteve: How about making a GET request to `http://0.0.0.0:4723/wd/hub/sessions` instead? (Credit: SO).

Cawllec: Following in person conversation where we initially decided on a different approach I believe keeping this style of check except with slightly more specific command would be better:
`netstat -vanp tcp | awk '{ print $4 }' | grep #{port}`

This would only look for the port address in the relevant netstat column (avoiding the issue above for the vast majority of cases).  In addition it would still fail if something is blocking the port which _isn't_ appium.

The face-to-face approach of "try and handle failure" doesn't feel appropriate due to the appium server launch:
- Not being in the main thread of the application some much more difficult to handle
- "Soft-failing" when the appium server is launched while the port is already busy which means an error response won't be present straight away

Cawllec:
Changed the running method to check if the appium server thread is present and alive instead of the PID, as there may be cases where the appium server exits (resulting in the thread exiting) but the server will look alive due to the PID being present.

lib/maze/local_appium_server.rb
twometresteve: Makes sense, happy to go with that.  Without wanting to overthink it too much, would it make sense to tweak the grep to `grep '\.#{port}$'` - avoiding false hits on (the arguably unlikely) 14723 etc?

twometresteve (approved):
LGTM

twometresteve (approved):

PR #208, opened 2021-01-25
Make platform-dependent tests independent of payload type

Cawllec:
## Goal

Generalises the payload type used when making platform-dependent checks, allowing platform-dependent assertions to be made against session payloads.


twometresteve (approved):
LGTM

twometresteve (approved):

PR #209, opened 2021-01-26
Sort received requests by Bugsnag-Sent-At header

twometresteve:
## Goal

Ensures the batch of requests captured by the `I wait to receive {int} {word}` step are sorted by the `Bugsnag-Sent-At` header, if it is present on all requests.  If requests are sent very close together (perhaps <0.1s), they can end up being stored out of order, resulting in test flakes.

## Design

It is important that sorting of requests does not intrude upon:
- requests previously captured and already processed
- requests that arrive due to subsequent actions, whilst the batch is still being processed.
The unit tests added are designed to ensure this.

## Changeset

- New method on the `RequestList` class to sort by header in-situ.  
- Plumbing into the Cucumber step
- Incidental pipeline change to reduce unnecessary noise (values are now set in the environment)

## Tests

Logic for sorting of the requests sub-list is covered by unit tests.  I've also run some Android tests locally against it that expect batches of requests to be received.  This shows that it expects the right form of data structure.

### DIFF
--- a/lib/features/steps/request_assertion_steps.rb
+++ b/lib/features/steps/request_assertion_steps.rb
@@ -30,6 +30,7 @@ def assert_received_requests(request_count, list, list_name)
   end
 
   assert_equal(request_count, list.size, "#{list.size} #{list_name} received")
+  list.sort_by_sent_at request_count
 end
 
 #
--- a/lib/features/steps/request_assertion_steps.rb
+++ b/lib/features/steps/request_assertion_steps.rb
@@ -30,7 +30,6 @@ def assert_received_requests(request_count, list, list_name)
   end
 
   assert_equal(request_count, list.size, "#{list.size} #{list_name} received")
-  list.sort_by_sent_at! request_count
 end
 
 #
--- a/lib/features/steps/request_assertion_steps.rb
+++ b/lib/features/steps/request_assertion_steps.rb
@@ -30,7 +30,7 @@ def assert_received_requests(request_count, list, list_name)
   end
 
   assert_equal(request_count, list.size, "#{list.size} #{list_name} received")
-  list.sort_by_sent_at request_count
+  list.sort_by_sent_at! request_count
 end
 
 #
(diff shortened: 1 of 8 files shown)


lib/maze/request_list.rb
```diff
@@ -64,5 +64,19 @@ def clear
       @current = 0
       @count = 0
     end
+
+    # Sorts the first `count` elements of the list by the Bugsnag-Sent-At header, if present in all of those elements
+    def sort_by_sent_at!(count)
```
imjoehaines, line 69: Is there the potential for other threads to mess with the list while this is happening?

My knowledge is a bit limited, but I think only the server runs on a separate thread and that only appends. so it should be OK ?

lib/maze/request_list.rb
```diff
@@ -64,5 +64,19 @@ def clear
...
+      return unless count > 1
+
+      header = 'Bugsnag-Sent-At'
+      sub_list = @requests[@current...@current + count]
+
+      return unless sub_list.all? { |r| !r[:request][header].nil? }
```
imjoehaines, line 75: nit: maybe this is a bit easier to read as "return if any element of the list has a `nil` sent-at header"

```suggestion
      return if sub_list.any? { |r| r[:request][header].nil? }
```

imjoehaines (approved):

lib/maze/request_list.rb
twometresteve: Yes, much better!

lib/maze/request_list.rb
Cawllec: I believe this is the case as aside from the server and CLI services like docker and the interactive commands maze-runner should be based around a single thread.

twometresteve: Yeah, it needed thinking about, but I think we're ok.  We need to account for the potential of more requests being added to the list, but only the MazeRunner (Cucumber) thread will move the current pointer on.  That's why the design includes the passing of the number of elements just received - it could break if it just sorted everything in the list every time.

lib/features/steps/request_assertion_steps.rb
```diff
@@ -42,13 +43,17 @@ def assert_received_requests(request_count, list, list_name)
...
+# If all expected requests are received and have the Bugsnag-Sent-At header, they
+# will be sorted by the header.
 #
 # @step_input request_type [String] The type of request (error, session, etc)
 # @step_input request_count [Integer] The amount of requests expected
 Then('I wait to receive {int} {word}') do |request_count, request_type|
-  assert_received_requests request_count, Maze::Server.list_for(request_type), request_type
+  list = Maze::Server.list_for(request_type)
+  assert_received_requests request_count, list, request_type
+  list.sort_by_sent_at! request_count
```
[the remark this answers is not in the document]
twometresteve: Ah, so it is... I've removed it from `assert_received_requests` as doing the sort their feels like a hidden side-effect.

PR #210, opened 2021-01-28
Nullable platform dependence

Cawllec:
## Goal

Adds `@null` and `@not_null` tags for use with platform-dependence steps.

This allows:
- Asserting that a value is present (with an unknown value) on a specific platform using `@not_null`
- Asserting that a value will not be present, or be null, on a specific platform using `@null`

## Tests

Added to the platform-dependence e2e tests.

### DIFF
--- a/lib/features/steps/app_automator_steps.rb
+++ b/lib/features/steps/app_automator_steps.rb
@@ -168,9 +168,9 @@ When('I clear and send the keys {string} to the element {string}') do |keys, ele
 end
 
 def get_expected_platform_value(platform_values)
-  raise('This step should only be used when running tests with Appium') if Maze.driver.nil?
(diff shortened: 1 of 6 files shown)


lib/features/steps/app_automator_steps.rb
```diff
@@ -164,7 +170,7 @@
 def get_expected_platform_value(platform_values)
   raise('This step should only be used when running tests with Appium') if Maze.driver.nil?
 
-  os = Maze.config.capabilities['os']
+  os = 'android'
```
twometresteve, line 173: Is this intentionally hard-coded?

Cawllec: Ah no, that was for testing, good spot!

twometresteve (approved):
LGTM

PR #211, opened 2021-01-29
[PLAT-PLAT-5827] Add root endpoint for connectivity checks

Cawllec:
## Goal

Adds an endpoint to the basic server on `/` that logs responds to all requests with a `200` status code.
This will be used for certain platforms that need to establish the server is reachable before attempting to run tests.

## Tests

Manually tested to ensure logging and status codes are correct

## Thoughts

- It may be useful to add more debug logging for received requests, but I'd be reluctant to do too much as this endpoint shouldn't typically be used.
- We could record a hit here and add steps for testing whether something has hit the endpoint, but this seems unnecessary for our purposes.

Changelog entry incoming

### DIFF
--- a/lib/maze/server.rb
+++ b/lib/maze/server.rb
@@ -93,6 +93,14 @@ module Maze
               Logger: $logger,
               AccessLog: []
             )
+
+            # Mount a block to respond to all requests with status:200
+            server.mount_proc '/' do |_request, response|
+              $logger.info 'Received request on server root, responding with 200'
+              response.header['Access-Control-Allow-Origin'] = '*'
+              response.status = 200
+            end
+
             # When adding more endpoints, be sure to update the 'I should receive no requests' step
             server.mount '/notify', Servlet, errors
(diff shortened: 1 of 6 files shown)


lib/maze/server.rb
```diff
@@ -93,6 +93,14 @@ def start
               Logger: $logger,
               AccessLog: []
             )
+
+            # Mount a block to respond to all requests with status:200
+            server.mount_proc '/' do |_request, response|
+              $logger.info 'Received request on server root, responding with 200'
+              response.header['Access-Control-Allow-Origin'] = '*'
+              response.status = 200
```
twometresteve, line 101: It's absolutely fine as it is, but would it be worth putting a "Hello from Maze Runner" or similar in the response body for a human reading it?

twometresteve (approved):
LGTM, just the changelog and version bump, as you noted.

twometresteve (approved):
LGTM

PR #213, opened 2021-02-03
Add build endpoint

Cawllec:
## Goal

Allows requests to be received on `/builds` and then tested using the payload steps with `build` as the request type

## Tests

Added basic tests to the payload helper e2e test set to verify requests are received and payload steps work correctly

### DIFF
--- a/lib/features/support/hooks.rb
+++ b/lib/features/support/hooks.rb
@@ -146,6 +146,17 @@ After do |scenario|
         Maze::LogUtil.log_hash(Logger::Severity::INFO, request)
       end
     end
+
+    if Maze::Server.builds.empty?
+      $logger.info 'No valid builds received'
+    else
+      count = Maze::Server.builds.size_all
+      $logger.info "#{count} builds were received:"
+      Maze::Server.builds.all.each.with_index(1) do |request, number|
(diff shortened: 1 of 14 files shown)


lib/features/support/hooks.rb
```diff
@@ -161,6 +172,7 @@
   # when a test fixture starts (which can be before the first Before scenario hook fires).
   Maze::Server.errors.clear
   Maze::Server.sessions.clear
+  Maze::Server.builds.clear
```
twometresteve, line 146: Ah, I had forgotten this on my branch where I'm adding logs!

lib/features/support/hooks.rb
```diff
@@ -161,11 +143,26 @@
...
   Maze::Server.invalid_requests.clear
   Maze::Runner.environment.clear
   Maze::Store.values.clear
 end
 
+def output_failed_requests(request_type)
```
twometresteve, line 152: Nice improvement factoring it into a method, but not that all received requests are now logged in v4, not just the failed ones.

twometresteve (commented):
Looks great, just the naming of the new method that could do with tweaking.

twometresteve (approved):
LGTM

PR #216, opened 2021-02-05
Logs endpoint fixes

twometresteve:
## Goal

Corrects some issues with the new `/logs` endpoint:
- Disables `Bugsnag-Integrity` header checking for all requests on the endpoint
- Fails gracefully if log message checks are attempts before waiting to receive them
- Prints logs received at the end of any failing scenarios

## Tests

Testing using the in-development Android test fixture on branch `PLAT-5905/ndk-on-error`.  The scenarios did not pass as-is (I assume because it is still in development), but I was able to demonstrate the receipt and checking of a log message from a real test fixture.

### DIFF
--- a/lib/maze/server.rb
+++ b/lib/maze/server.rb
@@ -124,7 +124,7 @@ module Maze
             server.mount '/notify', Servlet, errors
             server.mount '/sessions', Servlet, sessions
             server.mount '/builds', Servlet, builds
-            server.mount '/logs', Servlet, logs
+            server.mount '/logs', Servlet, logs, true
             server.start
           rescue StandardError => e
             $logger.warn "Failed to start mock server: #{e.message}"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 4.6.1 - 2021/02/08
+
+## Fixes
+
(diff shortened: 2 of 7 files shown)


Cawllec (approved):

lib/maze/server.rb
```diff
@@ -124,7 +124,7 @@ def start
             server.mount '/notify', Servlet, errors
             server.mount '/sessions', Servlet, sessions
             server.mount '/builds', Servlet, builds
```
fractalwrench, line 126: Should Bugsnag-Integrity checks be enabled for requests made to the /build endpoint?

fractalwrench (approved):
LGTM pending one query

lib/maze/server.rb
twometresteve: There are now a few levels of control now as to when the integrity header will be enforced.  This is the most blunt level of control before it flows down to the MR servlet (which will not enforce it for `multipart/form-data` requests).  Whilst I wouldn't currently expect any build requests to have the integrity header, I think it's reasonable to leave it to the downstream mechanisms to control because, in theory, the product change change and include the digest.  Log requests, however, are not a part of the product and can therefore be blanket excluded.

PR #219, opened 2021-02-09
Expand file paths for --app and --bs-local options

twometresteve:
## Goal

Expand file paths for command line options to allow the use of tildes in paths.

## Tests

Locally and unit tested.  I experienced a strange issue when trying to mock `File.exists` in the `processor` unit test where Mocha resulted in an infinite loop when trying to set up a stub or mock.  The same approach just worked in the `validator` test and I couldn't work it out after more than an hour looking at it.  I settled for refactoring the code, creating the `Helper.expand_path` method, to unblock myself - although this also turned out to be useful for handling the case where the path is `nil` (which is valid).


Cawllec (approved):
LGTM

PR #220, opened 2021-02-09
Enable the use of regexes when testing received STDOUT lines

Cawllec:
## Goal

This enables the use of regexes in testing stdout lines, which allows flexibility in certain circumstances (and mitigates flakes where the exact message is not important).

This updates the exit code steps to use a regex as well, meaning in the event the shell prompt is present in the stdout (which appears inconsistently) we are more resilient to it.

## Tests

Added scenario testing new regex steps to current CLI feature.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 4.8.0 - TBD
+
+## Enhancements
+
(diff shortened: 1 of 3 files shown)


imjoehaines (approved):

twometresteve (approved):
LGTM

PR #224, opened 2021-02-12
Add step to set response delay on next HTTP request

twometresteve:
## Goal

Adds a new step to allow tests to delay responses to an HTTP request.

## Design

Takes a similar approach to the existing status code functionality.

## Changeset

I've also added a new mechanism for running scripts with an interpreter - this was purely so that I could rename the script that's run to have an `.rb` extension and benefit from the language aware facilities of my IDE.

## Tests

Covered by CI, with a new test having been added..

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@
 
 - Add steps to delay the response to the next received HTTP request [#224](https://github.com/bugsnag/maze-runner/pull/224)
 
+# 4.9.1 - 2021/02/15
+
+## Fixes
+
+- Only expand `--app` option when uploading to BrowserStack [#225](https://github.com/bugsnag/maze-runner/pull/225)
+
 # 4.9.0 - 2021/02/12
 
 ## Enhancements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
-# 4.9.1 - 2021/02/15
+# 4.10.0 - 2021/02/16
 
-## Fixes
+## Enhancements
 
-- Only expand `--app` option when uploading to BrowserStack [#225](https://github.com/bugsnag/maze-runner/pull/225)
+- Add steps to delay the response to the next received HTTP request [#224](https://github.com/bugsnag/maze-runner/pull/224)
 
 # 4.9.0 - 2021/02/12
 
--- a/lib/maze/runner.rb
+++ b/lib/maze/runner.rb
@@ -61,16 +61,33 @@ module Maze
       # @return [Array] If blocking, the output and exit_status are returned
       def run_script(script_name, blocking: false, success_codes: [0])
         script_path = File.join(SCRIPT_PATH, script_name)
-        script_path = File.join(Dir.pwd, script_name) unless File.exists? script_path
+        script_path = File.join(Dir.pwd, script_name) unless File.exist? script_path
         if Gem.win_platform?
-          # windows does not support the shebang that we use in the scripts so it
+          # Windows does not support the shebang that we use in the scripts so it
           # needs to know how to execute the script. Passing `cmd /c` tells windows
-          # to use it's known file associations to execute this path. If ruby is
-          # installed on windows then it will know that `rb` files should be exceuted
-          # using ruby etc.
+          # to use its known file associations to execute this path. If Ruby is
+          # installed on Windows then it will know that `rb` files should be executed
+          # using Ruby etc.
           script_path = "cmd /c #{script_path}"
         end
         run_command(script_path, blocking: blocking, success_codes: success_codes)
+      end
+
+      # Runs a script with a given interpreter in the script directory indicated by the SCRIPT_PATH environment
+      # variable.
+      #
+      # @param script_name [String] The name of the script to run
+      # @param interpreter [String] The interpreter to use
+      # @param blocking [Boolean] Optional. Whether to wait for a return code before proceeding
+      # @param success_codes [Array] Optional. An array of integer codes which indicate the run was successful
+      #
+      # @return [Array] If blocking, the output and exit_status are returned
+      def run_interpreter(interpreter, script_name, blocking: false, success_codes: [0])
+        script_path = File.join(SCRIPT_PATH, script_name)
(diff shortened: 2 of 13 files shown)


lib/maze/server.rb
```diff
@@ -19,6 +19,9 @@ class << self
       # Allows overwriting of the server status code
       attr_writer :status_code
 
+      # Allows a delay before responding to HTTP requests to be set
```
imjoehaines, line 22:
```suggestion
      # Allows a delay in milliseconds before responding to HTTP requests to be set
```

imjoehaines (changes requested):

Cawllec (approved):

lib/maze/server.rb
```diff
@@ -35,6 +38,10 @@ def reset_status_code
         @reset_status_code ||= false
       end
 
+      def response_delay
```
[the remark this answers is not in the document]
twometresteve: Agreed, changed.

lib/maze/servlet.rb
```diff
@@ -50,6 +50,13 @@ def do_POST(request, response)
         }
       end
       @requests.add(hash)
+
+      # For the response, delaying if configured to do so
+      if Server.response_delay.positive?
+        $logger.info "Waiting #{Server.response_delay} milliseconds before responding"
+        sleep Server.response_delay / 1000.0
+        Server.response_delay = 0
```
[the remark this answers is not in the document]
twometresteve: I think that's a good question, the implementation is quite limited currently - perhaps too much so.  I'll discuss IRL with @nickdowell to decide how to proceed with it.

lib/maze/runner.rb
```diff
@@ -61,18 +61,35 @@ def run_command(cmd, blocking: true, success_codes: [0])
...
+      # Runs a script with a given interpreter in the script directory indicated by the SCRIPT_PATH environment
+      # variable.
+      #
+      # @param script_name [String] The name of the script to run
+      # @param interpreter [String] The interpreter to use
+      # @param blocking [Boolean] Optional. Whether to wait for a return code before proceeding
+      # @param success_codes [Array] Optional. An array of integer codes which indicate the run was successful
+      #
+      # @return [Array] If blocking, the output and exit_status are returned
+      def run_interpreter(interpreter, script_name, blocking: false, success_codes: [0])
```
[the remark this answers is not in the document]
twometresteve: I agree, but I was nervous of changing the existing implementation as it would be quite easy to break notifier tests that are reliant on it.  We have an upcoming task to review our support for running on Windows, which could well involve touching this area - so I've added a TODO to merge the new method in.  I'd rather spent a bit longer on it to make sure we don't break something in the process.

lib/maze/runner.rb
```diff
@@ -61,18 +61,36 @@ def run_command(cmd, blocking: true, success_codes: [0])
...
           script_path = "cmd /c #{script_path}"
         end
         run_command(script_path, blocking: blocking, success_codes: success_codes)
       end
 
+      # Runs a script with a given interpreter in the script directory indicated by the SCRIPT_PATH environment
+      # variable.
+      #
+      # @param script_name [String] The name of the script to run
+      # @param interpreter [String] The interpreter to use
```
imjoehaines, line 80: These are the wrong way around:

```suggestion
      # @param interpreter [String] The interpreter to use
      # @param script_name [String] The name of the script to run
```

lib/maze/runner.rb
```diff
@@ -61,18 +61,36 @@ def run_command(cmd, blocking: true, success_codes: [0])
...
+      # @param success_codes [Array] Optional. An array of integer codes which indicate the run was successful
+      #
+      # @return [Array] If blocking, the output and exit_status are returned
+      # TODO: Merge this into run_script above
+      def run_interpreter(interpreter, script_name, blocking: false, success_codes: [0])
+        script_path = File.join(SCRIPT_PATH, script_name)
+        script_path = File.join(Dir.pwd, script_name) unless File.exists? script_path
+        command = "#{interpreter} #{script_path}"
+
+        run_command(command, blocking: blocking, success_codes: success_codes)
```
imjoehaines, line 91: As this isn't using either keyword parameter directly, I think it makes sense to just forward them on for `run_command` to deal with. This saves having to duplicate the default values

```suggestion
      def run_interpreter(interpreter, script_name, **kwargs)
        script_path = File.join(SCRIPT_PATH, script_name)
        script_path = File.join(Dir.pwd, script_name) unless File.exists? script_path
        command = "#{interpreter} #{script_path}"

        run_command(command, **kwargs)
```

lib/maze/runner.rb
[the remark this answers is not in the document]
imjoehaines: I don't think it'd be possible to break anything with this change, but I've added a couple more comments if this method is staying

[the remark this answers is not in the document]
twometresteve: You're right, I was being overly paranoid - and having thought about it more I don't think any notifiers that use this routine have moved up to the latest version yet anyway, reducing any risk further.

lib/maze/runner.rb
twometresteve: Now removed altogether.

lib/maze/runner.rb
twometresteve: Method now removed.

lib/maze/servlet.rb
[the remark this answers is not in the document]
twometresteve: I've now brought the response delay steps in line with the existing steps for altering the status code.  If this doesn't fulfil the current use case, let's consider taking them both forward together from here.

CHANGELOG.md
```diff
@@ -1,8 +1,8 @@
-# 4.9.1 - 2021/02/15
+# 4.10.0 - 2021/02/16
```
imjoehaines, line 1: This looks like we're losing the 4.9.1 changelog

imjoehaines (approved):

CHANGELOG.md
twometresteve: Well spotted, seems I hashed the rebase.

PR #230, opened 2021-03-01
Output Application logs when running on MacOS

Cawllec:
## Goal

Makes Maze output a log file when testing a MacOS app with the `--farm=local` option set.

## Testing

The file and resulting upload can be seen here: https://buildkite.com/bugsnag/bugsnag-cocoa/builds/2101#575036dc-e9db-4caa-856f-e748db2d2b16/307-308

### DIFF
--- a/lib/features/support/hooks.rb
+++ b/lib/features/support/hooks.rb
@@ -209,6 +209,7 @@ at_exit do
     Maze::AppiumServer.stop if Maze::AppiumServer.running
   when :local
     # Acquire and output the logs for the current session
-    Maze::Runner.run_command("log show --predicate '(process == \"#{Maze.config.app}\")' --style syslog --start '#{Maze::Logger.instance.start_time}' > #{Maze.config.app}Output.log")
+    pp "Running an exit command"
+    Maze::Runner.run_command("log show --predicate '(process == \"#{Maze.config.app}\")' --style syslog --start '#{Maze::Logger.instance.start_time}' > #{Maze.config.app}.log")
   end
 end
--- a/lib/features/support/hooks.rb
+++ b/lib/features/support/hooks.rb
@@ -201,9 +201,14 @@ at_exit do
   # future test runs are from a clean slate.
   Maze::Docker.down_all_services
 
-  next if Maze.config.farm == :none
-
-  # Stop the Appium session and server
-  Maze.driver.driver_quit unless Maze.config.appium_session_isolation
-  Maze::AppiumServer.stop if Maze::AppiumServer.running
+  # Specific shutdown when a device farm is configured
+  case Maze.config.farm
+  when :none
+    # Stop the Appium session and server
+    Maze.driver.driver_quit unless Maze.config.appium_session_isolation
+    Maze::AppiumServer.stop if Maze::AppiumServer.running
+  when :local
+    # Acquire and output the logs for the current session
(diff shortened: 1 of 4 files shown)


lib/features/support/hooks.rb
```diff
@@ -201,9 +201,14 @@ def output_received_requests(request_type)
...
+  # Specific shutdown when a device farm is configured
+  case Maze.config.farm
+  when :none
+    # Stop the Appium session and server
+    Maze.driver.driver_quit unless Maze.config.appium_session_isolation
+    Maze::AppiumServer.stop if Maze::AppiumServer.running
+  when :local
+    # Acquire and output the logs for the current session
+    Maze::Runner.run_command("log show --predicate '(process == \"#{Maze.config.app}\")' --style syslog --start '#{Maze::Logger.instance.start_time}' > #{Maze.config.app}.log")
+  end
```
twometresteve, line 213: There are a couple of problems with the logic here.  Previously `farm == :none` would do nothing and `:local` also includes the use of iOS and Android devices held locally, which we won't want to capture system logs for.

`hooks.rb` probably needs a wholesale refactor at some point to abstract away the complications of different test environments, but for the time being I'd be tempted to just revert to what we had before and add the new line guarded by `if Maze.config.os == 'macos'` (mirroring what we do in the `Before` hook).

```suggestion
  next if Maze.config.farm == :none

  # Stop the Appium session and server
  Maze.driver.driver_quit unless Maze.config.appium_session_isolation
  Maze::AppiumServer.stop if Maze::AppiumServer.running

  if Maze.config.os == 'macos'
    # Acquire and output the logs for the current session
    Maze::Runner.run_command("log show --predicate '(process == \"#{Maze.config.app}\")' --style syslog --start '#{Maze::Logger.instance.start_time}' > #{Maze.config.app}.log")
  end
```

lib/maze/logger.rb
```diff
@@ -2,11 +2,15 @@
...
 require 'logger'
 require 'singleton'
+require 'time'
 
 # Logger classes
 module Maze
   # A logger, with level configured according to the environment
   class Logger < Logger
+    # @return [String] A timestamp of when the test run started, using the Logger initialization as a basis
+    attr_reader :start_time
```
twometresteve, line 12: Although we've added this to support the extraction of logs, it's not related to the `Maze::Logger` class itself.  I think `maze.rb` would be a better place for the new state, just like we have `Maze.driver`.

twometresteve (commented):
The contents of the extracted log looks strange to me.  The Buildkite step linked to ran from 19.19:56 to 19.20:33 on Feb 28th, but the contents of the log runs 11.20:11 to 11.20:29.  Has the time calculation gone awry?

Cawllec:
Regarding the time calculation, that's because the server is hosted in a different time zone to our buildkite settings.  As it's US west is appears to be 8 hours behind.

twometresteve:
> Regarding the time calculation, that's because the server is hosted in a different time zone to our buildkite settings. As it's US west is appears to be 8 hours behind.

Ah ok, I wondered if timezones could be at play, but I couldn't get my head around why.  I also tricked myself into thinking there should have been more in the logs, but upon re-inspection it does look right ("Running scenario:" mentioned twice).

lib/maze/logger.rb
```diff
@@ -2,11 +2,13 @@
 
 require 'logger'
 require 'singleton'
+require 'time'
```
twometresteve, line 5: Can now be removed.

twometresteve (approved):
LGTM, I'm taking another look at the servers that are causing the build step to fail.

twometresteve (approved):

PR #231, opened 2021-03-01
Add --bind-address and --port options

twometresteve:
## Goal

Adds options to control the bind address and port of the mock server.

## Tests

Covered by unit tests combined with manual testing to confirm that the bind address and port options take effect correctly, accessing the mock server from a separate machine.

### DIFF
--- a/bin/maze-runner
+++ b/bin/maze-runner
@@ -31,23 +31,25 @@ class MazeRunnerEntry
   # Removes Maze Runner specific args from the array, as these will cause Cucumber to error.
   def remove_maze_runner_args
     remove = [
-      Maze::Option::SEPARATE_SESSIONS,
-      Maze::Option::APP,
-      Maze::Option::BS_LOCAL,
-      Maze::Option::BS_DEVICE,
-      Maze::Option::BS_BROWSER,
-      Maze::Option::OS,
-      Maze::Option::OS_VERSION,
-      Maze::Option::FARM,
-      Maze::Option::USERNAME,
       Maze::Option::ACCESS_KEY,
+      Maze::Option::APPLE_TEAM_ID,
       Maze::Option::APPIUM_SERVER,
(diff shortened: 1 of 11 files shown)


bin/maze-runner
```diff
@@ -31,23 +31,25 @@ class MazeRunnerEntry
   # Removes Maze Runner specific args from the array, as these will cause Cucumber to error.
   def remove_maze_runner_args
     remove = [
```
imjoehaines, line 33: Not something to do now, but I think this could be replaced with `remove = Maze::Option.constants`

lib/maze/option/processor.rb
```diff
@@ -12,6 +12,8 @@ class << self
         # @param config [Configuration] MazeRunner configuration to populate
         # @param options [Hash] Parsed command line options
         def populate(config, options)
+          config.bind_address = options[Maze::Option::BIND_ADDRESS]
+          config.port = options[Maze::Option::PORT]
```
imjoehaines, line 16: Is it worth forcing this to be an integer? I guess WEBrick will crash if it's not

```suggestion
          config.port = Integer(options[Maze::Option::PORT])
```

imjoehaines (approved):

bin/maze-runner
twometresteve: Updated on the next PR.

lib/maze/option/processor.rb
twometresteve: I don't think we need to.  Thanks to the default on the option (an integer), Optimist treats it as a must-be integer option and rejects anything else:

```
$ bundle exec maze-runner --port=asd
Optimist::CommandlineError: option 'port' needs an integer
```

imjoehaines: fancy ✨

PR #234, opened 2021-03-03
Do not set explicit default for —bind-address

twometresteve:
## Goal

Do not set a default for the recently introduced `--bind-address` option, allowing `WEBrick::HTTPServer` to follow its default behaviour, which has generally served us well.

## Design

Setting a default of `127.0.0.1` when introducing the option caused some areas to regress (JS Browser and Node testing).  Making this change should remove the need to remedial configuration in those pipelines.

## Tests

Sufficiently covered by unit and e2e tests to merge to `master`.  Further testing against the JS repo will be performed before release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# TDB
+
+## Fixes
+
+- Do not set explicit default for mock server bind address [#234](https://github.com/bugsnag/maze-runner/pull/234)
+
 # 4.11.3 - 2021/03/03
 
 ## Fixes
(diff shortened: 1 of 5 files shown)


imjoehaines (commented):
It'd be good to add a unit test for this too; all the existing tests add the `BindAddress` option

twometresteve:
> It'd be good to add a unit test for this too; all the existing tests add the `BindAddress` option

Added, together with checks on a few other default values as appropriate.

imjoehaines (approved):

PR #235, opened 2021-03-03
Minors fixes for Node V4

Cawllec:
## Goal

A couple of extra fixes required to run Node e2e tests using Maze-runner v4:
- Ensures `Maze::Network` is imported and available to steps that require it
- Updates the endpoint > env-var step to allow environment variables to be set to the correct endpoints


twometresteve (approved):
LGTM

PR #237, opened 2021-03-04
Loosen requirements on Lambda responses

imjoehaines:
## Goal

Currently it's impossible to test a Lambda function that returns no response or returns a response body which is not JSON

A Lambda can avoid responding by forcefully exiting; when this happens AWS return some generic JSON to the user. If we get a JSON error when parsing the last line of output, it's now assumed that there was no response. We now validate that there's an "END" marker in SAM's output so we can still provide a useful error message if something really goes wrong

The Lambda response body does not _have_ to be JSON, so we now allow non-JSON bodies without immediately failing the test. This allows testing against projects that return HTML, for example

### DIFF
--- a/lib/maze/aws/sam.rb
+++ b/lib/maze/aws/sam.rb
@@ -62,9 +62,7 @@ module Maze
         #
         # @return [Hash]
         def parse(output)
-          begin
-            parsed_output = JSON.parse(output.last)
-          rescue JSON::ParserError
+          unless valid?(output)
             raise <<~ERROR
               Unable to parse Lambda output!
               The likely cause is:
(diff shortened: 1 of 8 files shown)


lib/maze/aws/sam.rb
```diff
@@ -62,23 +62,48 @@ def build_invoke_command(lambda, event)
...
           # Error output has no "body" of additional JSON so we can stop here
           return parsed_output unless parsed_output.key?('body')
 
-          parsed_output['body'] = JSON.parse(parsed_output['body'])
+          # The body is _usually_ JSON but doesn't have to be. We attempt to
+          # parse it anyway because it allows us to assert against it easily,
+          # but if this fails then it may just be in another format, e.g. HTML
+          begin
+            parsed_output['body'] = JSON.parse(parsed_output['body'])
+          rescue JSON::ParserError
```
twometresteve, line 93: Rubocop doesn't like empty rescue blocks.

```suggestion
          rescue JSON::ParserError
            # Ignore
```

twometresteve (approved):
LGTM

PR #239, opened 2021-03-16
Add a step asserting a minimum number of requests to have received

Cawllec:
## Goal

This step should be used exclusively for stress-testing, as ensuring we don't receive excessive requests is an important criteria in our end-to-end testing.

### DIFF
--- a/lib/features/steps/request_assertion_steps.rb
+++ b/lib/features/steps/request_assertion_steps.rb
@@ -55,6 +55,16 @@ Then('I wait to receive {int} {word}') do |request_count, request_type|
   list.sort_by_sent_at! request_count
 end
 
+# Verify that at least a certain amount of requests have been received
+# This step is only intended for use in stress tests
+#
+# @step_input min_received [Integer] The minimum amount of requests required to pass
+# @step_input request_type [String] The type of request (error, session, build, etc)
+Then('I have received at least {int} {word} requests') do |min_received, request_type|
+  list = Maze::Server.list_for(request_type)
+  assert(list.size >= min_received, "Actually received #{list.size} #{request_type} requests")
+end
+
 # Assert that the test Server hasn't received any requests - of a specific, or any, type.
 #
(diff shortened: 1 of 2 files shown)


lib/features/steps/request_assertion_steps.rb
```diff
@@ -55,6 +55,16 @@ def assert_received_requests(request_count, list, list_name)
   list.sort_by_sent_at! request_count
 end
 
+# Verify that at least a certain amount of requests have been received
+# This step is only intended for use in stress tests
+#
+# @step_input min_received [Integer] The minimum amount of requests required to pass
+# @step_input request_type [String] The type of request (error, session, build, etc)
+Then('I have received at least {int} {word} requests') do |min_received, request_type|
```
imjoehaines, line 63: I think this should be "I have received at least 2 errors" for consistency with "I wait to receive 2 errors" and "I should receive no errors"

```suggestion
Then('I have received at least {int} {word}') do |min_received, request_type|
```

lib/features/steps/request_assertion_steps.rb
```diff
@@ -55,6 +55,16 @@ def assert_received_requests(request_count, list, list_name)
...
+  list = Maze::Server.list_for(request_type)
+  assert(list.size >= min_received, "Actually received #{list.size} #{request_type} requests")
```
imjoehaines, line 65: `assert_operator` should give a slightly nicer diff if it fails:

```suggestion
  assert_operator(list.size, :>=, min_received, "Actually received #{list.size} #{request_type} requests")
```

imjoehaines (commented):
Do we need to sleep, similar to "I should receive no errors"? I think it'd be possible to start sending a bunch of requests but assert against the received list before they've all been delivered

Cawllec:
I didn't want to add that as the place where this is used waits until a message has been received on the interactive CLI to show the test is complete.  Only after that do we want to assert a number of requests have been received.

👍  to the other suggestions

imjoehaines (approved):

PR #240, opened 2021-03-23
Working sauce-labs prototype

Cawllec:
## Goal

To produce an MVP running against Android and iOS demonstrating the use of sauce-labs in appium based tests.


lib/features/support/hooks.rb
```diff
@@ -14,6 +14,10 @@
   # Start mock server
   Maze::Server.start
   config = Maze.config
+
+
+  puts "Farm is #{config.farm}"
+
```
twometresteve, line 20: I imagine we'll remember to take this out at some point. :-)

twometresteve (approved):
LGTM

PR #241, opened 2021-03-23
Sauce-labs iOS PoC (Using pre-built app)

Cawllec:
## Goal

Should follow on from #240

Adds an iOS PoC/Test app using the new sauce labs additions


twometresteve (approved):
LGTM

PR #245, opened 2021-03-30
Update user interface for Sauce Labs use

twometresteve:
## Goal

Update the command line options and environment variables for use with Sauce Labs.

## Design

Centres around use of the `--device`, `--os` and `--os-version` command line options to control the device selection as allowed by Sauce Labs. 

## Changeset

- `Devices` class is now Browser Stack specific - `BrowserStackDevice`
- `MAZE_DEVICE_FARM_USERNAME` removed in favour of SL and BS specific variables.

## Tests

Covered by CI.


Cawllec (approved):
LGTM 👍

PR #249, opened 2021-04-01
User interface refinements and validation

twometresteve:
## Goal

The Maze Runner v5.0.0 had some rough edges (but consciously so for the benefit of an early release).  This PR round off some of those edges, correcting the processing of some command line options and improving input validation.

## Design

This PR "banks" a number of improvements, but there are likely to be further change sets before we make the next release.

## Tests

Covered by CI.

### DIFF
--- a/lib/features/support/hooks.rb
+++ b/lib/features/support/hooks.rb
@@ -45,7 +45,8 @@ AfterConfiguration do |_cucumber_config|
                                                config.access_key
   elsif config.farm == :sl
     tunnel_id = SecureRandom.uuid
-    if config.device
+
+    if config.device || config.os || config.os_version
 
       config.app = Maze::SauceLabsUtils.upload_app config.username,
                                                    config.access_key,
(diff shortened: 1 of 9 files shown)


lib/features/support/hooks.rb
```diff
@@ -64,8 +65,10 @@
 
       config.capabilities['app'] = "storage:#{config.app}"
     else
+      $logger.error 'Browser support with Sauce Labs not yet implemented'
+      exit(1)
```
imjoehaines, line 69: Does this need to exit or can it raise instead?

lib/maze/option/validator.rb
```diff
@@ -74,6 +74,40 @@ def validate_bs(options, errors)
...
+          errors << "A device or browser option must be specified"
+        elsif browser
+          errors << 'Browsers not yet implemented on Sauce Labs'
+        else
+          # App
+          app = options[Option::APP]
+          if app.nil?
+            errors << "--#{Option::APP} must be provided when running on a device"
+          else
+            uuid_regex = Regexp.new '^[0-9a-f]{8}-([0-9a-f]{4}-){3}[0-9a-f]{12}$'
```
imjoehaines, line 98: `^` and `$` match the start & end of a line rather than the entire string. This should probably be `\A` and `\z`, which match the start & end of the string — https://rubyapi.org/3.0/o/regexp#class-Regexp-label-Anchors

```suggestion
            uuid_regex = /\A[0-9a-f]{8}-([0-9a-f]{4}-){3}[0-9a-f]{12}\z/
```

imjoehaines (approved):

lib/features/support/hooks.rb
twometresteve: Changed to raise.

PR #252, opened 2021-04-09
Adds logging to a file for the appium server

Cawllec:
## Goal

Allows the appium server to output to a file which can be attached to the buildkite build when running in local mode.

Adds an `--appium-logfile` option denoting the file output, which defaults to `appium_server.log`

## Testing

- Tested manually and confirmed that the `appium_server.log` file is created by default and contains appium logs
- In addition, the `--appium-logfile` override successfully sets the file path of the logfile

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,10 @@
-# 5.0.2 - TBD
+# 5.1.0 - TBD
+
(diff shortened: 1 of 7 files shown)


twometresteve (approved):
LGTM

PR #253, opened 2021-04-12
Add license finder to pipeline

twometresteve:
## Goal

Add a check to the pipeline to ensure that any added dependencies comply with company policy.

## Design

A script is provided to run the license checks locally.  Decision files for the [LicenseFinder](https://github.com/pivotal/LicenseFinder) tool are held centrally with our [license-audit tool](https://github.com/bugsnag/license-audit) for process control and to avoid duplication.

## Changeset

New pipeline step to perform license auditing.

## Tests

Covered by CI and by running the script locally.  I've confirmed that a dependency not covered by an approved license would cause CI to fail.

### DIFF
--- a/dockerfiles/Dockerfile.audit
+++ b/dockerfiles/Dockerfile.audit
@@ -1,11 +1,9 @@
-FROM ubuntu:20.04 as ci
+FROM licensefinder/license_finder
 
-RUN apt-get update && apt-get install -y ruby-full apt-utils docker-compose wget unzip bash bundler \
-                                        # Needed by nokogiri (a dependency of appium_lib)
-                                        zlib1g libpng-dev zlibc zlib1g zlib1g-dev curl \
-                                        # Needed by curb (a dependency of Cucumber)
-                                        libcurl4 libcurl4-openssl-dev
+RUN apt-get update && apt-get install \
+                      # Needed by curb (a dependency of Cucumber)
+                      libcurl3 libcurl4-openssl-dev
 
-WORKDIR /app
+WORKDIR /scan
 
-CMD /app/scripts/license_finder.sh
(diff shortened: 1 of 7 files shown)


scripts/license_finder.sh
```diff
@@ -0,0 +1,3 @@
+curl https://raw.githubusercontent.com/bugsnag/license-audit/master/config/decision_files/global.yml -o config/global.yml
```
imjoehaines, line 1: If we're adding this to every pipeline, I think we should have a more robust way of including this file — if it moves or if the branch is renamed then this will break

Maybe we could build the config into a dockerfile that can be downloaded & used in each pipeline? Then any change to the config file can be automatically picked up, assuming everything points to the `latest` tag

dockerfiles/Dockerfile.audit
```diff
@@ -0,0 +1,11 @@
+FROM ubuntu:20.04 as ci
```
imjoehaines, line 1: License Finder has its own dockerfile which already includes all dependencies and the tool itself — https://hub.docker.com/r/licensefinder/license_finder

I think we should use that so we don't need to write custom dockerfiles for each repo

twometresteve: Done, although we still need to build our own image `FROM` it as there are some extra dependencies we need.

scripts/license_finder.sh
twometresteve: Agreed, there's a lot more we could be doing with the licence-audit tool going forwards.  For now I'm happy to take the risk that we need to updated all the pipelines in response to any changes in it (it's a low risk given what we're planning on doing).

imjoehaines (approved):

PR #255, opened 2021-05-10
Fix issue with os not being defined in local test runs

Cawllec:
## Goal

When running local MacOS tests, the platform-dependent tables fail due to the OS not being correctly set. This resolves the issue by using the configuration option as a default.


twometresteve (approved):
LGTM

PR #257, opened 2021-05-27
Avoid jcenter() dependency by updating android test fixture

fractalwrench:
## Goal

Avoids use of jcenter() which is at end-of-life by updating the Android test fixture. Changes include:

- Update to AGP 4.2.1, which has no transitive dependencies that are only hosted on jcenter()
- Bump gradle wrapper to 6.8 (required for AGP bump)
- Specify the ndk version explicitly
- Suppress pre-existing lint warnings

## Tests

Relied on existing test coverage.


fractalwrench:
The Saucelabs CI job seems to be failing for what looks like unrelated reasons to these changes - wasn't too sure where to start debugging that.

twometresteve:
> The Saucelabs CI job seems to be failing for what looks like unrelated reasons to these changes - wasn't too sure where to start debugging that.

We no longer have a Sauce Labs account - could you simply remove the step please?

twometresteve (approved):
LGTM - thanks for the contribution.  :-)

If you could just take the Sauce Labs pipeline steps out for us too, that would be great.

twometresteve (approved):
LGTM

PR #258, opened 2021-06-03
Add —always-log option

twometresteve:
## Goal

Adds an `--always-log` option, defaulting to `false`, to allow all received requests to be logged at the end of a scenario, whether it passes or fails.

## Tests

Tested locally and unit test updated.


lemnik (approved):
LGTM

PR #261, opened 2021-06-23
Add multipart steps usable against any request queue

Cawllec:
Adds new multipart request steps that allow multipart field checking on any received requests.

Deprecated steps have not been updated (as there are up to date replacements that should be used instead).

To keep backwards compatibility the original steps call into the new steps using the `error` queue by default.

### DIFF
--- a/lib/features/steps/multipart_request_steps.rb
+++ b/lib/features/steps/multipart_request_steps.rb
@@ -20,7 +20,7 @@ end
 
 # Verifies that the request contains multipart form-data
 Then('the request is valid multipart form-data') do
-  valid_multipart_form_data?(Maze::Server.errors.current)
+  step 'Then the error request is valid multipart form-data'
 end
 
 # Verifies that any type of request contains multipart form-data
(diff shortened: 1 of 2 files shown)


lib/features/steps/multipart_request_steps.rb
```diff
@@ -18,28 +18,77 @@ def valid_multipart_form_data?(request)
...
+  valid_multipart_form_data?(list.current)
+end
+
+# (Deprecated) Retained for backwards compatibility
+# Use "the {word} request is valid multipart form-data"
+#
 # Verifies that the request contains multipart form-data
 Then('the request is valid multipart form-data') do
-  valid_multipart_form_data?(Maze::Server.errors.current)
+  $logger.warn 'This step is deprecated and may be removed in a future release'
```
twometresteve, line 34: Could we do something like raise a ticket under a v6 epic to ensure we do remove this at the next major?

twometresteve (approved):
LGTM

twometresteve (approved):
LGTM

PR #262, opened 2021-06-23
Log received requests to file for all scenarios

twometresteve:
## Goal

Writes all received requests to file beneath `maze_outputs`.

## Design

Files will be grouped in folders by scenario and then by pass/fail status of the scenario.  This allows CI runs to select only failed scenarios for upload should they wish.

## Changeset

This changes the default behaviour when running locally to only log received requests to file and not to the console.  Console logging behaviour on Buildkite will remain the same.

Users wanting to retain console logging of requests for failed scenarios:

1. Alias the execution of maze-runner:
```
alias bem='bundle exec maze-runner --log-requests'
```
2. Set the config directly in their `env.rb` file:
```
AfterConfiguration do |_config|
  Maze.config.log_requests = true
end

```

## Tests

Generation and upload of logged files now added to the Maze Runner CI pipeline.  Combinations of options for running locally tested manually.


lib/features/support/hooks.rb
```diff
@@ -1,13 +1,18 @@
...
 require 'json'
 require 'securerandom'
 require 'selenium-webdriver'
 require 'uri'
 
 AfterConfiguration do |_cucumber_config|
 
+  # Clear out maze_output folder
+  maze_output = File.join(Dir.pwd, 'maze_output')
+  FileUtils.rm_rf(maze_output)
```
imjoehaines, line 14: Doesn't this delete the entire directory? This would break the idea of adding a `.gitignore` to it

This could be something like:

```suggestion
  # Clear out maze_output folder
  maze_output = Dir.glob(File.join(Dir.pwd, 'maze_output', '*'))
  FileUtils.rm_rf(maze_output)
```

lib/maze/option/parser.rb
```diff
@@ -138,10 +138,10 @@ def parse(args)
             text 'Logging options:'
 
             opt Option::LOG_REQUESTS,
-                "Log a list of received requests in the event of test failure",
+                "Log a list of received requests in the event of test failure.  Defaults to true if the BUILDKITE environment variable is set",
                 short: :none,
                 type: :boolean,
-                default: true
+                default: false
```
imjoehaines, line 144: Is there a downside to keeping the default `true`?

lib/features/support/hooks.rb
```diff
@@ -207,6 +215,47 @@ def output_received_requests(request_type)
...
+
+  request_types.each do |request_type|
+    list = Maze::Server.list_for(request_type).all
+    next if list.empty?
+
+    list.each do |request|
+      filename = "#{request_type}.log"
+      filepath = File.join(path, filename)
+
+      File.open(filepath, 'w+') do |file|
```
imjoehaines, line 241: This doesn't change per-request so could be outside of the `list.each` to save opening the same file multiple times

lib/features/support/hooks.rb
twometresteve: Good point, nice suggestion.

lib/maze/option/parser.rb
twometresteve: It upsets fewer people setting it false.  Most people have asked for the logging of requests to disappear from the console by default (and not just in today's internal poll).

lib/features/support/hooks.rb
twometresteve: Good point, I've also added a heading for each request as well to make it easier to navigate through files.

lib/features/support/hooks.rb
```diff
@@ -172,6 +177,9 @@
     output_received_requests('logs')
   end
 
+  # Log all received requests to file
+  write_requests(scenario)
```
imjoehaines, line 181: There should be a way to turn this off

twometresteve: Yeah, that's fair - an option defaulting to true ok though?

Cawllec (approved):

lib/features/support/hooks.rb
imjoehaines: One of the options has to be true or you'll get no failure reasons 😬

lib/features/support/hooks.rb
```diff
@@ -1,13 +1,21 @@
...
 require 'uri'
 
 AfterConfiguration do |_cucumber_config|
 
+  # Clear out maze_output folder
+  maze_output = Dir.glob(File.join(Dir.pwd, 'maze_output', '*'))
+  unless maze_output.empty?
+    maze_output.each { |path| $logger.info "Clearing contents of #{path}" }
+  end
+  FileUtils.rm_rf(maze_output)
```
imjoehaines, line 17: Maybe this should be gated behind a `Maze.config.file_log` check? Or check if `maze_output` exists first?

lib/maze/option/parser.rb
```diff
@@ -137,12 +136,18 @@ def parse(args)
...
                 default: true
 
+            opt Option::LOG_REQUESTS,
+                "Log lists of received requests to the console in the event of scenario failure.  Defaults to true if the BUILDKITE environment variable is set",
+                short: :none,
+                type: :boolean,
+                default: false
+
             opt Option::ALWAYS_LOG,
                 "Always log all received requests at the end of a scenario, whether is passes or fails",
```
imjoehaines, line 152:
```suggestion
                "Always log all received requests to the console at the end of a scenario, whether is passes or fails",
```

lib/features/support/hooks.rb
twometresteve: It should always have been in the guard above, but I think it would also be surprising if the folder got removed when running with `file_log` off - so have added that guard.

lib/maze/option/parser.rb
twometresteve: 👍  changed

imjoehaines (approved):

PR #263, opened 2021-06-24
Remove short command line options to avoid clashes with cucumber

twometresteve:
## Goal

Slightly knee-jerk removal of short-form command line options to avoid an clashes with underlying Cucumber options.

## Design

Longer term, a wholesale reassessment of command line use is likely to take place, including the use of files for options and `--` separators in the CL (for instance).

## Tests

Covered by CI and existing unit tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 - Add `--file-log` option to write received requests to files [#262](https://github.com/bugsnag/maze-runner/pull/262)
 
+## Fixes
+
+- Remove short form from all command line options [#263](https://github.com/bugsnag/maze-runner/pull/263)
+
 # 5.3.0 - 2021/06/21
 
 ## Enhancements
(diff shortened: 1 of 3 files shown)


imjoehaines (approved):
Is this going to break any pipelines that automatically use the latest MR?

nickdowell:
Where are the long-form names defined ❓

imjoehaines:
> Where are the long-form names defined ❓

For Maze Runner? They are all here: https://github.com/bugsnag/maze-runner/blob/master/lib/maze/option.rb

`opt <constant>` becomes `--<constant>` on the command line

nickdowell (approved):

twometresteve:
> Is this going to break any pipelines that automatically use the latest MR?

It's a fair question - technically this is a breaking change, but I happen to know (99% confident) that the long form is always used.

PR #265, opened 2021-06-29
Upload apps to bitbar

Cawllec:
## Goal

Adds the scaffolding to uploading and reusing apps within the BitBar framework.

Adds:
- `--bitbar-api-key` (and equivalent ENV var `BITBAR_API_KEY`) as an option
- `bb` as an allowable `--farm` option
- Validation on the options a la previous device farm opts
- A `BitBarUtils` class including the upload/reuse method
- Hooks to call for `--app` to be uploaded via the above method, then `exit` (to avoid any further maze-running)

TODO:
- Add tests against the above
- Figure out a better app id validation.  The current system appears to be a 9 digit number [0-9], but the current validation regex is slighty more open (allowing 8-12 digits).  I'll ask support whether that 9 digit number as an ID is always correct.


lib/maze/bitbar_utils.rb
```diff
@@ -0,0 +1,54 @@
...
+  class BitBarUtils
+    class << self
+      attr_accessor :connect_shell
+
+      # Uploads an app to BitBar for later consumption
+      # @param api_key [String] The BitBar API key
+      # @param app [String] A path to the application file
+      def upload_app(api_key, app)
+        pp app
+        uuid_regex = /\A[0-9]{8,12}\z/
```
twometresteve, line 14: Allow any Long

twometresteve (approved):
LGTM

PR #266, opened 2021-06-29
Remove now unnecessary libcurl3 from license_finder docker image

Cawllec:
## Goal

libcurl3 and libcurl4-openssl-dev conflict on the latest version of the license-audit base image, causing the build to fail.  As only libcurl4-openssl-dev is required we can remove libcurl3 completely.


twometresteve (approved):
LGTM

PR #270, opened 2021-07-08
Correct guard for “platform-dependent” steps

twometresteve:
## Goal

Corrects the guard on "platform-dependent value" steps that was using a now false assumption that Appium must be in use for the `--os` option to be set.

## Tests

Verified locally that Unity steps that were failing due to this now pass.


rich-bugsnag (approved):
thanks @twometresteve !

PR #272, opened 2021-07-08
Add header is null step

twometresteve:
## Goal

Adds a Cucumber step to check for null headers.

### DIFF
--- a/lib/features/steps/header_steps.rb
+++ b/lib/features/steps/header_steps.rb
@@ -11,6 +11,19 @@ Then('the {word} {string} header is not null') do |request_type, header_name|
                  "The #{request_type} '#{header_name}' header should not be null")
 end
 
+# Tests that a request header is null
+#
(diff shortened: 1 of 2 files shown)


lib/features/steps/header_steps.rb
```diff
@@ -11,6 +11,19 @@
...
+# Tests that a request header is null
+#
+# @step_input request_type [String] The type of request (error, session, build, etc)
+# @step_input header_name [String] The header to test
+Then('the {word} {string} header is null') do |request_type, header_name|
+  request = Maze::Server.list_for(request_type).current[:request]
+
+  assert_nil(
+      request[header_name],
+      "The #{request_type} '#{header_name}' header should be null"
```
imjoehaines, line 23:
```suggestion
    request[header_name],
    "The #{request_type} '#{header_name}' header should be null"
```

imjoehaines (approved):

PR #274, opened 2021-07-13
Add document server functionality 

twometresteve:
## Goal

Adds command line options `--ds-root`, `--ds-bind-address` and `--ds-port` to allow clients to easily run up an HTTP server for tests to serve up documents.

## Design

New `DocumentServer` class starts a WEBrick server, fed by command line options in the standard way.

## Changeset

New server class, options and tests.

## Tests

Covered by new unit and e2e tests.


lib/maze/document_server.rb
```diff
@@ -0,0 +1,24 @@
...
+    class << self
+      def start
+        @thread = Thread.new do
+          options = {
+              Port: Maze.config.ds_port,
+              Logger: $logger,
+              AccessLog: []
+          }
+          options[:BindAddress] = Maze.config.ds_bind_address unless Maze.config.ds_bind_address.nil?
+          server = WEBrick::HTTPServer.new :Port => Maze.config.ds_port, :DocumentRoot => Maze.config.ds_root
```
imjoehaines, line 17: `options` is unused here, I think this should be:

```suggestion
          options = {
              Port: Maze.config.ds_port,
              DocumentRoot: Maze.config.ds_root,
              Logger: $logger,
              AccessLog: []
          }
          options[:BindAddress] = Maze.config.ds_bind_address unless Maze.config.ds_bind_address.nil?
          server = WEBrick::HTTPServer.new(options)
```

imjoehaines: I guess this means there's no test for `BindAddress`

lib/maze/option.rb
```diff
@@ -3,6 +3,11 @@
 # Provides the set of Maze Runner command line options
 module Maze
   module Option
+    # Document server options (ds = document server)
+    DS_BIND_ADDRESS = 'ds-bind-address'
+    DS_PORT = 'ds-port'
+    DS_ROOT = 'ds-root'
```
imjoehaines, line 9: I think if the name needs to be explained in a comment it could be clearer. `--document-server-bind-address` is a bit long but given this will be scripted almost all the time maybe that's ok?

Perhaps merging the bind address and port would let this be `--document-server`? Ruby's URI module can turn that back into a host/port pair:

```ruby
require "uri"

address = "localhost:8080" # from --document-server
# otherwise URI doesn't know what kind of URI it is
address = "http://#{address}" unless address.start_with?("http")

uri = URI(address)
port = uri.port # => 8080
host = uri.host # => "localhost"
```

We'd still need `--document-server-root` but that's not too much longer than `--ds-bind-address`

twometresteve: Combining them into `--document-server` seems like a nice approach, but would we want it to be inconsistent with the existing `--bind-address` and `--port` options?  And what if someone wants to change the port without giving a specific bind address (which I believe we had problems with before)?

twometresteve: How about `--doc-server-bind-address` as some kind of compromise?

lib/maze/document_server.rb
twometresteve: I'm very surprised my IDE didn't tell me that `options` wasn't read.

lib/maze/option.rb
imjoehaines: I think that still works so long as you always prefix the port with `:`, e.g. `--document-server :1234`:

```ruby
irb(main):001:0> a = URI('http://:1234')
irb(main):002:0> a.port
=> 1234
irb(main):003:0> a.host
=> nil
```

imjoehaines: I think that still works so long as you always prefix the port with `:`, e.g. `--document-server :1234`:

```ruby
irb(main):001:0> a = URI('http://:1234')
irb(main):002:0> a.port
=> 1234
irb(main):003:0> a.host
=> nil
```

`--doc-server-` is better than `--ds-` but I think I still prefer `--document-server-`. What about calling it a file server instead? `--file-server-` is only 1 char longer than `--doc-server-` but is a bit clearer IMO

twometresteve: Agreed to go with `--document-server...`, as we'd typically expect the config to be set in code anyway.

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 5.7.0 - 2021/07/13
+
+## Enhancements
+
+- Add document server (`--ds-root` and associated options) [#274](https://github.com/bugsnag/maze-runner/pull/274)
```
imjoehaines, line 5:
```suggestion
- Add document server (`--document-server-root` and associated options) [#274](https://github.com/bugsnag/maze-runner/pull/274)
```

imjoehaines (approved):

PR #276, opened 2021-07-14
Ensure maze after_configuration hooks are always called

twometresteve:
## Goal

Ensure the registered `Maze.hooks.after_configuration` blocks are always called.  They were previously skipped when no device farm was in use.

## Changeset

`hooks.rb` is very busy and in need of a refactor.  To start moving in that direction, I've moved a section of code into a separate routine.  At some point we can refactor into a separate class/file altogether.

## Tests

Covered by updating the existing framework tests.


lib/features/support/hooks.rb
```diff
@@ -62,7 +60,7 @@
       config.app = Maze::SauceLabsUtils.upload_app config.username,
                                                    config.access_key,
                                                    config.app
-      # Capabilities
+      # Capabilities``
```
imjoehaines, line 63:
```suggestion
      # Capabilities
```

imjoehaines (approved):
> `hooks.rb` is very busy and in need of a refactor. To start moving in that direction, I've moved a section of code into a separate routine. At some point we can refactor into a separate class/file altogether.

I think maybe the best way to do it would be to break each "farm" into its own class, which can then have methods for each hook to call, e.g.

```ruby
module Maze::Farm
  class BrowserStack
    def after_configuration
    end

    def before
    end

    def after
    end

    def something_else
    end
  end
end
```

Then `Maze::Config.farm` can hold an instance of that class, where `:none` just noops, then all of the logic goes away (apart from the CLI parser figuring out which class to create)

Obviously not part of this PR though!

twometresteve:
> I think maybe the best way to do it would be to break each "farm" into its own class, which can then have methods for each hook to call, e.g.
> 
> ```ruby
> module Maze::Farm
>   class BrowserStack
>     def after_configuration
>     end
> 
>     def before
>     end
> 
>     def after
>     end
> 
>     def something_else
>     end
>   end
> end
> ```
> 
> Then `Maze::Config.farm` can hold an instance of that class, where `:none` just noops, then all of the logic goes away (apart from the CLI parser figuring out which class to create)

Yes, nice suggestion - I've created a ticket to do the refactor.

PR #277, opened 2021-07-16
Add capabilities support to bitbar utils

Cawllec:
## Goal

Adds support for defining and adding regular and device capabilities, as well as a couple of changes to the bitbar tunnel logic to ensure it connects correctly.


lib/maze/bitbar_devices.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+module Maze
+  # Provides a source of capabilities used to run tests against specific BrowserStack devices
+  # noinspection RubyStringKeysInHashInspection
+  class BitBarDevices
+    APPIUM_1_9_1 = '1.9.1'
+    APPIUM_1_15_0 = '1.15.0'
+    APPIUM_1_20_2 = '1.20.2'
```
twometresteve, line 9: Looks like these are unused?

lib/maze/bitbar_devices.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+module Maze
+  # Provides a source of capabilities used to run tests against specific BrowserStack devices
```
twometresteve, line 4:
```suggestion
  # Provides a source of capabilities used to run tests against specific BitBar devices
```

twometresteve (approved):
LGTM, just a couple of tiny comments.

lib/maze/bitbar_devices.rb
Cawllec: This is configurable for Android, so will probably come up with the next step of running Android apps

PR #278, opened 2021-07-19
add a generic send keys methods for unity as it can't access elements

rich-bugsnag:
add a generic send keys methods for unity as it can't access elements

### DIFF
--- a/lib/maze/driver/appium.rb
+++ b/lib/maze/driver/appium.rb
@@ -119,10 +119,6 @@ module Maze
         @driver.send_keys(text)
       end
 
-      def hide_keyboard
-        @driver.hide_keyboard
-      end
-
       # Sends keys to a given element, clearing it first
       #
       # @param element_id [String] the element to clear and send text to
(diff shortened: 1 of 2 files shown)


lib/maze/driver/appium.rb
```diff
@@ -112,6 +112,17 @@ def set_rotation(orientation)
...
       end
 
+      # Send keys to the device without a specific element
+      #
+      # @param text [String] the text to send
+      def send_keys(text)
+        @driver.send_keys(text)
+      end
+
+      def hide_keyboard
```
twometresteve, line 122: nit: add a comment above the method

twometresteve:
Looks good, but please add an entry to `CHANGELOG.md` (add a new section for 5.8.0 with date TBD and list this as an enhancement).

lib/maze/driver/appium.rb
rich-bugsnag: i remoed this method, i don't need it in the end

twometresteve (approved):
LGTM

PR #279, opened 2021-07-25
Add support for running tests against local browsers

twometresteve:
## Goal

Add support for running Selenium tests on local browsers.

## Design

The change needed to support local Selenium sessions was pretty straightforward, but it meant adding yet another condition to the already very busy `hooks.rb` (now `internal_hooks.rb`), so I've now refactored it.  We now have an explicit concept of a mode of operation, which is one of:
- Appium (local or remote)
- Browser (Selenium - again, local or remote)
- Command (where the software under test is driven by local System calls)

The hooks file was effectively inferring this mode in several locations.  I've refactored the logic to do that inference just once, allowing hook logic to be implemented in a separate class per mode of operation.  The resulting logic is much simpler and clearer.  

In a future release we might consider adding an explicit `--mode` command line option (defaulting to appium).  This would both simplify the logic further and allow us to perform stricter checks on other command line option, providing more useful feedback to the user.

## Changeset

- New classes `AppiumHooks`, `BrowserHooks` and `CommandHooks`
- `hooks.rb` renamed to `internal_hooks.rb` and logic refactored to call into new classes.

## Tests

The CI pipeline for MazeRunner contains Appium and Command mode tests, but not Browser.  Given that and due to the nature of the change, I've performed additional testing by running the full CI pipeline for each of our following notifiers:
- JS (Browser)
- Unity (WebGL)
- Cocoa
- Android
- Bugsnag Android Gradle Plugin


lib/maze/hooks/command_hooks.rb
```diff
@@ -0,0 +1,20 @@
+# Contains logic for the Cucumber hooks when in Command mode
+module Maze
+  class CommandHooks
```
imjoehaines, line 3: Would it make sense to call this `NoopHooks` or `NullHooks` or something along those lines as there's no logic here?

lemnik (approved):
LGTM

lib/maze/hooks/browser_hooks.rb
```diff
@@ -0,0 +1,38 @@
...
+      # Create and start the relevant driver
+      if config.farm == :bs
+        selenium_url = "http://#{config.username}:#{config.access_key}@hub.browserstack.com/wd/hub"
+        Maze.driver = Maze::Driver::Browser.new :remote, selenium_url, config.capabilities
+      elsif config.farm == :local
+        Maze.driver = Maze::Driver::Browser.new Maze.config.browser.to_sym
+      end
+    end
+
+    def before
```
imjoehaines, line 26: To avoid having to always implement all of the hook methods, we could make an "abstract" class that noops and is extended by these hook classes. That way they only need to implement the methods that they require as they'll inherit the noops. i.e.

```rb
class Hooks
  def after_configuration
  end

  def before
  end

  def after
  end

  def at_exit
  end
end

class BrowserHooks < Hooks
  def after_configuration
    # stuff...
  end
end
```

imjoehaines (approved):

lib/maze/hooks/browser_hooks.rb
twometresteve: Ah, yes, that would be better.

lib/maze/hooks/command_hooks.rb
twometresteve: I think it makes sense as it is.  Whilst it is noop at the moment, it could potentially change at any point.  Conceptually it's the hooks needed for Command mode, they just happen to be empty at the moment - so the abstraction feels right to me.

Cawllec (approved):

imjoehaines (approved):

PR #280, opened 2021-07-26
Running iOS integration locally

Cawllec:
## Goal

Makes required changes to run the current iOS tests locally against BitBar.

Recommend command for running them is:

```
BITBAR_USERNAME= BITBAR_ACCESS_KEY=  bundle exec maze-runner --farm=bb --device=IOS_14 --app=144925378 --bb-local=~/SBSecureTunnel --app-bundle-id=com.bugsnag.iOSTestApp --os=iOS
```


twometresteve (approved):
LGTM and the test run passed 4 times in a row for me.

PR #281, opened 2021-07-27
Add sourcemap endpoints to Maze::Server

imjoehaines:
## Goal

This PR adds `/sourcemap` and `/react-native-source-map` endpoints to the server. This is needed for the JS sourcemap tool to update to MR v4 or v5 (currently on v3) after the request types were separated. It can't re-use an existing request endpoint because the tool only accepts a base URL as configuration and will internally append one of the sourcemap paths

The two endpoints share a single request list to help simplify some of the tools steps which can work on either type of sourcemap and currently doesn't know which type is being used. It also means we don't need to deal with the multi-word "React Native sourcemap" request type and no single test would upload both types of sourcemaps

## Tests

Tested by monkey patching the `Maze::Server` class in the sourcemap repo and running MR v5 — https://buildkite.com/bugsnag/bugsnag-source-maps/builds/304


twometresteve (commented):
LGTM, just the usual changelog entry to add.  If you'd like to update the version number and set a release date of today I'd be happy to release it for you.

twometresteve (approved):
LGTM, will merge and release.

PR #282, opened 2021-07-28
BitBar Integration

Cawllec:
## Goal

Adds support for the use of BitBar as a device farm provider for use in appium based e2e testing.

The majority of the changes have already been reviewed elsewhere, however this would still be worth reviewing from the beginning as many fixes have been applied since then.

## Testing

Besides the present testing, two scheduled branches have been set up for [Bugsnag Cocoa](https://buildkite.com/bugsnag/bugsnag-cocoa/builds/5874#0181a973-4f7c-41bf-b05b-ccb08a7188b1) and [Bugsnag Android](https://buildkite.com/bugsnag/bugsnag-android/builds/8025#0181a96f-5235-4dac-b8b7-abf3e2590d9c).  These have reached the point where they are successfully running the tests and while they aren't passing yet, the issues involved appear at first glance to be isolated to the test fixture, and will require other individual attention.

Any further changes required in the process of using BitBar will be added in separate tickets.


lib/maze/bitbar_utils.rb
```diff
@@ -0,0 +1,112 @@
...
+          http.request(request)
+        end
+        res
+      end
+
+      # Informs the test-management-service that in-use account id is no longer in use
+      # @param tms_uri [String] The URI of the test-management-service
+      def release_account(tms_uri)
+        uri = URI("#{tms_uri}/account/release?account_id=#{@account_id}")
+        request['Authorization'] = Maze.config.tms_token
```
twometresteve, line 104: Looks like this should go after the next line?  There's an error output in the scheduled tests saying that `request` is undefined.

twometresteve (approved):
Looks good - just the date in the changelog to update then we can release it?

PR #284, opened 2021-07-30
Support Ruby 3

twometresteve:
## Goal

Allow Maze Runner to be used with Ruby 3.

## Changeset

The only identified change needed is to add a dependency on `webrick`.  The remaining changes enhance the pipeline to run tests with both Ruby 2.7 and 3.

The addition of tests against Ruby 2.5 (the minimum support version) was also considered, but the additional work needed to the base image and pipeline steps did not seem worthwhile.

## Tests

Covered by updated CI pipeline and I've also verified that the Docker image pushed for the branch can be successfully used in another pipeline (Android).

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -6,21 +6,6 @@ steps:
         run: license_finder
     command: /bin/bash -lc '/scan/scripts/license_finder.sh'
 
-  - label: ':docker: Build CI image for Ruby 2.5'
-    key: "ci-image-ruby-2-5"
-    plugins:
-      - docker-compose#v3.7.0:
-          build:
-            - ci-ruby-2.5
-          image-repository: 855461928731.dkr.ecr.us-west-1.amazonaws.com/maze-runner
-          cache-from:
-            - ci:855461928731.dkr.ecr.us-west-1.amazonaws.com/maze-runner:${BRANCH_NAME}-ci-ruby-2.5
-      - docker-compose#v3.7.0:
-          push:
-            - ci:855461928731.dkr.ecr.us-west-1.amazonaws.com/maze-runner:${BRANCH_NAME}-ci-ruby-2.5
-    env:
-      RUBY_VERSION: "2.5"
-
   - label: ':docker: Build CI image for Ruby 2.7'
     key: "ci-image-ruby-2-7"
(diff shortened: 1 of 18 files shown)


test/fixtures/cli/Dockerfile
```diff
@@ -1,5 +1,5 @@
 ARG BRANCH_NAME
-FROM 855461928731.dkr.ecr.us-west-1.amazonaws.com/maze-runner:${BRANCH_NAME}-ci
+FROM 855461928731.dkr.ecr.us-west-1.amazonaws.com/maze-runner:${BRANCH_NAME}-ci-ruby-2.7
```
imjoehaines, line 2: Should this use `RUBY_VERSION` like the others?

```suggestion
ARG RUBY_VERSION
FROM 855461928731.dkr.ecr.us-west-1.amazonaws.com/maze-runner:${BRANCH_NAME}-ci-ruby-${RUBY_VERSION}
```

.buildkite/pipeline.yml
```diff
@@ -74,6 +112,8 @@ steps:
           run: payload-helper-tests
       - docker-compose#v3.7.0:
           run: docker-tests
+    env:
+      RUBY_VERSION: "3"
```
imjoehaines, line 116: Should this be `3.0`? If this is just `3` the version will change with every ruby release, but maybe that's what you want?

imjoehaines (approved):

.buildkite/pipeline.yml
twometresteve: That was my intention, on the assumption that users on Ruby are likely to follow the latest to a large extent - unless you think it's a bad idea for any reason.

test/fixtures/cli/Dockerfile
twometresteve: Absolutely, I was getting confused with the `cli` Docker image itself (but even then burying it in the Dockerfile would have been bad).  Now fixed - including passing of the argument through `docker-compose.yml`.

.buildkite/pipeline.yml
imjoehaines: No, that makes sense 👍

Cawllec (approved):

PR #288, opened 2021-08-27
BitBar: Add account service and iOS PoC on CI

Cawllec:
## GOAL:
- Create service to distribute accounts to MR instances running BitBar
- Integrate with service when `--farm=bb` and `BUILDKITE=true`
- Ensure environment variables can be passed through via Docker-compose


lib/maze/hooks/appium_hooks.rb
```diff
@@ -24,15 +24,19 @@ def after_configuration
...
-                                               config.username,
-                                               config.access_key
+                                              config.username,
+                                              config.access_key
           config.capabilities = Maze::Capabilities.for_bitbar_device config.access_key,
-                                                                     config.device,
-                                                                     config.capabilities_option
-
+                                                                    config.device,
+                                                                    config.capabilities_option
```
twometresteve, line 39: nit: I'm blinkering out some other Rubocop whinges here, but it would be nice if these were still aligned.

twometresteve:
Good to see the additional steps in the pipeline, but it would also be good to update the unit tests for validator/parser to keep them ticking over (without wanting to get bogged down in them at all).

twometresteve (approved):

PR #290, opened 2021-09-07
Add examples of all major device versions for further testing

Cawllec:
## Goal

Adds BitBar devices entries for all supported iOS and Android versions.  At the moment these select specific devices as the operation of the `find_device` capability is currently unknown.

### DIFF
--- a/lib/maze/bitbar_devices.rb
+++ b/lib/maze/bitbar_devices.rb
@@ -55,9 +55,24 @@ module Maze
         hash = {
           # Classic, non-specific devices for each Android version
           'ANDROID_11_0' => make_android_hash('Google Pixel 4 -US'),
+          'ANDROID_10_0' => make_android_hash('Google Pixel 2 Android 10'),
+          'ANDROID_9_0' => make_android_hash('Google Pixel Android 9'),
+          'ANDROID_8_1' => make_android_hash('Google Pixel 2 8.1 -US'),
+          'ANDROID_8_0' => make_android_hash('Google Pixel 8.0 -EU'),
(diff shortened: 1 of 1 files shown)


lib/maze/bitbar_devices.rb
```diff
@@ -55,9 +55,24 @@ def create_hash
...
+          'ANDROID_4_4' => make_android_hash('Motorola Moto X Ghost XT1056'),
 
           # iOS devices
-          'IOS_14' => make_ios_hash('Apple iPad 8 A2270 14.0.1'),
+          'IOS_14' => make_ios_hash('iPhone 12 A2172 14.7 -US'),
+          'IOS_13' => make_ios_hash('Apple iPhone 11 Pro 13.7 -US'),
+          'IOS_12' => make_ios_hash('Apple iPhone 6 A1586 12.4.9'),
+          'IOS_11' => make_ios_hash('Apple iPhone 7 A1660 11.4.1 -US'),
+          'IOS_10' => make_ios_hash('Apple iPhone 6S Plus A1687 10.2'),
+          'IOS_9_3' => make_ios_hash('Apple iPhone 4S A1387 9.3.5'),
```
twometresteve, line 75: Just an observation that no iOS 9 devices are showing on Bitbar this morning.  I presume it/they have been taken offline temporarily and I have no reason to suspect that the name is wrong.

twometresteve (approved):

lib/maze/bitbar_devices.rb
Cawllec: 🤔  Ok, good to know.  I'm creeping into the find_device capability but without some testing I'm not sure how it effects the actual search parameters yet.

PR #291, opened 2021-09-17
Add reflective servlet to document server

twometresteve:


lib/maze/document_server.rb
```diff
@@ -6,7 +6,10 @@ module Maze
   # HTTP server for a given document root
   class DocumentServer
     class << self
+      # Start the document server.  This is intended to be called only once per test run.
+      # Use manual_start for finer grained control.
       def start
+        @manual_start = false
```
Cawllec, line 12: This will override the `@manual_start = true` in `manual_start` won't it?

lib/features/steps/document_server_steps.rb
```diff
@@ -0,0 +1,7 @@
+# @!group Document steps
+
+# Starts the document server manually.  It will be stopped automatically at the end of each scenario
+# (if started in this way).
+When('I start the document server') do
```
Cawllec, line 5: I get why this step exists - but I'm wondering if there's a neater way to accomplish the same things.  I guess at least this is explicit and can be put in a background section?

lib/maze/reflective_servlet.rb
```diff
@@ -0,0 +1,63 @@
...
+      status = request.query['status']
+      reflect response, delay_ms, status
+    end
+
+    # Accepts a POST request to provide a reflective response to.
+    #
+    # @param request [HTTPRequest] The incoming GET request
+    # @param response [HTTPResponse] The response to return
+    def do_POST(request, response)
+      body = JSON.parse(request.body)
```
Cawllec, line 26: It may be worth having this check the content-type first, as other services may end up sending other mime-types (such as form-data) to test things like build/symbolication handling

Cawllec (commented):
A few comments, nothing major hopefully.

lib/features/steps/document_server_steps.rb
twometresteve: The document server can also be started on the command line or by setting a config value in `env.rb`.  If lots of scenarios needed it, I'd go that route rather than use a Background.  I only did it this way because it's the only scenario that needs the document server and other ways of starting it will keep it running for the whole test suite - but I'm not not precious about it either way.

lib/maze/reflective_servlet.rb
twometresteve: Do you mean so that it returns a reasonable response body and status code for unsupported content types, rather than a horrible 500?  Makes sense...

lib/maze/document_server.rb
twometresteve: Good spot - have moved this to the `manual_stop` method.

lib/maze/reflective_servlet.rb
twometresteve: Guard now added.

Cawllec (approved):

PR #292, opened 2021-09-29
BitBar: Use devices api to search devices before starting tests

Cawllec:
Note: This also contains an update of the latest master changes for integration/bitbar

## Goal

- Updates the BitBar devices methods to search using the devices API instead of using prescriptive device names.
- Uses the `--os` and `--os-version` to determine which device to use

## Notes

- Requires some additional error management which has been ticketed
- May require tweaks to appium versions based on the os_version - this is being investigated atm


twometresteve:
Looks like the integration branch needs updating from `master` as well?

twometresteve (dismissed):

twometresteve (commented):
Seems I approved it a bit early, as there are unit test failures.

lib/maze/bitbar_devices.rb
```diff
@@ -8,7 +9,56 @@ class BitBarDevices
...
+        }
+        all_devices = call_bitbar_api(path, query, api_key)
+        filtered_devices = all_devices['data'].reject { |device| device['locked'] }
+        filtered_devices.first['displayName']
+      end
+
+      def get_device(device_group, platform, platform_version, api_key)
+        device_group_id = DEVICE_GROUP_IDS[device_group]
+        device_name = get_filtered_device_name(device_group_id, api_key)
+        device_caps = case platform.downcase
```
twometresteve, line 50: Rubocop says this assignment before return is redundant and that `device_caps` can just be removed.  I'm not a great fan of Ruby's "magic" return behaviour, but that's not a battle I'm going to win and so I'm prepared to be more-or-less a slave to Rubocop.

lib/maze/bitbar_devices.rb
```diff
@@ -8,7 +9,56 @@ class BitBarDevices
...
+        when 'android'
+          automationName = 'UiAutomator1' if platform_version.start_with?('5')
```
twometresteve, line 52: Please use `snake_case` for variable names.

lib/maze/bitbar_devices.rb
```diff
@@ -8,7 +9,56 @@ class BitBarDevices
     APPIUM_1_15_0 = '1.15.0'
     APPIUM_1_20_2 = '1.20.2'
 
+    BASE_URI = 'https://cloud.bitbar.com/api/v2/me'
+    FILTER_PATH = 'devices/filters'
+
+    DEVICE_GROUP_IDS = {
+      # Classic, non-specific devices for each Android version
+      'ANDROID_10_0' => '46024',
```
twometresteve, line 17: I'm guessing these ids are very specific to our environment.  I started to wonder if they should be factored out into the environment (at a later stage), but I suspect that could be over-engineering it if we're not going to be changing groups around.

lib/maze/capabilities.rb
```diff
@@ -36,13 +36,16 @@ def for_browser_stack_browser(browser_type, local_id, capabilities_option)
...
       # @param device_type [String]
-      def for_bitbar_device(bitbar_api_key, device_type, capabilities_option)
+      def for_bitbar_device(bitbar_api_key, device_type, platform, platform_version, capabilities_option)
         capabilities = {
           'bitbar_apiKey' => bitbar_api_key,
-          'bitbar_testrun' => device_type,
-          'bitbar_findDevice' => false
+          'bitbar_testrun' => "#{platform} #{platform_version}",
+          'bitbar_findDevice' => false,
+          'bitbar_testTimeout' => 3600,
```
twometresteve, line 44: Is this the maximum allowed?  Some test suites take over an hour to run (if we run then all on the same device).  Can this be bumped to 1.5 or even 2 hours.

lib/maze/hooks/appium_hooks.rb
```diff
@@ -76,7 +71,7 @@ def after_all
           Maze::BrowserStackUtils.stop_local_tunnel
         elsif Maze.config.farm == :bb
           Maze::BitBarUtils.stop_local_tunnel
-          Maze::BitBarUtils.release_account(Maze.config.tms_uri)
+          Maze::BitBarUtils.release_account(Maze.config.tms_uri) if ENV['BUILDKITE']
```
twometresteve, line 74: We should probably think about n approach for doing this that doesn't rely on checking the environment variable in (potentially) various part of the code.  It's an existing problem (that I created!) for logging, though, so no action needed here.  Just something to think about.

twometresteve (approved):
Left a few style and minor comments, but LGTM for the current stage of development.

lib/maze/bitbar_devices.rb
Cawllec: I think for now it's fine, as they should be immutable within our bitbar env, but yes, it would potentially be better to load from the env at some point.

lib/maze/hooks/appium_hooks.rb
Cawllec: How's about a configuration option which defaults to False but is overwritten by the `BUILDKITE` env var?  I think it should be in a separate PR however

PR #295, opened 2021-10-12
Add auto retries for failing scenarios with correct tags or selenium errors

Cawllec:
## Goal

To enable retrying scenarios in specific circumstances, this feature:
- Will allow a single retry for scenarios tagged with `retry`, `retryable`, or `retriable` (neither of the last two are words allegedly)
- Will allow a single retry for scenarios failing due to selenium/appium errors

## TODO

- Add flag to disable _all_ retries
- Add flag to allow more than one retry for the above circumstances (@twometresteve opinion on this?)


bin/maze-runner
```diff
@@ -2,36 +2,45 @@
...
 require_relative '../lib/maze/network'
 require_relative '../lib/maze/proxy'
 require_relative '../lib/maze/reflective_servlet'
+require_relative '../lib/maze/retry_handler'
 require_relative '../lib/maze/runner'
+require_relative '../lib/maze/sauce_labs_utils'
 require_relative '../lib/maze/server'
 require_relative '../lib/maze/store'
 
+require_relative '../lib/maze/option'
```
twometresteve, line 26: Is the intention here just to make the list more ordered, or is it important that some appear before others?

twometresteve (commented):
> * Add flag to allow more than one retry for the above circumstances (@twometresteve opinion on this?)

Could be a good feature, but I would let this settle and see how it performs before adding that.

lib/maze/retry_handler.rb
```diff
@@ -0,0 +1,70 @@
...
+        '@retryable',
+        '@retriable'
+      ]
+
+      # Determines whether a failed test_case should be restarted
+      #
+      # @param test_case [Cucumber::RunningTestCase] The current test_case or scenario
+      # @param event [Cucumber::Core::Event] The triggering event
+      def should_retry?(test_case, event)
+        return false unless not_retried_previously?(test_case)
```
twometresteve, line 29: Suggest avoiding a double-negative here by changing to `retried_previously?` and flipping the `unless`.

bin/maze-runner
Cawllec: More an ordering things - I don't believe the order makes much difference as nothing is invoked until the rest of the script is run

lib/features/support/internal_hooks.rb
```diff
@@ -51,6 +51,10 @@
   Maze.hooks.call_before_all
 end
 
+InstallPlugins do |cucumber_config|
```
twometresteve, line 54: Looks like is should be singular:
```
undefined method `InstallPlugins' for main:Object
--
  | Did you mean?  InstallPlugin (NoMethodError)
```

Cawllec: That should be resolved now

PR #307, opened 2021-11-02
Make enable-retries consistent with other options

imjoehaines:
## Goal

This was using an underscore, meaning it needed to be provided as `--enable_retries` instead of using hyphens like other options (`--enable-retries`)


Cawllec (approved):

twometresteve:
It's a bug, but we should include a CHANGELOG entry to detail the fix.

PR #309, opened 2021-11-04
Retries: Add AppiumElementNotFound error as an allowable retry

Cawllec:
## Goal

Adds a new error thrown instead of an assertion when appium cannot find certain elements.
When this happens commonly the app only needs to be restarted, rather than failing the tests in their entirety, so said error has also been added to the list that allows a test to be restarted.

Also renames `selenium` errors and methods in the retry functionality to `driver` to account for appium specific errors being included.

## Tests

Updated unit tests.
Working demonstration of functionality can be seen [here](https://buildkite.com/bugsnag/maze-runner/builds/1873#1859763d-c293-4bb5-bbde-ffedb76aebd6/326-344)


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# TBD
+
+## Enhancements
+
+- Add automatic retries when appium element cannot be found [#TBD]()
```
twometresteve, line 5: Remember the PR link

twometresteve (approved):
Looks good!

PR #310, opened 2021-11-05
Feature flag steps

Cawllec:
## Goal

Adds specific steps to be used when verify featureFlag presence in a variety of ways.

Example of usage (and practical test) can be seen here: https://buildkite.com/bugsnag/bugsnag-js-browser/builds/1745#188a45c2-791e-4477-9a5a-6677fe94bb77/493-494


lib/features/steps/feature_flag_steps.rb
```diff
@@ -0,0 +1,64 @@
...
+# Example:
+#   | featureFlag | variant |
+#   | my_flag_1   | var_1   |
+#   | my_flag_2   | var_2   |
+#   | my_flag_3   |         | # Variant expected is nil
+#
+# @step_input table [Cucumber::MultilineArgument::DataTable] Table of expected values
+Then('the event contains the following feature flags:') do |table|
+  featureFlags = Maze::Helper.read_key_path(Maze::Server.errors.current[:body], "events.0.featureFlags")
+  assert_false(featureFlags.nil? || featureFlags.empty?, "Feature flags were nil or empty: #{featureFlags}")
```
imjoehaines, line 44: Same as above with the strictness of this

lib/features/steps/feature_flag_steps.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  expected_features = table.hashes
+  assert(featureFlags.size == expected_features.size, "Expected #{expected_features.size} features, found #{featureFlags.size}")
```
imjoehaines, line 47: It'd be good to see what the features actually were when this fails

lib/features/steps/feature_flag_steps.rb
```diff
@@ -0,0 +1,64 @@
...
+  expected_features.each do |expected|
+    if expected['variant'].nil? || expected['variant'].empty?
+      assert(featureFlags.one? { |flag| flag['featureFlag'] == expected['featureFlag'] && flag['variant'].nil?})
```
imjoehaines, line 50: Same thing about how strict this should be

lib/features/steps/feature_flag_steps.rb
```diff
@@ -0,0 +1,64 @@
...
+end
+
+# Verifies a feature flag with a specific variant is uniquely present
+#
+# @step_input flag_name [String] The featureFlag value expected
+# @step_input variant [String] The variant value expected
+Then('the event contains the feature flag {string} with variant {string}') do |flag_name, variant|
+  featureFlags = Maze::Helper.read_key_path(Maze::Server.errors.current[:body], "events.0.featureFlags")
+  assert_false(featureFlags.nil? || featureFlags.empty?, "Feature flags were nil or empty: #{featureFlags}")
+  assert(featureFlags.one? { |flag| flag['featureFlag'] == flag_name && flag['variant'] == variant })
```
imjoehaines, line 18: Can `variant` ever be anything other than a string? I think Cucumber will always provide a string when matching `{string}`, but I'm not 100% sure

e.g. if `flag['variant']` was an integer `123`, this should fail:

```gherkin
Then the event contains the feature flag "name" with variant "123"
```

lib/features/steps/feature_flag_steps.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+  end
+end
+
+# Verifies a feature flag a specific name is not present, regardless of variant
+#
+# @step_input flag_name [String] The featureFlag value not expected
+Then('the event does not contain the feature flag {string}') do |flag_name|
+  featureFlags = Maze::Helper.read_key_path(Maze::Server.errors.current[:body], "events.0.featureFlags")
+  assert_false(featureFlags.nil? || featureFlags.empty?, "Feature flags were nil or empty: #{featureFlags}")
```
imjoehaines, line 62: Same as above with the strictness of this

lib/features/steps/feature_flag_steps.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  expected_features = table.hashes
+  assert(featureFlags.size == expected_features.size, "Expected #{expected_features.size} features, found #{featureFlags.size}")
+  expected_features.each do |expected|
+    if expected['variant'].nil? || expected['variant'].empty?
+      assert(featureFlags.one? { |flag| flag['featureFlag'] == expected['featureFlag'] && flag['variant'].nil?})
+    else
+      assert(featureFlags.one? { |flag| flag['featureFlag'] == expected['featureFlag'] && flag['variant'] == expected['variant']})
```
imjoehaines, line 52: These assertions don't have a message when they fail; it'd be good to log the actual feature flags for debugging

imjoehaines (changes requested):
Left a bunch of comments but they're pretty much all one of two things:

1. the `event.featureFlags` has to be missing or an array; `null` isn't allowed: https://github.com/bugsnag/bugsnag-event-worker/blob/36516d2f55b72002be05c32c0b3ad252faab56ab/lib/event/event.coffee#L411-L414
2. the `variant` has to be missing or a string; it also doesn't allow `null`: https://github.com/bugsnag/bugsnag-event-worker/blob/36516d2f55b72002be05c32c0b3ad252faab56ab/lib/event/event.coffee#L427-L429

lib/features/steps/feature_flag_steps.rb
Cawllec: I read in the ED that variant should always be assumed to be a string, is this not the case?

imjoehaines: Yes, sorry that was what I was trying to say — it should be impossible for the `variant` variable to be anything other than a string, otherwise the `==` could pass e.g. if they were both integers

I think it's fine as-is, just wanted to check how cucumber handles `{string}` with values that could be considered non-strings (`123`/`null`/etc&hellip;)

lib/features/steps/feature_flag_steps.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+# @!group Feature flag steps
+
+# Verifies that the are no feature flags present
+Then('the event has no feature flags') do
+  featureFlags = Maze::Helper.read_key_path(Maze::Server.errors.current[:body], "events.0.featureFlags")
```
twometresteve, line 7: Variable naming would be more consistent as `feature_flags` and not violate Rubocop, if we turned it on.

lib/features/steps/feature_flag_steps.rb
twometresteve: Cucumber is type sensitive, e.g. for :
```
{
  "values": {
    "string":"123",
    "integer":123"
  }
}

This both pass:
```
    And the error payload field "values.string" equals "123"
    And the error payload field "values.integer" equals 123
```

and these both fail:
```
    And the error payload field "values.string" equals 123
    And the error payload field "values.integer" equals "123"
```

Comment (twometresteve) [reply lib/features/steps/feature_flag_steps.rb:18]: ...so the fact that the step uses `variant {string}` and not `variant {int}` means it's ok.

File: lib/features/steps/feature_flag_steps.rb [hunk 8]
```diff
@@ -0,0 +1,144 @@
...
+#
+# @step_input flag_name [String] The featureFlag value not expected
+Then('the event does not contain the feature flag {string}') do |flag_name|
+  steps %(
+    Then event 0 does not contain the feature flag #{flag_name}
+  )
+end
+
+def verify_feature_flags_with_table(event, table)
+  assert(has_feature_flags?(event), "Expected feature flags were not present in event #{event_id}: #{event}")
```
imjoehaines, line 117: `event_id` isn't in scope here

```suggestion
  assert(has_feature_flags?(event), "Expected feature flags were not present in event: #{event}")
```

lib/features/steps/feature_flag_steps.rb
```diff
@@ -0,0 +1,144 @@
...
+Then('the event contains the feature flag {string} with no variant') do |flag_name|
+  steps %(
+    Then event 0 contains the feature flag #{flag_name} with no variant
+  )
+end
+
+# Verifies that a number of feature flags outlined in a table are all present and unique in the given event
+#
+# The DataTable used for this step should have `featureFlag` and `variant` columns, containing the appropriate
+# values.  For missing or `null` variants, leave the `variant` column blank.
```
imjoehaines, line 71: Null isn't allowed as a value, so I don't think we need to mention it here:

```suggestion
# values. For flags without a variant, leave the `variant` column blank.
```

lib/features/steps/feature_flag_steps.rb
```diff
@@ -0,0 +1,144 @@
...
+#
+# Example:
+#   | featureFlag | variant |
+#   | my_flag_1   | var_1   |
+#   | my_flag_2   | var_2   |
+#   | my_flag_3   |         | # Variant expected is nil
```
imjoehaines, line 77:
```suggestion
#   | my_flag_3   |         | # 'my_flag_3' should not have a variant
```

lib/features/steps/feature_flag_steps.rb
```diff
@@ -0,0 +1,144 @@
...
+# @step_input flag_name [String] The featureFlag value expected
+# @step_input variant [String] The variant value expected
+Then('event {int} contains the feature flag {string} with variant {string}') do |event_id, flag_name, variant|
+  event = Maze::Helper.read_key_path(Maze::Server.errors.current[:body], "events.#{event_id}")
+  assert(has_feature_flags?(event), "Expected feature flags were not present in event #{event_id}: #{event}")
+  feature_flags = event['featureFlags']
+  assert(
+    feature_flags.one? { |flag| flag['featureFlag'] == flag_name && flag['variant'] == variant },
+    "Feature flag: #{flag_name} with variant: #{variant} not found. Present flags: #{feature_flags}"
+  )
```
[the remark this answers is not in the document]
imjoehaines: Is it possible to write some tests for these steps? The logic is quite complicated so it'd be good to prove they work & have a record of how they should work

lib/features/steps/feature_flag_steps.rb
```diff
@@ -0,0 +1,144 @@
...
+# Verifies a feature flag with no variant (either null or missing) is uniquely present in a given event
+#
+# @step_input event_id [Integer] The id of the event in the payloads array
+# @step_input flag_name [String] The featureFlag value expected
+Then('event {int} contains the feature flag {string} with no variant') do |event_id, flag_name|
+  event = Maze::Helper.read_key_path(Maze::Server.errors.current[:body], "events.#{event_id}")
+  assert(has_feature_flags?(event), "Expected feature flags were not present in event #{event_id}: #{event}")
+  feature_flags = event['featureFlags']
+  assert(
+    feature_flags.one? { |flag| flag['featureFlag'] == flag_name && !flag.has_key?('variant') },
```
imjoehaines, line 54: This could allow duplicates if there are multiple flags with the same `featureFlag` but only one has no `variant`

lib/features/steps/feature_flag_steps.rb
```diff
@@ -0,0 +1,144 @@
...
+def verify_feature_flags_with_table(event, table)
+  assert(has_feature_flags?(event), "Expected feature flags were not present in event #{event_id}: #{event}")
+  feature_flags = event['featureFlags']
+
+  expected_features = table.hashes
+  assert(feature_flags.size == expected_features.size, "Expected #{expected_features.size} features, found #{feature_flags}")
+  expected_features.each do |expected|
+    if expected['variant'].nil? || expected['variant'].empty?
+      assert(
+        feature_flags.one? { |flag| flag['featureFlag'] == expected['featureFlag'] && !flag.has_key?('variant') },
```
imjoehaines, line 125: same thing about duplicates

lib/features/steps/feature_flag_steps.rb
```diff
@@ -0,0 +1,144 @@
...
+        "Feature flag: #{flag_name} was present with a variant in: #{feature_flags}"
+      )
+    else
+      assert(
+        feature_flags.one? { |flag| flag['featureFlag'] == expected['featureFlag'] && flag['variant'] == expected['variant'] },
```
imjoehaines, line 130: same thing about duplicates

imjoehaines (changes requested):

Cawllec:
I believe I've addressed the duplication problem - I had to decouple some of the assertions but hopefully it still reads logically.  I've added some basic integration tests running [here](https://buildkite.com/bugsnag/maze-runner/builds/1886#f07664af-6049-42f2-953b-258419b171bd/1199) to hopefully give some confidence.

PR #311, opened 2021-11-05
Retry: Browser driver fix

Cawllec:
## Goal

The Browser driver does not have a `restart` method, so instead we should call the `refresh` method (to refresh the webpage) when a driver error happens.

This likely won't fix any critical driver issues, but if the webpage doesn't load correctly for some reason it may help, and it'll also float the original error out.

### DIFF
--- a/lib/maze/retry_handler.rb
+++ b/lib/maze/retry_handler.rb
@@ -31,7 +31,11 @@ module Maze
 
         if retry_on_driver_error?(event)
           $logger.warn "Retrying #{test_case.name} due to driver error: #{event.result.exception}"
-          Maze.driver.restart
+          if Maze.driver.is_a?(Maze::Driver::Appium) || Maze.driver.is_a?(Maze::Driver::ResilientAppium)
+            Maze.driver.restart
+          elsif Maze.driver.is_a?(Maze::Driver::Browser)
+            Maze.driver.refresh
+          end
           increment_retry_count(test_case)
           true
         elsif retry_on_tag?(test_case)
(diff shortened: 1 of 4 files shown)


lib/maze/retry_handler.rb
```diff
@@ -31,7 +31,11 @@ def should_retry?(test_case, event)
 
         if retry_on_driver_error?(event)
           $logger.warn "Retrying #{test_case.name} due to driver error: #{event.result.exception}"
-          Maze.driver.restart
+          if Maze.driver.is_a?(Maze::Driver::Appium) || Maze.driver.is_a?(Maze::Driver::ResilientAppium)
+            Maze.driver.restart
+          elsif Maze.driver.is_a?(Maze::Driver::Browser)
+            Maze.driver.refresh
```
imjoehaines, line 37: Does `Maze::Driver::Browser` have a `refresh` method? I don't see it or any `method_missing` shenanigans

Cawllec: Yeah, added a passthrough for that

imjoehaines: Is it worth even calling it `restart`? I guess it's kind of misleading that way, but means the `is_a` checks aren't needed. I'm happy either way though

imjoehaines (approved):

twometresteve (approved):

lib/maze/retry_handler.rb
Cawllec: I think it's worth being explicit that it's a refresh, no a restart, although I'm not opposed to changing it in the future

PR #314, opened 2021-11-17
Allow multiple `--device` options to be specified

Cawllec:
## Goal

Allows multiple `--device` options when running maze-runner.  These options will then be iterated through when attaching to the appropriate device farm, moving onto the next if one fails, allowing us to specify several appropriate devices for a test.

## Testing

Unit tests cover the basic functionality of the added/refactored methods

Not tested in anger - I'd suggest using this release in a test job against android/cocoa before merging.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Enhancements
 
 - All `--app` to be specified in a file using `@` prefix [#312](https://github.com/bugsnag/maze-runner/pull/312)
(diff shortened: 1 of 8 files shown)


twometresteve (approved):
Looks good to me - I've seen this work in anger having randomly chosen `ANDROID_4_4_SAMSUNG_GALAXY_NOTE_4` which turns out to be an invalid device and gets rejected immediately.

I just a couple of comments, but I'd be happy to see this merged and released based on the testing I've done.

PR #323, opened 2021-12-07
Correct capability to disableAnimation

twometresteve:
## Goal

Correct capability name to `disableAnimation`.  

See https://www.browserstack.com/app-automate/capabilities.

## Design

Android 6 and below doesn't support this capability and BrowserStack will raise an error if you try to set it. 

## Testing

In addition to CI, I have manually run tests on Android against versions 5,6 and 7 (i.e. across the v6 boundary) and on iOS versions 10 and 15.

### DIFF
--- a/lib/maze/capabilities.rb
+++ b/lib/maze/capabilities.rb
@@ -12,7 +12,6 @@ module Maze
           'browserstack.console' => 'errors',
           'browserstack.localIdentifier' => local_id,
           'browserstack.local' => 'true',
-          'disableAnimations' => 'true',
           'noReset' => 'true'
         }
         capabilities.merge! BrowserStackDevices::DEVICE_HASH[device_type]
--- a/lib/maze/capabilities.rb
+++ b/lib/maze/capabilities.rb
@@ -12,7 +12,7 @@ module Maze
           'browserstack.console' => 'errors',
           'browserstack.localIdentifier' => local_id,
           'browserstack.local' => 'true',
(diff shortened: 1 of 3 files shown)


Cawllec (dismissed):

Cawllec:
Oops, put that on before looking at the tests.  Looks like we'll need to disable that capability for Android 6 and below

twometresteve:
> Oops, put that on before looking at the tests. Looks like we'll need to disable that capability for Android 6 and below

Well I wasn't expecting that!  I'm tempted just to remove it completely, given that it's been doing nothing all this time anyway?

Cawllec:
I think it's worth using, as it should speed things up, but yeah, not overly fussed

lib/maze/capabilities.rb
```diff
@@ -12,7 +12,7 @@ def for_browser_stack_device(device_type, local_id, appium_version, capabilities
           'browserstack.console' => 'errors',
           'browserstack.localIdentifier' => local_id,
           'browserstack.local' => 'true',
-          'disabledAnimations' => 'true',
+          'disableAnimations' => 'true',
```
nickdowell, line 15: I've been noticing Appium warnings about strings being passed instead of booleans, e.g.
`Capability 'noReset' changed from string to boolean. This may cause unexpected behavior`

```suggestion
          'disableAnimations' => true,
```

Cawllec (approved):

PR #324, opened 2021-12-09
Use the app path instead of app name

Cawllec:
## Goal

Preheat the application so we no-longer need to copy the app into the `/Applications` directory

## Testing

[This run](https://buildkite.com/bugsnag/bugsnag-cocoa/builds/4230) shows the functionality working
[This run](https://buildkite.com/bugsnag/bugsnag-cocoa/builds/4234) will be required to work with the cocoa library, and should cover all of the desired macOS versions.

### DIFF
--- a/lib/maze/hooks/appium_hooks.rb
+++ b/lib/maze/hooks/appium_hooks.rb
@@ -31,7 +31,7 @@ module Maze
           pp "Open app"
           pp config.app_path
           pp config.app
-          Maze::Runner.run_command(`open #{config.app_path}`) if config.os == 'macos'
+          Maze::Runner.run_command("open #{config.app_path}") if config.os == 'macos'
 
           # Attempt to start the local appium server
           appium_uri = URI(config.appium_server_url)
--- a/lib/maze/hooks/appium_hooks.rb
+++ b/lib/maze/hooks/appium_hooks.rb
@@ -28,7 +28,7 @@ module Maze
                                                    config.access_key
         when :local
           # Open the app beforehand to ensure it runs correctly and is accessible
-          Maze::Runner.run_command(`open #{fixture_dir}/#{app_file}`) if config.os == 'macos'
+          Maze::Runner.run_command(`open #{config.app_path}`) if config.os == 'macos'
 
           # Attempt to start the local appium server
           appium_uri = URI(config.appium_server_url)
--- a/lib/maze/hooks/appium_hooks.rb
+++ b/lib/maze/hooks/appium_hooks.rb
@@ -28,10 +28,6 @@ module Maze
                                                    config.access_key
         when :local
           # Open the app beforehand to ensure it runs correctly and is accessible
-          pp "Open app"
-          pp config.app_path
-          pp config.app
-          Maze::Runner.run_command("open #{config.app_path}") if config.os == 'macos'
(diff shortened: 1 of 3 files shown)


lib/maze/hooks/appium_hooks.rb
```diff
@@ -27,6 +27,9 @@ def before_all
                                                    config.username,
                                                    config.access_key
         when :local
+          # Open the app beforehand to ensure it runs correctly and is accessible
+          Maze::Runner.run_command("/System/Library/Frameworks/CoreServices.framework/Versions/Current/Frameworks/LaunchServices.framework/Versions/Current/Support/lsregister -f -R -trusted #{config.app_path}") if config.os == 'macos'
```
nickdowell, line 31: I've tested this locally and found that, unfortunately, this step alone is not enough - macOS continues to open the "older" version of the fixture 😞 

Adding another step to reset the LaunchServices database works around this, but it unregisters **all** apps and so could cause trouble elsewhere.

```
/System/Library/Frameworks/CoreServices.framework/Versions/A/Frameworks/LaunchServices.framework/Versions/A/Support/lsregister -kill -r -v -apps u,s,l
```

Another approach we could take is to define a `Maze.driver.launch_app` method for macOS (I believe this is currently not implemented) that simply runs `open #{config.app_path}"`

nickdowell: The best documentation I've found for `lsregister` is https://eclecticlight.co/2019/03/25/lsregister-a-valuable-undocumented-command-for-launchservices/

twometresteve: > Another approach we could take is to define a `Maze.driver.launch_app` method for macOS (I believe this is currently not implemented) that simply runs `open #{config.app_path}"`

If that works I think it would clean up the code in the cocoa repo a bit?

Cawllec: > Another approach we could take is to define a Maze.driver.launch_app method for macOS (I believe this is currently not implemented) that simply runs open #{config.app_path}"
I tried this, and didn't think it had worked.  I'll make the change, if you could test it afterwards that'd be great (I'll let you know on slack)

> > Another approach we could take is to define a `Maze.driver.launch_app` method for macOS (I believe this is currently not implemented) that simply runs `open #{config.app_path}"`
> 
> If that works I think it would clean up the code in the cocoa repo a bit?

How do you mean, removing the unzipping process in `env.rb`?

twometresteve: I just meant a couple of places where we call `launch_app` on iOS and do something else on macOS, but it may not apply in those situation to be honest.  E.g. step `I relaunch the app after a crash`.

Cawllec: This would only be relevant before launching the app for the first time, rather than every re-launch

nickdowell (approved):

nickdowell (commented):
I was too quick to approve this.

Turns out that `open #{config.app_path}` prior to starting appium does not fix the underlying issue - it doesn't guarantee that appium will open the correct copy of the app 😞

twometresteve:
> I was too quick to approve this.
> 
> Turns out that `open #{config.app_path}` prior to starting appium does not fix the underlying issue - it doesn't guarantee that appium will open the correct copy of the app 😞

Ok, let's close the PR for now.  A different approach is being taken in bugsnag-cocoa and this remain in the list for reference should something change in time.

Thanks for all the energy that you put into this, @Cawllec .

PR #328, opened 2021-12-17
Refactor remaining codebase to use Maze.check

twometresteve:
## Goal

Follow-up to #327, refactoring the remaining codebase to use the new `Maze.check` and adding a few methods that were missed at first.

## Changeset

Refactoring and new methods added.

## Tests

I have successfully tested local branches for the `bugsnag-js`, `bugsnag-cocoa` and `bugsnag-android` repos with updates to use this.

### DIFF
--- a/lib/features/steps/error_reporting_steps.rb
+++ b/lib/features/steps/error_reporting_steps.rb
@@ -184,15 +184,13 @@ end
 
 Then('the event {string} string is empty') do |keypath|
   value = Maze::Helper.read_key_path(Maze::Server.errors.current[:body], keypath)
-  Maze.check.block("The #{keypath} is not empty: '#{value}'") do
-    value.nil? || value.empty?
-  end
+  Maze.check.true(value.nil? || value.empty?, "The #{keypath} is not empty: '#{value}'")
 end
 
 Then('the event {string} is greater than {int}') do |keypath, int|
   value = Maze::Helper.read_key_path(Maze::Server.errors.current[:body], "events.0.#{keypath}")
   Maze.check.false(value.nil?, "The event #{keypath} is nil")
-  Maze.check.true(value > int)
+  Maze.check.operator(value, :>, int)
 end
 
 # Tests whether a value in the first exception of the first event entry starts with a string.
--- a/lib/features/steps/error_reporting_steps.rb
+++ b/lib/features/steps/error_reporting_steps.rb
@@ -97,7 +97,7 @@ Then('the error payload contains the payloadVersion {string}') do |payload_versi
   body_set = payload_version == body_version
   event_version = Maze::Helper.read_key_path(Maze::Server.errors.current[:body], 'events.0.payloadVersion')
   event_set = payload_version == event_version
-  assert_true(
+  Maze.check.true(
     body_set || event_set,
     "The payloadVersion was not the expected value of #{payload_version}. " \
     "#{body_version} found in body, #{event_version} found in event"
@@ -184,15 +184,15 @@ end
 
 Then('the event {string} string is empty') do |keypath|
   value = Maze::Helper.read_key_path(Maze::Server.errors.current[:body], keypath)
-  assert_block("The #{keypath} is not empty: '#{value}'") do
+  Maze.check.block("The #{keypath} is not empty: '#{value}'") do
     value.nil? || value.empty?
   end
 end
 
 Then('the event {string} is greater than {int}') do |keypath, int|
   value = Maze::Helper.read_key_path(Maze::Server.errors.current[:body], "events.0.#{keypath}")
-  assert_false(value.nil?, "The event #{keypath} is nil")
-  assert_true(value > int)
+  Maze.check.false(value.nil?, "The event #{keypath} is nil")
+  Maze.check.true(value > int)
 end
 
 # Tests whether a value in the first exception of the first event entry starts with a string.
@@ -307,13 +307,13 @@ end
 # @param payload_value [Any] The thread identifier value
 def validate_error_reporting_thread(payload_key, payload_value)
   threads = Maze::Server.errors.current[:body]['events'].first['threads']
-  assert_kind_of Array, threads
+  Maze.check.kind_of Array, threads
   count = 0
 
   threads.each do |thread|
     count += 1 if thread[payload_key].to_s == payload_value && thread['errorReportingThread'] == true
   end
-  assert_equal(1, count)
+  Maze.check.equal(1, count)
 end
 
 # Tests whether an event has the correct attributes we'd expect for un/handled events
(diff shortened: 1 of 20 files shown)


lib/features/steps/payload_steps.rb
```diff
@@ -108,7 +109,7 @@
 Then('the {word} payload field {string} is greater than {int}') do |request_type, field_path, int_value|
   list = Maze::Server.list_for(request_type)
   value = Maze::Helper.read_key_path(list.current[:body], field_path)
-  assert_kind_of Integer, value
+  Maze.check.kind_of Integer, value
   assert(value > int_value, "The payload field '#{field_path}' (#{value}) is not greater than '#{int_value}'")
```
imjoehaines, line 113:
```suggestion
  Maze.check.operator(value, :>, int_value, "The payload field '#{field_path}' (#{value}) is not greater than '#{int_value}'")
```

lib/features/steps/payload_steps.rb
```diff
@@ -180,7 +181,7 @@
 Then('the {word} payload field {string} is a non-empty array') do |request_type, field|
   list = Maze::Server.list_for(request_type)
   value = Maze::Helper.read_key_path(list.current[:body], field)
-  assert_kind_of Array, value
+  Maze.check.kind_of Array, value
   assert(value.length.positive?,
```
imjoehaines, line 185:
```suggestion
  Maze.check.true(value.length.positive?,
```

lib/maze/checks/assert_check.rb
```diff
@@ -39,6 +39,20 @@ def operator(operand1, operator, operand2, message = nil)
...
+        assert_block(message, &block)
+      end
+
+      def include(collection, object, message = nil)
+        assert_include(collection, object, message)
+      end
+      alias includes include
+
+      def refute_include(collection, object, message = nil)
+        Minitest::Assertions.refute_includes(collection, object, message)
```
imjoehaines, line 53: [`refute_includes` does exist in Test Unit v3+](https://ruby-doc.org/stdlib-3.0.0/libdoc/test-unit/rdoc/Test/Unit/Assertions.html#method-i-refute_includes), we might just need to update?

If we do that then we may be able to remove Minitest as a dependency of Maze Runner entirely

imjoehaines: It might also make sense to change the step that uses this to not use it, if it's just here for one or two tests

imjoehaines (changes requested):
There are still a fair few places using `assert` that could be `Maze.check.true` instead, unless they weren't changed on purpose? I've commented on some, but there are others in places that GitHub doesn't let me comment on

Cawllec:
Along with what Joe has reviewed, is standard `assert` the only one not covered by our Maze checks class now?

lib/maze/checks/assert_check.rb
twometresteve: I certainly see some benefit in removing the minitest dependency if we can.  I'm slightly worried about breaking pipelines by removing the dependency, but that wouldn't be too hard to mop up.

I'd also be very happy to see `assert_not_include` used rather than a refute.

twometresteve:
> 

Th

twometresteve:
> Along with what Joe has reviewed, is standard `assert` the only one not covered by our Maze checks class now?

There are quite a few methods in [Test::Unit::Assertions](https://www.rubydoc.info/gems/test-unit/3.5.1/Test/Unit/Assertions#assert-instance_method) that aren't included, but I haven't aimed for completeness in order to keep the amount of code to maintain to a minimum.  But of course we can add more if we need them.

lib/maze/checks/assert_check.rb
imjoehaines: 👍 `assert_not_includes` sounds better to me too and is more consistent with `assert_nil`/`assert_not_nil`

twometresteve:
> There are still a fair few places using `assert` that could be `Maze.check.true` instead, unless they weren't changed on purpose? I've commented on some, but there are others in places that GitHub doesn't let me comment on

I used my IDE to search for all instances and I deliberately ignored instances in unit tests.  Unfortunately I didn't notice that there were more than 100 of those and my IDE only displays the first 100!

lib/features/steps/error_reporting_steps.rb
```diff
@@ -184,15 +184,15 @@
 
 Then('the event {string} string is empty') do |keypath|
   value = Maze::Helper.read_key_path(Maze::Server.errors.current[:body], keypath)
-  assert_block("The #{keypath} is not empty: '#{value}'") do
+  Maze.check.block("The #{keypath} is not empty: '#{value}'") do
     value.nil? || value.empty?
   end
```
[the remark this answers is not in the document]
twometresteve: Agreed, neater.

lib/features/steps/error_reporting_steps.rb
```diff
@@ -184,15 +184,15 @@
...
 end
 
 Then('the event {string} is greater than {int}') do |keypath, int|
   value = Maze::Helper.read_key_path(Maze::Server.errors.current[:body], "events.0.#{keypath}")
-  assert_false(value.nil?, "The event #{keypath} is nil")
-  assert_true(value > int)
+  Maze.check.false(value.nil?, "The event #{keypath} is nil")
+  Maze.check.true(value > int)
```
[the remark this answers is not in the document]
twometresteve: Yes, agreed.

lib/features/steps/payload_steps.rb
twometresteve: 👍

lib/features/steps/payload_steps.rb
```diff
@@ -120,7 +121,7 @@
 Then('the {word} payload field {string} is less than {int}') do |request_type, field_path, int_value|
   list = Maze::Server.list_for(request_type)
   value = Maze::Helper.read_key_path(list.current[:body], field_path)
-  assert_kind_of Integer, value
+  Maze.check.kind_of Integer, value
   fail_message = "The #{request_type} payload field '#{field_path}' (#{value}) is not less than '#{int_value}'"
   assert(value < int_value, fail_message)
```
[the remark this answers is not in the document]
twometresteve: 👍

lib/features/steps/proxy_steps.rb
```diff
@@ -1,5 +1,7 @@
 # @!group Proxy steps
 
+include Test::Unit::Assertions
+
```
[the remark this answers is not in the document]
twometresteve: No, they should also be removed once `Maze.check` is used throughout.

imjoehaines (approved):

Cawllec (approved):
LGTM

PR #329, opened 2021-12-20
Add timer mechanism

twometresteve:
## Goal

Add general simple timer mechanism, with specific application to select Appium operations.

## Design

This will help pipelines understand where automation operations are taking time, hopefully allowing us/them to make performance improvements.

## Changeset

New `Timer` and `Timers` classes, with use woven into the Cucumber hooks and Appium driver.

## Tests

Covered by unit tests and by inspection of the CI build for this PR.

### DIFF
--- a/lib/maze/timers.rb
+++ b/lib/maze/timers.rb
@@ -5,7 +5,6 @@ module Maze
     attr_accessor :total
 
     def initialize
-      @started = 0
       @total = 0
     end
 
--- a/lib/maze/timers.rb
+++ b/lib/maze/timers.rb
@@ -52,7 +52,7 @@ module Maze
 
       $logger.info 'Timer totals:'
       @timers.sort.each do |name, timer|
-        $logger.info "#{name}: #{timer.total}"
+        $logger.info "  #{name}: #{timer.total}"
       end
     end
   end
--- a/lib/maze/timers.rb
+++ b/lib/maze/timers.rb
@@ -48,6 +48,8 @@ module Maze
     end
 
     def report
+      return if @timers.empty?
+
       $logger.info 'Timer totals:'
       @timers.sort.each do |name, timer|
         $logger.info "#{name}: #{timer.total}"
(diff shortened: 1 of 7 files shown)


lib/maze/timers.rb
```diff
@@ -0,0 +1,61 @@
...
+    attr_accessor :total
+
+    def initialize
+      @started = 0
+      @total = 0
+      @running = false
+    end
+
+    def run
+      raise 'Timer already running' if @running
```
twometresteve, line 14: I put these in to make it obvious if a client tries to use it in the wrong way, but it could be asking for trouble if there are error conditions that, for example, prevent a timer being stopped after it is run.  I might be better logging and error and flagging at the end that the timer should not be trusted?

imjoehaines: Would a block-based API work? That way you can avoid the timer getting into an invalid state e.g.

```rb
class Timer
  attr_accessor :total

  def initialize
    @total = 0
  end

  def time(&block)
    start = Time.now

    block.call

    @total += Time.now - start
  end
end
```

It might also be useful to return the result of the block so you can measure code that needs to return something

twometresteve: As I was putting the calls in I knew there must be a better way of doing it.  Would the last need need to go in an `ensure` block?

imjoehaines: Yeah, it probably should do. That would also make it return the result of the block as well

twometresteve: Done, that makes it much cleaner.

lib/maze/timers.rb
```diff
@@ -0,0 +1,52 @@
+module Maze
+
+  # A simple run/stop timer
+  class Timer
+    attr_accessor :total
+
+    def initialize
+      @started = 0
```
Cawllec, line 8: Can this be removed, it doesn't seem to be used anywhere

Cawllec (approved):
Minor query, otherwise LGTM

lib/maze/timers.rb
twometresteve: Yes, you're quite right.

imjoehaines (approved):

PR #330, opened 2021-12-20
Set capabilities to speed up App Hang tests locally

twometresteve:
## Goal

Reduce the time the iOS App Hang scenarios take to run locally.  For me it reduced the time from 8 minutes to 5.5.

## Design

These capabilities are set by default on BrowserStack, providing some confidence that they won't break tests.

## Tests

Tested will a full test run locally.


nickdowell (approved):
Making our local runs use the same Appium capabilities as BS makes a lot of sense 👍

PR #332, opened 2022-01-11
Add command endpoint

twometresteve:
## Goal

Add `/command` endpoint.

## Design

Allows clients to add any "command" they wish to use onto a list, which will be served to the test fixture after making a GET request to the `/command` endpoint.

## Changeset

I've also done a bit of refactoring to push the Servlets into their own module with a `BaseServlet` to factor out some common behavior.  Unfortunately GitHub show the whole file as new rather than a move and select lines having changed.

## Tests

Covered by existing unit tests for the `RequestList`, with additional testing having been performed locally with `bugsnag-cocoa`.


lib/maze/servlets/command_servlet.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+require 'json'
+
+module Maze
+  module Servlets
+
+    # Receives log requests sent from the test fixture
```
nickdowell, line 8: I don't think that's what this does 😈

lib/maze/servlets/command_servlet.rb
```diff
@@ -0,0 +1,44 @@
...
+        $logger.info 'Asked for a command'
+
+        response.header['Access-Control-Allow-Origin'] = '*'
+
+        commands = Maze::Server.commands
+        if commands.empty?
+          response.body = 'No commands to provide'
+          response.status = 400
+        else
+          response.body = JSON.pretty_generate(commands.current)
```
nickdowell, line 24: Does `.current` get the head or tail of the list?

I'm not planning to queue up multiple commands, but curious about the behaviour.

lib/maze/servlets/command_servlet.rb
twometresteve: Corrected

lib/maze/servlets/command_servlet.rb
twometresteve: It should work in the natural way - add 3 things to queue and GET requests to `/commands` will return each in the order that they we added.  It's exactly the same mechanism that's used when expecting more than one error/session, with the `I discard the oldest log` Cucumber step moving to the next in the list.

nickdowell (approved):

lib/maze/server.rb
```diff
@@ -2,10 +2,11 @@
 
 require 'json'
 require 'webrick'
-require_relative './servlet'
-require_relative './log_servlet'
 require_relative './logger'
 require_relative './request_list'
+require_relative './servlets/command_servlet'
```
Cawllec, line 7: Is this redundant as it's already imported above: https://github.com/bugsnag/maze-runner/pull/332/files#diff-8f360d1c0c26c88707e927813ccd865245e4596f406eae23809d354b2ed5d53cR25?

Cawllec (commented):
One small quibble, but could we add some more formal tests to one of our integration test fixtures?  Given this will end up driving pretty much all of our mobile testing it would be worth ensuring we've got good coverage.

lib/maze/server.rb
twometresteve: Yes, I think you're right.  I think the order of requires matters and I got it wrong in `maze_runner` at first, leading to me adding these unnecessarily.  It just means adding some requires to the unit test.

twometresteve:
> One small quibble, but could we add some more formal tests to one of our integration test fixtures? Given this will end up driving pretty much all of our mobile testing it would be worth ensuring we've got good coverage.

Yes, agreed.  I've now integrated it into the Android test fixture and added a single scenario to show that multiple commands can be issues and processed in order.

test/fixtures/android-appium-test/features/steps/steps.rb
```diff
@@ -0,0 +1,11 @@
+When('I issue a command to notify a {string} handled error') do |metadata|
```
Cawllec, line 1: Are we intending to bake these steps into Maze-runner with a pre-defined structure, or are we happy for them to be implemented on a per-platform basis for now?

Cawllec (approved):
LGTM.  Left a small question, but any further work can ticketed separately.  Looking forward to the improved testing times!

lib/maze/server.rb
```diff
@@ -153,13 +159,14 @@ def start
...
-            server.mount '/sourcemap', Servlet, sourcemaps
-            server.mount '/react-native-source-map', Servlet, sourcemaps
-            server.mount '/logs', LogServlet
+            server.mount '/notify', Servlets::Servlet, errors
+            server.mount '/sessions', Servlets::Servlet, sessions
+            server.mount '/builds', Servlets::Servlet, builds
+            server.mount '/uploads', Servlets::Servlet, uploads
+            server.mount '/sourcemap', Servlets::Servlet, sourcemaps
+            server.mount '/react-native-source-map', Servlets::Servlet, sourcemaps
+            server.mount '/commands', Servlets::CommandServlet
```
nickdowell, line 168: From the client POV I think this should be `/command` (no plural) - i.e. `GET /command` to return the next command.

`GET /commands` would make sense if this returned an array of commands, which it does not.

nickdowell (changes requested):

test/fixtures/android-appium-test/features/steps/steps.rb
twometresteve: My original implementation included defined Command classes for RunScenario and StartBugsnag.  I personally liked the abstraction and it saved a couple of lines of code, but Nick didn't see the point and was concerned about having to wait for changes to Maze Runner before being able to implement new commands.  I can see it both ways, really, putting more into Maze Runner can save boilerplate code, but can also put the onus on it to do more testing and/or be more reactive to changes that are needed.

lib/maze/server.rb
twometresteve: Agreed - changed.

nickdowell (approved):

PR #333, opened 2022-01-11
Add a browser smoke test platform to CI

Cawllec:
## Goal

Adds a new test fixture based on the JS Browser fixture to CI, ensuring breaking changes will be caught before release.

## Discussion

This currently only works on Ruby 2.7.  We'll need to update Selenium Webdriver to `~> v4.1`, which will also require Appium to be updated to `~> v12`.

Unfortunately that doesn't seem like a trivial update ([as demonstrated here](https://buildkite.com/bugsnag/maze-runner/builds/2033#cff7583f-16d6-4368-a959-ae00609bf452/223-242)), but will hopefully be fairly quick.

I'd suggest we merge this PR and I'll ticket the driver updates separately.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -179,9 +179,10 @@ steps:
         run: browser-tests
         use-aliases: true
         verbose: true
-        command:
-          - --farm=bs
-          - --browser=firefox_latest
+        command: >-
+          bundle exec maze-runner
+          --farm=bs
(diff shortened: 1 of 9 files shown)


twometresteve (approved):
This is a great addition.  I recently look at updating the Selenium Webdriver (very briefly) but it seems I didn't get around to raising a ticket to do it properly - so please do go ahead and create one,

PR #336, opened 2022-01-20
Invalid request JSON formatting fix

Cawllec:
## Goal

The previous release for outputting invalid requests introduced an additional flag to determine whether we should attempt to parse the body into JSON.  This was implemented incorrectly and resulted in none of the strings being formatted nicely as JSON.

This change:
- Makes the logic clearer by renaming the `error` flag to `invalid`
- Fixes the faulty logic added in the previous addition

## Testing

This has been tested manually to ensure the formatting is fixed.  It is currently being tested to ensure invalid requests are still logged correctly, and an update will be made when that has been confirmed.


Cawllec:
I've confirmed through local running that invalid requests are still output as desired

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 6.9.2 - 2022/01/18
```
twometresteve, line 1: Correct the date before release.

twometresteve (approved):
Just the release date to update 👍

twometresteve (approved):

PR #338, opened 2022-03-01
Retry Failed BrowserStack App Uploads

joshedney:
## Goal

Retry uploading the app to BrowserStack for a max of 3 times or until we get a success.

## Design

Decided to use a while loop rather than `Maze::Wait` as the upload could take longer than the timeout specified.

## Changeset

Add while loop to the `upload_app` function in the `browser_stack_utils.rb` to retry the app upload for max of 3 attempts or until we get a success

## Tests

Full CI

### DIFF
--- a/lib/maze/browser_stack_utils.rb
+++ b/lib/maze/browser_stack_utils.rb
@@ -39,7 +39,7 @@ module Maze
                 # Successful upload
                 break
               end
-            rescue
+            rescue JSON::ParserError
               $logger.error "Error: expected JSON response, received: #{body}"
             end
 
--- a/lib/maze/browser_stack_utils.rb
+++ b/lib/maze/browser_stack_utils.rb
@@ -21,13 +21,13 @@ module Maze
           request.basic_auth(username, access_key)
           request.set_form({ 'file' => File.new(expanded_app, 'rb') }, 'multipart/form-data')
 
-          wait = Maze::Wait.new(interval: 3, timeout: 30)
+          upload_tries = 0
 
-          wait_result = wait.until(&lambda do 
+          while upload_tries < 3
             res = Net::HTTP.start(uri.hostname, uri.port, use_ssl: true) do |http|
               http.request(request)
             end
-  
+
             begin
               body = res.body
               response = JSON.parse body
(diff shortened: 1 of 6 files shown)


lib/maze/browser_stack_utils.rb
```diff
@@ -21,19 +21,26 @@ def upload_app(username, access_key, app)
...
+              body = res.body
+              response = JSON.parse body
+              raise "Upload failed due to error: #{response['error']}" if response.include?('error')
+              raise "Upload failed, response did not include and app_url: #{res}" unless response.include?('app_url')
+            rescue JSON::ParserError
+              raise "Error: expected JSON response, received: #{body}"
+            end
+
+            upload_tries = 3 if !response.include?('error') 
+            upload_tries = tries.next unless !response.include?('error')
```
twometresteve, line 41: I think you could achieve the same aim with less code by:
- Adding a `break` immediately before the `rescue` line (i.e. assume success if it gets past the previous lines)
- Remove these lines in preference to simply `upload_tries += 1`

lib/maze/browser_stack_utils.rb
```diff
@@ -21,19 +21,26 @@ def upload_app(username, access_key, app)
...
+          while upload_tries < 3
+            res = Net::HTTP.start(uri.hostname, uri.port, use_ssl: true) do |http|
+              http.request(request)
+            end
 
+            begin
+              body = res.body
+              response = JSON.parse body
+              raise "Upload failed due to error: #{response['error']}" if response.include?('error')
+              raise "Upload failed, response did not include and app_url: #{res}" unless response.include?('app_url')
```
twometresteve, line 35: If either of these hit then the stack will unwind and the retry logic have no chance to take effect.  I think you could change each `raise` to `$logger.error` here and raise after the while loop if it's not succeeded (see separate comment for my suggestion).

lib/maze/browser_stack_utils.rb
```diff
@@ -21,19 +21,26 @@ def upload_app(username, access_key, app)
...
+            rescue JSON::ParserError
+              raise "Error: expected JSON response, received: #{body}"
+            end
+
+            upload_tries = 3 if !response.include?('error') 
+            upload_tries = tries.next unless !response.include?('error')
+          end
+
```
twometresteve, line 43: Coupled with my first comment, I'd suggest adding:
```
if response.include?('error') || !response.include?('app_url')
  raise "Failed to upload app after #{upload_tries} attempts"
end
```

twometresteve (approved):
LGTM, just remember to update `CHANGELOG.md` and version (after which you could release).

PR #341, opened 2022-03-18
Rename browsers.yml

twometresteve:
## Goal

Rename browsers.yml to browsers_bs.yml. as it is BrowserStack specific.

## Changeset

I've also tweaked a couple of help messages to be BrowserStack agnostic.

I haven't included a CHANGELOG.md entry for this as there is no interesting impact on users of the tool.

## Tests

Covered by the Maze Runner CI pipeline.


joshedney (approved):
LGTM!

PR #342, opened 2022-03-21
Add support for CrossBrowserTesting

twometresteve:
## Goal

Add support for running Selenium tests on CrossBrowserTesting.

## Design

Follows the existing approach for BrowserStack browsers - the usage pattern is analogous. 

## Changeset

- New option for the `--farm` switch
- Capabilities YML file to support a range of browsers
- New SmartBear utils class for running the secure tunnel (this will overlap with Bitbar once added)

## Tests

Unit tests updated.  Also ran the browser test in the repo against each browser option in the YML file.  I've not updated the pipeline to run against CBT just yet, as we need to perform some further testing before we move everything across.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,10 @@
-# TBD
+# 6.10.0 - TBD
 
 ## Enhancements
 
 - Add method to download device logs from BrowserStack [#344](https://github.com/bugsnag/maze-runner/pull/344)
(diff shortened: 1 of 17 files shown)


Cawllec (commented):
Looks good, but I think we need to add the SecureTunnel binary download to the dockerfile.ci build process.

twometresteve:
> Looks good, but I think we need to add the SecureTunnel binary download to the dockerfile.ci build process.

Ah yes, good thinking.  I was just starting on some Docker-based testing and that would have caught me out.

Cawllec (approved):

PR #343, opened 2022-03-21
[WIP] Add terminating server

Cawllec:
## Goal
Adds a server to replicate the behaviour of the pipeline when a payload is too large: Cut the connection returning a 400.
This doesn't exactly match that, as the payload of the request isn't actually read, but should do a good job of triggering the behaviour in the app.

## Tests
As of yet untested, some will be added before this is ready for review.


tomlongridge:
It would be interesting to run this against the branch on bugsnag/bugsnag-android#1633 for testing initially of your server and then adding it to @kstenerud's scenario.

Cawllec:
Ok, I believe this is ready for review now that the desired behaviour has been pinned down offline

test/fixtures/http-response/features/terminating_requests.feature
```diff
@@ -0,0 +1,15 @@
...
+        And I wait to receive an error
+        And the terminating server has received 1 requests
+
+        # This doesn't have different behaviour on Ruby, but may do on other platforms
+    Scenario: Terminating server can be started and respond to a large request
+        Given I set the terminating server data threshold to 10 bytes
+        And I start the terminating server
+        When I run the script "features/scripts/terminating_request.rb" using ruby synchronously
+        And I wait to receive an error
+        And the terminating server has received 1 requests
```
twometresteve, line 15: Should the expectations be different in this scenario to the previous one?  I can't see what this test is demonstrating.

twometresteve (approved):

PR #346, opened 2022-03-31
Fix received errors match step

Cawllec:
## Goal
Fixes an issue where the `the received errors match` step fails due to the `request_matches_row` method having been moved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 6.9.6 - 2022/03/31
+
+## Fixes
+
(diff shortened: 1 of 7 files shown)


nickdowell:
Verified the fix on my end 👍

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 6.9.6 - 2022/03/31
```
twometresteve, line 1: I think the `VERSION` and `Gemfile`/``.lock` will need updating?

Cawllec: Hmm, VERSION updated but the Gemfile didn't

PR #347, opened 2022-04-01
Decrease noise in Bugsnag by skipping assertion failures

Cawllec:
## Goal

We've implemented Bugsnag error reporting in Cucumber, but it's very noisy due to the presence of test failures preventing us from seeing unexpected errors as easily.

## Changes

1. I've changed the "payloads not received" error from a standard `RuntimeError` to a `Test::Unit::AssertionFailedError`.  This isn't strictly correct, as there are many places where a systematic error can cause this assertion to fail, but it can also be reflective of a genuine test failure.  @twometresteve I'd suggest we make a custom error for this failure, then we can determine how to handle it on a separate basis (e.g. route it to a different bugsnag api-key)
2. I've added a filter to our Bugsnag config to drop events involving the `Test::Unit::AssertionFailedError` class, which will filter assertion failures out of our reporting errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Enhancements
 
 - Add method to download device logs from BrowserStack [#344](https://github.com/bugsnag/maze-runner/pull/344)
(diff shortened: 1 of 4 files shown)


joshedney (approved):
LGTM!

PR #350, opened 2022-04-20
Retry failed app uploads for ten minutes

twometresteve:
## Goal

Retry failed app uploads for ten minutes, giving BrowserStack more of a chance to repair itself.

## Changeset

Number of iterations of retry loop increased, with sleep between and logging improved.

## Tests

Covered by updating the existing unit test.

### DIFF
--- a/lib/maze/browser_stack_utils.rb
+++ b/lib/maze/browser_stack_utils.rb
@@ -14,7 +14,6 @@ module Maze
           $logger.info "Using pre-uploaded app from #{app}"
         else
           expanded_app = Maze::Helper.expand_path(app)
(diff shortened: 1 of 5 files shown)


lib/maze/browser_stack_utils.rb
```diff
@@ -44,6 +45,10 @@ def upload_app(username, access_key, app)
             end
 
             upload_tries += 1
+            if upload_tries < allowed_tries
+              $logger.info 'Retrying upload in 60s'
+              Kernel::sleep 60
```
twometresteve, line 50: I found that without the `Kernel::`, the unit test stub would not take effect.

joshedney (approved):

PR #351, opened 2022-04-20
Ensure devices are unlocked

twometresteve:
## Goal

Ensure devices are unlocked at the start of each test run.

## Design

If a device is already unlocked then Appium simply ignores the call.

## Changeset

Expose the Appium `driver.unlock` method and call it after starting the test session.

## Tests

Covered by CI and I have tested that every version of Android and iOS that we run against is happy with the change.


joshedney (approved):
LGTM!

PR #356, opened 2022-04-29
Dynamic retries

twometresteve:
## Goal

Provide a mechanism to allow scenarios to decide for themselves if a retry should be allowed on failure.

## Design

Builds on the existing `RetryHandler` by adding `Maze.dynamic_retry` for scenarios to dynamically set when a retry is allowed.  It's reset to false at the start of each scenario to ensure that the allowed retry doesn't carry over to subsequent scenarios.

For now retries are limited to one, although if scenarios are deciding for themselves if a retry is allowed then we may well want to extend this to unlimited retries (at least in this case).

## Changeset

`retried_previously?` also made public for the purposes of testing.

## Tests

New e2e tests added.  I've also verified that the allowed retry doesn't carry over to subsequent scenarios.


test/fixtures/framework/features/steps/steps.rb
```diff
@@ -1,3 +1,14 @@
+When('I fail on the first attempt') do
+  if $first_attempt
+    $first_attempt = false
+    fail 'Failing on the first attempt'
+  end
+end
+
+When('I let the scenario retry') do
```
Cawllec, line 8: Should this be moved into the main steps?

twometresteve: It was intentional to have it just in the localised test steps, for a couple of reasons:

"I let the scenario retry" isn't needed in general - if a blanket retry is allowed then we already have `@retry`.

The way I saw the feature working was for `Maze.dynamic_retry` to be the API for test suites to work with.  Dynamic retries should only be allowed in specific circumstances, and I couldn't see a way to encapsulate that in a generic step.  For example, the reason I'm adding this feature is to allow an exception with a specific message to occur, but not any other.

Cawllec: Okay, that makes sense

Cawllec (approved):

PR #362, opened 2022-05-09
Allow invalid requests to be ignored on specific endpoints

twometresteve:
## Goal

Allow invalid requests to be ignored on specific endpoints.

## Design

By default, Maze will captured invalid (e.g. unparsable as JSON) requests that are received and fail the scenario.  This isn't always desirable, though, so a new config value `captured_invalid_requests` is added to specify the types of requests that should be monitored for invalidity.

To use, a test suite would remove the request type it wants to remove in its `BeforeAll` hook, e.g:

```
Maze.config.captured_invalid_requests -= [:sessions]
```

## Changeset

Includes a slight refactor to pass a symbol from `Server` to each `Servlet` so that the `Servlet` know what type of request it is handling.

## Tests

New test added, which fails if the `When I ignore invalid errors` step is commented out.

### DIFF
--- a/lib/maze/configuration.rb
+++ b/lib/maze/configuration.rb
@@ -9,7 +9,7 @@ module Maze
       self.receive_no_requests_wait = 30
       self.receive_requests_wait = 30
       self.enforce_bugsnag_integrity = true
-      self.captured_invalid_requests = Set[:errors, :sessions, :builds, :uploads]
+      self.captured_invalid_requests = Set[:errors, :sessions, :builds, :uploads, :sourcemaps]
     end
 
     #
--- a/lib/maze/configuration.rb
+++ b/lib/maze/configuration.rb
@@ -9,7 +9,7 @@ module Maze
       self.receive_no_requests_wait = 30
       self.receive_requests_wait = 30
       self.enforce_bugsnag_integrity = true
-      self.captured_invalid_requests = [:errors, :sessions, :builds, :uploads]
+      self.captured_invalid_requests = Set[:errors, :sessions, :builds, :uploads]
     end
 
     #
--- a/lib/maze/configuration.rb
+++ b/lib/maze/configuration.rb
@@ -9,6 +9,7 @@ module Maze
       self.receive_no_requests_wait = 30
       self.receive_requests_wait = 30
       self.enforce_bugsnag_integrity = true
+      self.captured_invalid_requests = [:errors, :sessions, :builds, :uploads]
     end
 
     #
@@ -55,6 +56,9 @@ module Maze
 
     # Enables bugsnag reporting
     attr_accessor :enable_bugsnag
+
+    # The server endpoints for which invalid requests should be captured and cause tests to fail
+    attr_accessor :captured_invalid_requests
 
     #
     # General appium configuration
--- a/lib/maze/server.rb
+++ b/lib/maze/server.rb
@@ -47,10 +47,10 @@ module Maze
 
       # Provides dynamic access to request lists by name
       #
-      # @param type [String] Request type
+      # @param type [String, Symbol] Request type
       # @return Request list for the type given
       def list_for(type)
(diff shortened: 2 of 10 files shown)


lib/maze/configuration.rb
```diff
@@ -9,6 +9,7 @@ def initialize
       self.receive_no_requests_wait = 30
       self.receive_requests_wait = 30
       self.enforce_bugsnag_integrity = true
+      self.captured_invalid_requests = [:errors, :sessions, :builds, :uploads]
```
imjoehaines, line 12: I'd make this a set since it doesn't make sense to have duplicates and `includes?` is faster than on an array (almost definitely makes 0 difference when there are so few elements though)

```suggestion
      self.captured_invalid_requests = Set[:errors, :sessions, :builds, :uploads]
```

lib/maze/server.rb
```diff
@@ -50,6 +50,7 @@ def reset_response_delay
       # @param type [String] Request type
       # @return Request list for the type given
       def list_for(type)
+        type = type.to_s if type.is_a? Symbol
         case type
```
imjoehaines, line 54: `to_s` is valid on a string (it just returns the string) so this can be simplified:

```suggestion
        case type.to_s
```

lib/maze/server.rb
```diff
@@ -161,9 +162,9 @@ def start
...
             # When adding more endpoints, be sure to update the 'I should receive no requests' step
-            server.mount '/notify', Servlets::Servlet, errors
-            server.mount '/sessions', Servlets::Servlet, sessions
-            server.mount '/builds', Servlets::Servlet, builds
+            server.mount '/notify', Servlets::Servlet, :errors
+            server.mount '/sessions', Servlets::Servlet, :sessions
+            server.mount '/builds', Servlets::Servlet, :builds
             server.mount '/uploads', Servlets::Servlet, uploads
             server.mount '/sourcemap', Servlets::Servlet, sourcemaps
             server.mount '/react-native-source-map', Servlets::Servlet, sourcemaps
```
imjoehaines, line 170: Don't these also need to use a symbol now?

```suggestion
            server.mount '/uploads', Servlets::Servlet, :uploads
            server.mount '/sourcemap', Servlets::Servlet, :sourcemaps
            server.mount '/react-native-source-map', Servlets::Servlet, :sourcemaps
```

test/fixtures/http-response/features/steps/steps.rb
```diff
@@ -0,0 +1,3 @@
+When('I ignore invalid {word}') do |type|
+  Maze.config.captured_invalid_requests -= [type.to_sym]
```
imjoehaines, line 2: It might be a little clearer to use `delete` here (and wherever this is actually going to be used)

```suggestion
  Maze.config.captured_invalid_requests.delete(type.to_sym)
```

imjoehaines (changes requested):

lib/maze/server.rb
twometresteve: Absolutely, I must have got distracted mid way through.

nickdowell:
Will this allow doing `Maze.config.captured_invalid_requests -= [:sessions]` in a standard `Before` block? (i.e. not `BeforeAll` ?)

lib/maze/configuration.rb
```diff
@@ -9,6 +9,7 @@ def initialize
       self.receive_no_requests_wait = 30
       self.receive_requests_wait = 30
       self.enforce_bugsnag_integrity = true
+      self.captured_invalid_requests = Set[:errors, :sessions, :builds, :uploads]
```
imjoehaines, line 12: Should this also include `:sourcemaps`?

twometresteve: Yes

nickdowell (approved):
Thanks!

PR #364, opened 2022-05-13
Update Appium and Selenium dependencies (enforcing w3c capabilities)

twometresteve:
## Goal

Update the Appium and Selenium dependencies to the latest available, which will hopefully lead to greater stability in the longer term.

## Design

Upgrading to Appium 12 means using Selenium 4, which enforces use of w3c capabilities.  That in turn means using Appium 1.15 or later, which would cause a problem for our iOS 10/11 tests.  They are currently using Appium 1.8 to dodge various bugs on either side that caused our tests to flake (heavily at times).  

I'm therefore parking this as a draft for the time being while we consider where we take our testing of these versions.

## Changeset

- Appium/Selenium versions bumped.  
- Capabilities updated to comply with w3c
- Deprecation warnings resolved by moving to new driver methods (`terminate_app` etc)

## Tests

So far tested with the full bugsnag-cocoa pipeline, subset of bugsnag-android tests and the (limited) Maze Runner browser tests.  We should conduct some further testing before releasing this, hence why I have now created an integration branch.

### DIFF
--- a/lib/maze/helper.rb
+++ b/lib/maze/helper.rb
@@ -101,7 +101,7 @@ module Maze
           os = os&.downcase
         end
 
-        
+
         raise('Unable to determine the current platform') if os.nil?
 
         os
--- a/lib/maze/helper.rb
+++ b/lib/maze/helper.rb
@@ -94,13 +94,14 @@ module Maze
         else
           os = case Maze.config.farm
                when :bs
-                 Maze.config.capabilities['os']
+                 Maze.config.capabilities['platformName']
                else
                  Maze.config.os
                end
           os = os&.downcase
         end
 
+        
         raise('Unable to determine the current platform') if os.nil?
 
         os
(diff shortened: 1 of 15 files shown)


lib/maze/helper.rb
```diff
@@ -89,18 +89,21 @@ def read_at_arg_file(argument)
...
+        else
+          os = case Maze.config.farm
+               when :bs
+                 Maze.config.capabilities['platformName'].downcase
+               when :sl
+                 Maze.driver.capabilities['platformName']
+               else
+                 Maze.config.os
+               end
+          os = os&.downcase
```
Cawllec, line 103: If we're going to use downcase anyway, can we remove it from the `:bs` option and merge `bs` and `sl`?

twometresteve:
Converted to draft as its base now includes out Bitbar integration, which will updating as well.

PR #366, opened 2022-05-17
Update start-driver retry logic to re-enable device lists

Cawllec:
## Goal

Updates the logic of retrying driver initialisation to:
- Retry 6 times on driver failures if a single device is provided
- Retry by iterating through the given list of devices if one is present

## Tests

I've updated the current tests, and run some functional tests using existing e2e test fixtures which have worked as expected - this should be ready for review now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 6.18.1 - 2022/05/31
+
+## Fixes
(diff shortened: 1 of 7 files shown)


twometresteve (approved):
May just need to reassess the changelog/version number.

PR #368, opened 2022-05-19
Add  option to list available device farm devices

Cawllec:
## Goal
Adds an option to list all available device keys, along with the device and OS they represent in the command line.

Currently this only works for BrowserStack, but the method can be expanded upon to add more device farms as necessary.


twometresteve (approved):
Just a changelog entry...

PR #374, opened 2022-06-24
Bitbar integration (v7 candidate base)

twometresteve:


twometresteve:
No longer applicable after merging `integration/bitbar` for v6.

PR #377, opened 2022-07-06
Browser navigation restart and retry

Cawllec:
## Goal
When an error occurs during our only selenium driver interaction, browser navigation, restart the driver and retry the navigation.

Note - this change is made with the large amount of browsers available under BrowserStack in mind, so may have unexpected consequences with BitBar web.  With that in mind, I'm going to be doing some manual testing to ensure compatibility, but would appreciate input upfront anyway.


twometresteve (approved):
Seems reasonable to me.  If you've seen the retry effectively work with BrowserStack then I'd be happy to see it get merged and released (just update the changelog).

We didn't add support for BitBar web to v6 in the end, so I think we can worry about that aspect when we come to add it in v7.

PR #380, opened 2022-07-14
Bitbar web versions

Cawllec:
## Goal

Add newer BitBar-web browsers to the yml.

Note - this doesn't include the mobile browsers, as while the documentation implies they should be available at times, they apparently aren't.  This could probably be resolved with a support requst.

### DIFF
--- a/lib/maze/browsers_bb.yml
+++ b/lib/maze/browsers_bb.yml
@@ -19,6 +19,13 @@ chrome_99:
   version: '99'
   resolution: '1920x1080'
 
+chrome_102:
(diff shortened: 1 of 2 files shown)


twometresteve:
I suspect they have now aged off some browsers (supporting only latest minus 2/5, depending on the browser).  If so then we should remove those entries now.

Cawllec:
After double checking it seems that all the browsers in the file are still supported by BitBar currently.

twometresteve (approved):
LGTM

PR #381, opened 2022-07-15
Display scenario locations in grey

twometresteve:
## Goal

Lessens the visual impact of displaying scenario locations in Buildkite.

## Design

Uses ANSI color codes to make the locations grey (the same approach as Cucumber).

## Changeset

I've also wrapped the line in a guard so that it only gets printed on Buildkite.  Locally the line has always been a duplicate to what Cucumber produces, which we accepted for the benefit of being able to fold the logs.

## Tests

See https://buildkite.com/bugsnag/maze-runner/builds/2435#018201ab-0de5-4964-80fd-63b0b51f2564.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 6.22.1 - 2022/07/15
+
+## Fixes
+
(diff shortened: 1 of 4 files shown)


joshedney (approved):
LGTM!

PR #382, opened 2022-07-19
BitBar web credentials from TMS

Cawllec:
## Goal

Allows the sue of more credentials by acquiring them from the TMS alongside BitBar devices.

Requires #380 to be merged first


twometresteve (approved):
Oh wow, that was easier than I expected!

PR #390, opened 2022-09-06
Refactor BrowserStack specific source

twometresteve:
## Goal

Work towards a more OO code base by bringing BrowserStack-specific Appium/Selenium source code together under a new namespace of `Maze::Farm::BrowserStack`.

## Design

Regrouping source code to sit together based on the farm and protocol type makes it more OO and it will be easier to support, for example, W3C and non-W3C mode per farm.  The code was quite procedural before, with code grouped by function (e.g. in `capabilities.rb`), which made sense on one level, but doesn't help when looking to model device/browser farms as a cohesive units.

## Changeset

The overall diff looks a little nasty, but it should be relatively easy to review one commit at a time.  I've been careful to refactor only a single file at a time in each commit, with only module name and file names/paths changing.  I haven't changed the inherent content of the code in any way.

## Tests

Covered by the device and browser tests we have for the BrowserStack farm in CI (and is why I chose to refactor the BrowserStack source code first to establish a new code structure - refactoring our BitBar support will require a greater test effort).


tomlongridge (approved):
Yep, refactorings make sense to me commit-by-commit

PR #392, opened 2022-09-07
Handle nil device list

twometresteve:
## Goal

Handle a `nil` device list, which happens when using `--farm=local` (because a `--device` option is not required).

## Changeset

Addition of a simple nil guard.

## Tests

Tested locally to verify that:
- A session can be successfully created
- Retries still occur if the driver cannot be started


nickdowell (approved):
👍 fixes the issue for me, thanks!

PR #393, opened 2022-09-08
Add Android 13 and iOS 16 Beta on BrowserStack

twometresteve:
## Goal

Add Android 13 and iOS 16 Beta on BrowserStack.


joshedney (approved):
LGTM!

PR #394, opened 2022-09-08
Fix support for BitBar (using W3C capabilities)

twometresteve:
## Goal

Fix support for BitBar in Maze v7, where W3C capabilities are now mandatory (having upgraded appium_lib to v12).

## Design

To some extent the release of v7 was a little hasty, as we hadn't had the chance to retest and fix the BitBar integration.  This PR brings BitBar back as an option. 

## Changeset

- Capabilities updated to be W3C compliant
- Various tweaks to improve logging
- BitBar Appium server URL change to US, as our dedicated devices are located there

## Tests

Tested locally:
- Android - successfully ran a single smoke test scenario several times.
- iOS - capabilities appear to be in order, but the server is complaining that the app (the Id given for the uploaded IPA) cannot be found.  Have raised with the BitBar team.

### DIFF
--- a/lib/maze/hooks/appium_hooks.rb
+++ b/lib/maze/hooks/appium_hooks.rb
@@ -1,25 +1,4 @@
 # Contains logic for the Cucumber hooks when in Appium mode
-
-module Selenium
-  module WebDriver
-    module Error
-      class ServerError < StandardError
-        def initialize(response)
-          if response.is_a? String
-            super(response)
-          else
-            $logger.error "Server response: #{response.inspect}"
-            super("status code #{response.code}")
-          end
-        end
-      end # ServerError
-    end # Error
-  end # WebDriver
-end # Selenium
-
-
-
-
 module Maze
   module Hooks
     # Hooks for Appium mode use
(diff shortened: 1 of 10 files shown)


lib/maze/hooks/appium_hooks.rb
```diff
@@ -121,9 +121,9 @@ def device_capabilities(config, tunnel_id = nil)
                                                               config.os,
                                                               config.os_version,
                                                               config.capabilities_option
-          capabilities['bitbar_app'] = config.app
-          capabilities['bundleId'] = config.app_bundle_id
```
tomlongridge, line 125: Deliberately removed?

tomlongridge (approved):
These changes look reasonable to me - just one thing I found where it looks like we've dropped a field, but that could be intentional. Happy to approve if that's the case.

lib/maze/hooks/appium_hooks.rb
twometresteve: Good point, actually, it may have been the cause for the issue on iOS.  I'm discussing with out Bitbar colleagues now.

twometresteve: Having discussed it, Bitbar are changing their software so that the bundle Id will not be needed on iOS (great news for us).  That should be realised within the next couple of weeks, so I'm happy to merge this as-is now.

PR #399, opened 2022-09-26
Appium/Selenium client refactor

twometresteve:
## Goal

- Consolidate code related to the same platform (BrowserStack or BitBar) into the same area/files.
- Make the Appium client code more OO, making it easier to add a legacy (non-W3C) mode of operation
- Make better use of Ruby namespaces

## Design

Apologies for the unfriendly looking diff.  However, this isn't a change that is well-suited to review-by-diff, no matter how it is presented.  I would review the new codebase as a whole, coupled with the testing argument that I've outlined below.

The main thrust of the change is to try and group Appium/Selenium code for each device farm - first by technology type (under `lib/maze/client`) and then for each device farm.

## Documentation

Updated in a couple of places for filename changes.

## Changeset

Appium/Selenium code is now centralized under `lib/maze/client` and group by farm rather than specific function.  E.g. `capabilities.rb` has been removed and its functionality absorbed into the class for that farm.

A base class `Maze::Client::Appium::BaseClient` handles the common Appium tasks, with derived classes handling behaviour for each farm (BrowserStack, BitBar and local).  A utils class still exists for each remote farm, although further refactoring could also eliminate these and for now they are a least placed in a nearby namespace). 

## Tests

- BitBar browser step added to the existing CI pipeline.
- Unit tests updated as appropriate
- Exploratory testing performs on the CLI (e.g. to ensure that `--help` and `--list-devices` continue to work)
- Further testing will also be performed naturally as part of migrating our pipelines to v7


Cawllec (approved):
Approved following discussion/comments over slack

PR #400, opened 2022-09-28
Allow use of legacy Appium/Selenium clients

twometresteve:


lib/maze/hooks/appium_hooks.rb
```diff
@@ -12,7 +12,13 @@ def before_all
...
           @client = Maze::Client::Appium::BitBarClient.new session_uuid
         when :bs
-          @client = Maze::Client::Appium::BrowserStackClient.new session_uuid
+          if ENV['USE_LEGACY_DRIVER']
+            $logger.info 'Using the JSON-WP Appium client'
+            @client = Maze::Client::Appium::BrowserStackClient.new session_uuid
+          else
+            $logger.info 'Using the W3C Appium client'
+            @client = Maze::Client::Appium::BrowserStackJsonWPClient.new session_uuid
+          end
```
Cawllec, line 21: As a naming convention would it make more sense if one client was named `BrowserStackLegacyClient` as that'll essentially be static moving forward, and the other named `BrowserStackClient`?

twometresteve: Yes, that would make sense - I think I've used the term "legacy" everywhere else.

Cawllec: Resolved offline

Cawllec (approved):

PR #402, opened 2022-10-01
Add traces endpoint

twometresteve:
## Goal

Adds a new end for receiving trace requests, on `/traces`.

## Design

Follows the same approach as existing endpoints, e.g. #213.

A future PR will introduce JSON schema based validation of requests received on the endpoint.

## Changeset

New end point and tests.

## Tests

Unit tests updated and new e2e test added to CI.

### DIFF
--- a/lib/maze/server.rb
+++ b/lib/maze/server.rb
@@ -182,7 +182,7 @@ module Maze
             server.mount '/builds', Servlets::Servlet, :builds
             server.mount '/uploads', Servlets::Servlet, :uploads
             server.mount '/sourcemap', Servlets::Servlet, :sourcemaps
-            server.mount '/trace', Servlets::Servlet, :traces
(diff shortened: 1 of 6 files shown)


lib/maze/server.rb
```diff
@@ -173,6 +182,7 @@ def start
             server.mount '/builds', Servlets::Servlet, :builds
             server.mount '/uploads', Servlets::Servlet, :uploads
             server.mount '/sourcemap', Servlets::Servlet, :sourcemaps
+            server.mount '/trace', Servlets::Servlet, :traces
```
nickdowell, line 185: Wouldn't `/traces` be more fitting? (following `/sessions` and `/logs`)

twometresteve: Changed to `/traces` - makes sense when `/sessions` and `/logs` are used more than `/sourcemap`.

nickdowell (approved):

PR #406, opened 2022-10-07
 Add error code handling for pipeline management

Cawllec:
## Goal

A method of determining specific error codes for specific reasons for maze-runner tests failing.
Currently, this will:
- Verify if an error falling into a specific group has occurred throughout the whole test run (currently only targeting appium failures)
- Check if a code has been specified by a step (for example, failing to receive payloads, or commands not being read)
- Otherwise the current exit code as determined by existing cucumber logic will be used

## Design

The code is separated into two main pieces, a plugin and an exit handler.
- The `ErrorCodePlugin` hooks into Cucumber, and stores any errors that have caused a test failure in the `ErrorCodeHook` class.
- The `ErrorCodeHook` class registers an `at_exit` handler, which will process the list of testing errors, as well as verifying if a user-submitted error code has been specified.  If one has, or if one of the errors matches one on a pre-determined list, we will output a different exit code by invoking `exit(error_code)`.


Cawllec:
Following discussion and re-evaluation, a few changes have been made:
- The relevant hooks will now only be registered if the `--fail-fast` option is enabled in cucumber.  This will allow us to use it in our most common mode - on CI, without worrying about distorting other test outcomes.
- The lookup table for error codes has been combined with the retry table for automatically retrying after certain errors occur.  This table has been moved to `lib/maze/errors.rb`

Not all of the functionality is fully implemented yet, in-progress is:
- Adding specific error codes for payloads not being received or commands not being read

lib/maze/errors.rb
```diff
@@ -16,5 +20,33 @@ def initialize(message='Element not found', element='No element specified')
...
     end
+
+    ERROR_CODES = {
+      ::Selenium::WebDriver::Error::UnknownError => {
+        retry: true,
+        error_code: 10
+      },
+      ::Selenium::WebDriver::Error::WebDriverError => {
+        retry: true,
+        error_code: 10
```
twometresteve, line 31: Did you mean to re-use 10?

twometresteve:
> > I've added some E2E tests to cover exit codes. That was an... interesting experience.
> 
> A number of new scenarios contain only a `Given` step with no expectations, are they still in progress? Also is there a reason why there are several feature files each with only a single scenario in? Could there just be one feature file with 6 or 7 scenarios in?

Oh hang on, I'm starting to see how it's set up now, please ignore that comment!

twometresteve (commented):
Looks good, could we just put a short new section in `DOCS.md` so we have something to point users at?

twometresteve (approved):

PR #407, opened 2022-10-07
Ensure String patterns behave as expected when calling Maze.check.match

twometresteve:
## Goal

Ensure String patterns behave as expected when calling Maze.check.match.

## Design

It's infuriating when you write a new Cucumber step that uses `Maze.check.match` and forget to convert the pattern into a `Rexexp` option.  It's completely unobvious to a developer when it seems it should just work - this change means you can pass either a `String` or `Rexexp` object.

## Tests

New unit test added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 7.2.2 - 2022/10/07
+
+## Fixes
+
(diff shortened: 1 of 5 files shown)


joshedney (approved):
LGTM!

Cawllec (approved):

PR #408, opened 2022-10-11
Ensure appium-lib-core is temporarily restricted to v5.4.0 in w3c mode

Cawllec:
## Goal

Appium-lib-core 5.5.0 has released, which has broken some core functionality of the notifier.  This change restricts our W3C running mode to appium-lib-core 5.4.*, which runs as expected.

Note - the appium-lib-core release does not effect legacy running.


twometresteve (approved):
Pending CI, but thank you!

PR #409, opened 2022-10-17
Add step to allow setting a status code for specific connection types

Cawllec:
## Goal

Allows a "verb" to be specified when setting an HTTP response, so we can make our tests involving `OPTIONS` requests more deterministic

## Tests

Expanded http_payload tests to cover the usage of the new steps.


twometresteve (approved):
Just a `bundle update` needed to update the `Gemfile.lock`?

PR #410, opened 2022-10-20
Add support for gzipped JSON bodies

Cawllec:
## Goal

One of the methods of delivering performance traces will be via a gzip compressed JSON payload.
This change adds support for that method into our test server by decoding the gzipped payload before JSON parsing it.  Any errors that occur at this stage will be caught by our existing error handling.


twometresteve (approved):
Looks good, just needs a CHANGELOG entry.

PR #413, opened 2022-10-21
Various fixes

twometresteve:
## Goal

Fixes various issues with the W3C protocol upgrade, having embarked on porting bugsnag-android to use Maze v7.

## Tests

Covered by CI and from local testing with bugsnag-android.

### DIFF
--- a/lib/maze/servlets/command_servlet.rb
+++ b/lib/maze/servlets/command_servlet.rb
@@ -16,13 +16,15 @@ module Maze
         response.header['Access-Control-Allow-Origin'] = '*'
 
         commands = Maze::Server.commands
+
         if commands.size_remaining == 0
-          response.body = 'No commands to provide'
-          response.status = 400
+          response.body = '{"action": "noop", "message": "No commands queued"}'
+          response.status = 200
         else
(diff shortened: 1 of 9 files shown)


lib/maze/servlets/command_servlet.rb
```diff
@@ -16,13 +16,15 @@ def do_GET(_request, response)
         response.header['Access-Control-Allow-Origin'] = '*'
 
         commands = Maze::Server.commands
+
         if commands.size_remaining == 0
-          response.body = 'No commands to provide'
-          response.status = 400
+          response.body = '{"action": "noop", "message": "No commands queued"}'
+          response.status = 200
```
twometresteve, line 22: I realise this change may be contentious and worth a discussion IRL.

Cawllec: Will this effect any currently running pipelines, or just something we should be mindful of going forward?

twometresteve: Unity pipeline now tested against this and working fine.

Cawllec (approved):

PR #415, opened 2022-11-01
Use correct capability for OS version

twometresteve:
## Goal

Correct the setting of `Maze.config.os_version` when using the w3c protocol on BrowserStack.

## Changeset

Correct the capability that we pull out for use at the OS version in config.

## Tests

Tested as part of updating bugsnag-android (OS version based `@skip`a were not working before this).


joshedney (approved):
LGTM!

PR #416, opened 2022-11-02
Initial schema validation implementation

Cawllec:
## Goal

Adds a mechanism through which trace payloads can be implicitly validated as they are received by being compared to a pre-designated schema.

The schemas, in the `json-schema` format, are stored internally and consumed as maze-runner starts, so versioning the files should be fairly easy.

Currently this behaviour is implicit and always enabled.  In the future we may wish to add an option to explicitly enable this behaviour.

The trace schema is currently not implemented, so a placeholder will be used until that's ready and working.

## Tests

End to end tests have been added, although one is currently disabled due to the actual schema not having been implemented yet. Once we've determined the exact setup of the schema we'll be using going forward these can be updated to correctly test the verification.


test/fixtures/payload-helpers/features/traces_support.feature
```diff
@@ -10,4 +10,12 @@ Feature: Testing gzip support on traces endpoint
...
\ No newline at end of file
+        And the trace payload field "array" is a non-empty array
+
+    Scenario: The trace endpoint can identify a valid request
+        When I send a "valid" trace request
+        Then I wait to receive a trace
+
+    Scenario: The trace endpoint can identify an invalid request
+        When I send a "invalid" trace request
+        Then I wait to receive a trace
```
twometresteve, line 21: Presumably you'll need do work some magic here to stop the scenario failing the test run?

twometresteve (commented):
General approach and example log output look good to me 👍

test/fixtures/payload-helpers/features/traces_support.feature
Cawllec: Yes - this is more for testing/output examples right now

twometresteve (approved):

PR #417, opened 2022-11-03
Provide more intuitive failure message for Maze.check.match

twometresteve:
## Goal

The message provided by test-unit's `assert_match` method for failures, whilst possibly technically correct, is counter-intuitive for the reader.  This changes adds a message to try and make it clearer.

## Changeset

Also took the opportunity to set the world turning again.

## Tests

Tested locally by forcing a failure in one of the e2e tests.

image


joshedney (approved):
lgtm!

PR #419, opened 2022-11-04
Add steps to set sampling probability response header

twometresteve:
## Goal

Add steps to set the `Bugsnag-Sampling-Probability` response header for traces.

## Design

Follows the same basic pattern as the existing steps we have for setting the HTTP response code.  To implement the setting of the response header, I opted for creating a new specialised servlet class for traces, refactoring the existing Servlet class a little to allow the setting of the response headers.

## Changeset

- New steps and server state for handling sampling probability.
- New specialised class for handling the /trace endpoint.
- Test framework for sending requests refactored from a script into a Ruby method, allowing the response to be captured and checked.

## Tests

Covered by CI.


Cawllec (approved):

nickdowell:
@twometresteve will this reset the `Bugsnag-Sampling-Probability` header between scenarios? I suspect not, given my confusing test failures that started recently 🤔

PR #424, opened 2022-11-23
Handle lists of HTTP status codes to return

twometresteve:
## Goal

Adds the Cucumber step `'I set the HTTP status code for the next requests to {string}` (where `{string}` is a comma separated list of values.

## Design

To implement this is was necessary to reimplement the inner working of existing mechanism for setting specific HTTP status codes that subsequently return to the default.  This is implemented in the `Generator` class, which encapsulates a `Queue` that is continually "topped up" by an underlying `Enumerator`.  The `Enumerator` allows more sophisticated logic to be implemented, but the `Queue` is needed to allow access from multiple threads.

## Changeset

- Existing Cucumber steps reordered to group in area of functionality
- New step added
- Mechanism for queuing up status codes to return reworked
- `reset!` metod added to `Maze::Server`

## Tests

Unit and e2e tests added.  I've also checked that the bugsnag-js tests that use the verb specific steps still work as expected.


lib/maze/server.rb
```diff
@@ -35,26 +28,23 @@ class << self
...
-          code = @status_code ||= 200
-          if reset_status_code
-            @status_code = 200
-            @status_override_verb = nil
-          end
-        else
-          code = 200
-        end
-        code
+      # @param [String] verb HTTP verb for which the status code is wanted
```
imjoehaines, line 38:
```suggestion
      # @param verb [String] HTTP verb for which the status code is wanted
```

lib/maze/server.rb
```diff
@@ -35,26 +28,23 @@ class << self
       # @return [String] The UUID attached to all command requests for this session
       attr_reader :command_uuid
 
+      def set_status_code_generator(generator, verb = DEFAULT_HTTP_VERB)
+        @generators ||= {}
+        @generators[verb] = generator
```
imjoehaines, line 33: It feels wrong to me to only set POST when no verb is given

I don't _think_ it could actually cause bugs, but I'd feel more comfortable if not passing a verb was equivalent to setting it for all verbs rather than just POST

lib/maze/generator.rb
```diff
@@ -0,0 +1,61 @@
...
+  # Create a thread that will constantly append to @queue with values from the
+  # given enumerator
+  #
+  # @param enumerator [Enumerator]
+  # @return [Thread]
+  def create_queue_filler(enumerator)
+    # By passing the enumerator as an argument to Thread.new, it creates a copy
+    # local to that thread and therefore we're not sharing an enumerator across
+    # threads
+    Thread.new(enumerator) do |status_code_generator|
```
imjoehaines, line 35: This thread is never explicitly killed so it might stay around forever

I'm not actually sure how Ruby handles this — the Generator instance could be garbage collected when Server doesn't have a reference to it anymore, which may kill this thread at the same time, but this thread could also potentially prevent the Generator from being garbage collected at all (especially since it has a reference to `@queue`)

imjoehaines: I think it's safer to replace the enumerator instance on the Generator, so that it can handle killing the thread internally

Replacing the Generator instance in Server was a neat idea, but if we need to kill this thread explicitly then I think you'll struggle to handle that solely in Server as you'd need to track when you drop the last reference to an existing Generator

lib/maze/server.rb
imjoehaines: Or is the assumption in existing tests using the all verbs step that it actually only applies to POST and not e.g. OPTIONS?

twometresteve: > if not passing a verb was equivalent to setting it for all verbs rather than just POST

Yes, I think you're right and the current implementation definitely doesn't work in some cases (I believe when OPTIONS get involved).  Will rework along the lines you suggest.

lib/maze/server.rb
twometresteve: Well spotted.

lib/maze/generator.rb
twometresteve: I've added a `close` method to the Generator now to handle this, which we can call before replacing it.  It means we can keep the internal implementation details of the Generator hidden away (but more importantly I avoid a headache refactoring what I already had).

lib/maze/server.rb
```diff
@@ -35,26 +28,34 @@ class << self
...
+      def set_status_code_generator(generator, verb = nil)
+        @generators ||= {}
+        if verb.nil?
+          ALLOWED_HTTP_VERBS.each do |allowed_verb|
+            @generators[allowed_verb]&.close
+            @generators[allowed_verb] = generator
+          end
         else
-          override_status = true
+          @generators[verb]&.close
```
imjoehaines, line 43: This doesn't work if you set a generator for all codes, then set another generator for a specific code — you'll close a generator that's still being used

This is what I meant by:

> I think you'll struggle to handle that solely in Server as you'd need to track when you drop the last reference to an existing Generator

lib/maze/server.rb
```diff
@@ -35,26 +28,34 @@ class << self
...
+      # generator will be shared across all allowable HTTP verbs.
       #
-      # @return [Integer] The HTTP status code, defaults to 200
-      def status_code(verb=nil)
-        if @status_override_verb
-          override_status = @status_override_verb.eql?(verb)
+      # @param verb [String] HTTP verb
+      def set_status_code_generator(generator, verb = nil)
+        @generators ||= {}
+        if verb.nil?
```
imjoehaines, line 37: This can be way more Ruby:

```ruby
Array(verb || ALLOWED_HTTP_VERBS).each do |verb|
  @generators[verb] = generator
end
```

lib/maze/server.rb
twometresteve: Discussed IRL and method updated to account for generators still in use.

lib/maze/server.rb
twometresteve: Done (I will never make it as a Ruby style guide)

lib/maze/server.rb
```diff
@@ -35,26 +28,32 @@ class << self
...
-            @status_code = 200
-            @status_override_verb = nil
-          end
+      # The intended HTTP status code on a successful request
+      #
+      # @param verb [String] HTTP verb for which the status code is wanted
+      #
+      # @return [Integer] The HTTP status code for the verb given
+      def status_code(verb)
+        if @generators[verb].nil? || @generators[verb].empty?
```
imjoehaines, line 52: As above; I don't think `empty?` could ever return true

```suggestion
        if @generators[verb].nil?
```

lib/maze/server.rb
```diff
@@ -244,6 +239,18 @@ def stop
...
+      def reset!
+        Maze::Server.errors.clear
+        Maze::Server.sessions.clear
+        Maze::Server.builds.clear
+        Maze::Server.uploads.clear
+        Maze::Server.sourcemaps.clear
+        Maze::Server.traces.clear
+        Maze::Server.logs.clear
+        Maze::Server.invalid_requests.clear
+        @generators&.values&.each { |generator| generator.close }
```
imjoehaines, line 252: Can `values` ever return `nil`? If `@generators` is only `nil` or a hash, it should be fine to always call `values` here

Maybe also worth nilling out `@generators`, so there's no chance of calling a closed one

```suggestion
        @generators&.values.each(&:close)
        @generators = nil
```

lib/maze/generator.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    # @return The next value
+    def next
+      @queue.pop
+    end
+
+    # @return Whether the internal queue is empty
+    def empty?
+      @queue.empty?
+    end
```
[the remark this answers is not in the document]
imjoehaines: Oh I see — this is for when the generator has been closed

I think in that case it'd be better to be more explicit:

```suggestion
    # @return Whether the generator has been closed
    def closed?
      @queue.closed?
    end
```

lib/maze/server.rb
imjoehaines: Assuming you rename `empty?` &rarr; `closed?`:

```suggestion
        if @generators[verb].nil? || @generators[verb].closed?
```

lib/maze/server.rb
```diff
@@ -241,15 +241,22 @@ def stop
...
+        errors.clear
+        sessions.clear
+        builds.clear
+        uploads.clear
+        sourcemaps.clear
+        traces.clear
+        logs.clear
+        invalid_requests.clear
+        @generators&.values.each(&:close)
+        @generators = nil
```
imjoehaines, line 259: Need to close the generators before overwriting them

```suggestion
        @generators&.values.each(&:close)
        @response_delay_ms = 0
        @reset_response_delay = false
        errors.clear
        sessions.clear
        builds.clear
        uploads.clear
        sourcemaps.clear
        traces.clear
        logs.clear
        invalid_requests.clear

        # Default to HTTP 200 return status after each scenario
        set_status_code_generator(Maze::Generator.new [Maze::Server::DEFAULT_STATUS_CODE].cycle)
```

imjoehaines: Oh actually, if we're overwriting them then we'll already handle closing them, so this can be:

```suggestion
        # Default to HTTP 200 return status after each scenario
        set_status_code_generator(Maze::Generator.new [Maze::Server::DEFAULT_STATUS_CODE].cycle)

        # Similarly for the response delay
        @response_delay_ms = 0
        @reset_response_delay = false
        errors.clear
        sessions.clear
        builds.clear
        uploads.clear
        sourcemaps.clear
        traces.clear
        logs.clear
        invalid_requests.clear
```

twometresteve: Oops, of course - mindless cut/paste of other reset code.

twometresteve: 2nd commit now pushed for the follow-up comment.

imjoehaines (approved):

PR #426, opened 2022-11-29
Handle lists of sampling probabilities to return

twometresteve:
## Goal

Adds the Cucumber step `I set the sampling probability for the next traces to {string}`.

## Design

Converts the existing sampling probably steps to use the `Generator` introduced in #424.

## Tests

e2e tests updated.

### DIFF
--- a/lib/maze/server.rb
+++ b/lib/maze/server.rb
@@ -254,10 +254,6 @@ module Maze
         logs.clear
         invalid_requests.clear
       end
-
-      private
-
-      attr_writer :sampling_probability_generator
     end
   end
 end
--- a/lib/maze/server.rb
+++ b/lib/maze/server.rb
@@ -10,15 +10,10 @@ module Maze
   # Receives and stores requests through a WEBrick HTTPServer
   class Server
     ALLOWED_HTTP_VERBS = %w[OPTIONS GET POST PUT DELETE HEAD TRACE PATCH CONNECT]
+    DEFAULT_SAMPLING_PROBABILITY = 1
     DEFAULT_STATUS_CODE = 200
 
     class << self
-      # Allows overwriting of the trace sampling probability
-      attr_writer :sampling_probability
-
-      # Dictates if the probability should be reset after use
-      attr_writer :reset_sampling_probability
-
       # Allows a delay in milliseconds before responding to HTTP requests to be set
       attr_writer :response_delay_ms
 
@@ -28,18 +23,27 @@ module Maze
       # @return [String] The UUID attached to all command requests for this session
       attr_reader :command_uuid
 
+      # Sets the sampling probability generator.
+      #
+      # @param generator [Maze::Generator] The new generator
+      def set_sampling_probability_generator(generator)
+        @sampling_probability_generator&.close
+        @sampling_probability_generator = generator
+      end
+
       # Sets the status code generator for the HTTP verb given.  If no verb is given then the
       # generator will be shared across all allowable HTTP verbs.
       #
+      # @param generator [Maze::Generator] The new generator
       # @param verb [String] HTTP verb
       def set_status_code_generator(generator, verb = nil)
-        @generators ||= {}
+        @status_code_generators ||= {}
         Array(verb || ALLOWED_HTTP_VERBS).each do |verb|
-          old = @generators[verb]
-          @generators[verb] = generator
+          old = @status_code_generators[verb]
+          @status_code_generators[verb] = generator
 
           # Close the old generator unless it's still being used by another verb
-          old&.close unless @generators.value?(old)
+          old&.close unless @status_code_generators.value?(old)
         end
       end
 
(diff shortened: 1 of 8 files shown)


lib/maze/server.rb
```diff
@@ -255,6 +254,10 @@ def reset!
         logs.clear
         invalid_requests.clear
       end
+
+      private
+
+      attr_writer :sampling_probability_generator
```
imjoehaines, line 260: I don't think this is used

```suggestion
```

test/fixtures/payload-helpers/features/steps/scripting_steps.rb
```diff
@@ -13,3 +13,7 @@
 When('The HTTP response header {string} equals {string}') do |header, value|
   Maze.check.equal value, $http_response[header]
 end
+
+When('The HTTP response header {string} is null') do |header|
+  Maze.check.nil $http_response[header]
+end
```
imjoehaines, line 19: Should this really be checking the header isn't there?

```suggestion
When('The HTTP response header {string} is not present') do |header|
  Maze.check.false($http_response.key(header))
end
```

test/fixtures/payload-helpers/features/traces_support.feature
```diff
@@ -30,7 +30,28 @@ Feature: Testing gzip support on traces endpoint
...
+        # Sampling probability can be set for a series of trace requests, returning to the default
+        Then I set the sampling probability for the next traces to "0.7,null,0.2"
+        And I send a "trace"-type request
+        And I wait to receive a trace
+        Then The HTTP response header "Bugsnag-Sampling-Probability" equals "0.7"
+        And I discard the oldest trace
+
+        And I send a "trace"-type request
+        And I wait to receive a trace
+        Then The HTTP response header "Bugsnag-Sampling-Probability" is null
```
imjoehaines, line 44:
```suggestion
        Then The HTTP response header "Bugsnag-Sampling-Probability" is not present
```

test/fixtures/payload-helpers/features/steps/scripting_steps.rb
twometresteve: I did consider (and would prefer) naming naming it "is not present", but there was already a precedent in `the {word} {string} header is null`.  Will change the implementation though.

test/fixtures/payload-helpers/features/traces_support.feature
twometresteve: As above.

lib/maze/server.rb
twometresteve: Without it my IDE told me that `@sampling_probability_generator` on L64 was undefined.  Perhaps there's a better way of defining it?

imjoehaines: This doesn't define `@sampling_probability_generator`, so I have no idea what your IDE is talking about or why it thinks this defines a class variable 😀

An `attr_writer` defines a setter method for the given symbol, that updates the instance variable of the same name. So this `attr_writer` expands to:

```ruby
def sampling_probability_generator=(value)
  @sampling_probability_generator = value
end
```

You only need it if you want to write `sampling_probability_generator = something` instead of `@sampling_probability_generator = something`, but you're using the instance variable everywhere (as you should!)

It's also totally valid to use instance variables without having given them a value before; they're just `nil`:

```ruby
class Abc
  class << self
    def x
      @abc
    end
    
    def y
      @abc = 1234
    end
  end
end

Abc.x # => nil
Abc.y # => 1234
Abc.x # => 1234
```

This definition doesn't do anything because it's never called and it'd be valid to use an instance variable before "defining" it anyway, so this sounds like a bug in your IDE

imjoehaines: I guess it's upset because you _could_ call `@sampling_probability_generator.next` before having set it, which would crash. I think we would want that to crash though, since that would be a bug in Maze Runner and should never happen as there's a default generator set automatically

test/fixtures/payload-helpers/features/steps/scripting_steps.rb
imjoehaines: OK — I think that step is also wrong since a HTTP header can only ever be a string so checking it's null doesn't make sense, but it's probably to late to change that

lib/maze/server.rb
twometresteve: Actually, I've just removed it and the IDE is now fine.  It must either have been confused, or I genuinely got something wrong at that point of the refactor.

imjoehaines (approved):

test/fixtures/payload-helpers/features/steps/scripting_steps.rb
imjoehaines: Not super important for this case but something that would help with that kind of thing is adding the new step in a v7 release, then removing the old one in v8

That way everything can be migrated over without having to also update to a new major release with other breaking changes

I'm not really convinced this is really worth changing TBH — there's no way the header could be null without it being missing, so it's really only a pedantic thing

twometresteve: Not a bad suggestion, although it comes with the risk that we forget to remove the unwanted one and carry arguably worse tech debt into a new major version (i.e. two steps that do the same thing).  I'm on board with ironing out pedantic issues, though - I am basically a professional pedant after all.

PR #427, opened 2022-11-30
Refactor response delay steps to use Maze::Generator

twometresteve:
## Goal

Refactor the response delay steps to use Maze::Generator, meaning that all the steps in `network_steps.rb` use a consistent approach (and code in `Maze::Server` etc is consistent as a result).

## Design

I considered adding a new step for controlling the response delay over a series of requests, but decided there was no need for this and adding it was potentially only creating a maintenance burden for ourselves.

## Tests

Covered by existing e2e tests.


Cawllec (approved):
LGTM based on this and the previous changes regarding our usage of generators.  I've restarted some failing CI steps, as I doubt they're anything to do with this change, and will look into the issue separately.

PR #428, opened 2022-11-30
Add negative breadcrumb type and message step with tests

Cawllec:
## Goal

Enables us to verify that a breadcrumb with a specific message isn't present, even if a breadcrumb with the same type is present.

Added tests to verify positive conditions, and manually verified the step catches when the breadcrumb is present.

### DIFF
--- a/lib/features/steps/breadcrumb_steps.rb
+++ b/lib/features/steps/breadcrumb_steps.rb
@@ -39,6 +39,20 @@ Then('the event does not have a {string} breadcrumb') do |type|
   raise("Breadcrumb with type: #{type} matched") if found
 end
 
+# Test whether the first event entry does not contain a breadcrumb with a specific type and message.
+# Used for confirming filtering of breadcrumbs
+#
+# @step_input type [String] The type of the breadcrumb expected to be absent
+# @step_input message [String] The message of the breadcrumb expected to be absent
+Then('the event does not have a {string} breadcrumb with message {string}') do |type, message|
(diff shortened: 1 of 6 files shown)


lib/features/steps/breadcrumb_steps.rb
```diff
@@ -39,6 +39,20 @@
...
+# Test whether the first event entry does not contain a breadcrumb with a specific type and message.
+# Used for confirming filtering of breadcrumbs
+#
+# @step_input type [String] The type of the breadcrumb expected to be absent
+# @step_input message [String] The message of the breadcrumb expected to be absent
+Then('the event does not have a {string} breadcrumb with message {string}') do |type, message|
+  value = Maze::Server.errors.current[:body]['events'].first['breadcrumbs']
+  found = false
+  value.each do |crumb|
+    found = true if crumb['type'] == type && crumb['metaData'] && crumb['metaData']['message'] == message
```
twometresteve, line 51: Can be written more succinctly by just assigning the result of the boolean expression to §found§:

```suggestion
    found = crumb['type'] == type && crumb['metaData'] && crumb['metaData']['message'] == message
```

twometresteve (approved):
Looks good, but remember a changelog entry.

lib/features/steps/breadcrumb_steps.rb
twometresteve: In fact, you can go further and just say:
```
found = value.any? { |crumb| crumb['type'] == type && crumb['metaData'] && crumb['metaData']['message'] == message }
```

PR #429, opened 2022-12-06
Use 'docker compose' not 'docker-compose'

imjoehaines:
## Goal

`docker compose` (with a space) is v2 of the docker compose CLI. Currently Maze Runner uses `docker-compose`, which is now out of date and doesn't implement some commands (e.g. `docker compose cp`)

This is supposed to be backwards compatible, so we now always run docker compose commands with `docker compose`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# TBD
+
+## Enhancements
+
+- Use `docker compose` instead of `docker-compose` [429](https://github.com/bugsnag/maze-runner/pull/429)
+
 # 7.8.0 - 2022/12/05
 
 ## Enhancements
--- a/dockerfiles/Dockerfile.ci
+++ b/dockerfiles/Dockerfile.ci
@@ -3,9 +3,24 @@ ARG RUBY_VERSION
 FROM ruby:${RUBY_VERSION}-bullseye as ci-base
 
 RUN apt-get update && DEBIAN_FRONTEND=noninteractive \
-    apt-get install -y apt-utils docker-compose wget unzip bash bundler \
+    apt-get install -y apt-utils wget unzip bash bundler \
                        # Needed for symbolication
-                       llvm-11
+                       llvm-11 \
+                       # Needed to install docker compose
+                       ca-certificates \
+                       curl \
+                       gnupg \
+                       lsb-release
+
+# https://docs.docker.com/engine/install/ubuntu/#set-up-the-repository
+RUN mkdir -p /etc/apt/keyrings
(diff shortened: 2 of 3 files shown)


twometresteve:
> However, it does require repos to be updated to install v2 of docker compose[1](https://github.com/bugsnag/maze-runner/pull/429#user-content-fn-1-2c5f41cfbd55a500a0d9556cd94c777b): https://docs.docker.com/compose/install/

Is that really true?  I'd have thought that just the container that runs `docker compose` commands (i.e. the Maze Runner one) would need it?  

If that were the case then I'd ask what the worst that would happen if we just changed it to use the v2 CLI.  At worst the repo could pin to a version of Maze Runner while we fix it properly.  I realise there is a breaking change consideration here, but when it's meant to be backward compatible, would that not be a reasonable stance?  It would avoid the extra complexity (which isn't heinous, to be fair).

imjoehaines:
> > However, it does require repos to be updated to install v2 of docker compose[1](https://github.com/bugsnag/maze-runner/pull/429#user-content-fn-1-2c5f41cfbd55a500a0d9556cd94c777b): https://docs.docker.com/compose/install/
> 
> Is that really true? I'd have thought that just the container that runs `docker compose` commands (i.e. the Maze Runner one) would need it?

Yeah I think you're right and I was just thrown off by the Maze Runner setup with the docker tests in this repo

If everywhere else is either using the MR image or running it locally, `docker compose` should work fine

> If that were the case then I'd ask what the worst that would happen if we just changed it to use the v2 CLI. At worst the repo could pin to a version of Maze Runner while we fix it properly. I realise there is a breaking change consideration here, but when it's meant to be backward compatible, would that not be a reasonable stance? It would avoid the extra complexity (which isn't heinous, to be fair).

I think it prevents a potential unexpected break on CI, but it means not having to migrate everything to use it; it's probably OK either way 🤷

twometresteve:
> I think it prevents a potential unexpected break on CI, but it means not having to migrate everything to use it; it's probably OK either way 🤷

Personally I'd rather risk a breakage that we can react quickly to, rather than add the extra complexity to the codebase.

twometresteve (approved):

imjoehaines:
> > I think it prevents a potential unexpected break on CI, but it means not having to migrate everything to use it; it's probably OK either way 🤷
> 
> Personally I'd rather risk a breakage that we can react quickly to, rather than add the extra complexity to the codebase. But equally I can deal with it.

Ok, I dropped https://github.com/bugsnag/maze-runner/commit/9a43308d18ef3aa435e3d0cd2d282365180df469

PR #431, opened 2022-12-07
Add --tunnel option

twometresteve:
## Goal

Allow users to not use the secure tunnels provided by BrowserStack/Bitbar.

## Design

Adds a `--tunnel` (and `-no-tunnel` by virtue of it being a boolean).

There are some deployment environments where the secure tunnel is not required.  Not starting it saves some time and removes something potentially breakable from the equation, reducing flakes overall.

## Tests

Default case is covered by CI and I've updated the unit tests for the command line option processing.  I've also performed some extra testing locally to ensure that the `--no-tunnel` option can be used with BrowserStack in both W3C and JWP modes of operations (which did uncover a bug - the capabilities should not be set in this case).

### DIFF
--- a/lib/maze/client/appium/bs_client.rb
+++ b/lib/maze/client/appium/bs_client.rb
@@ -8,9 +8,11 @@ module Maze
           config.app = Maze::Client::BrowserStackClientUtils.upload_app config.username,
                                                                         config.access_key,
                                                                         config.app
-          Maze::Client::BrowserStackClientUtils.start_local_tunnel config.bs_local,
-                                                                   @session_uuid,
-                                                                   config.access_key
+          if Maze.config.start_tunnel
+            Maze::Client::BrowserStackClientUtils.start_local_tunnel config.bs_local,
+                                                                     @session_uuid,
+                                                                     config.access_key
+          end
         end
 
         def device_capabilities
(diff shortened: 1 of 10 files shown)


lib/maze/client/appium/bs_client.rb
```diff
@@ -31,6 +30,11 @@ def device_capabilities
           unless device_caps['platformName'] == 'android' && device_caps['platformVersion'].to_i <= 6
             capabilities['bstack:options']['disableAnimations'] = 'true'
           end
+          if Maze.config.start_tunnel
+            capabilities['bstack:options']['local'] = 'true'
+            capabilities['bstack:options']['localIdentifier'] = @session_uuid
+          end
```
imjoehaines, line 36: Is there any possibility of breaking things because this is now after the deep merges?

I'm guessing no, since it wouldn't work without the local tunnel and needs this setup either way

twometresteve: As long as `deep_merge!` didn't do something annoying like remove the empty hash for `bstack:options` I think it should be fine.  Maze Runner's CI does cover use of the BrowserStack tunnel though (with both W3C and JWP protocols), so we would notice a breakage now and I don't think it can shift beneath us.

imjoehaines (approved):

PR #433, opened 2022-12-09
Add steps to assert an event has _x_ breadcrumbs

imjoehaines:
## Goal

Adds two steps to allow tests to be more precise when dealing with breadcrumbs:

> the event has {int} breadcrumb(s)

> the event has no breadcrumbs

The current steps only allow checking if a breadcrumb exists at all, but not e.g. if it's the only instance of that breadcrumb. Context for this is wanting to check that breadcrumbs are cleared in between running queued jobs in Laravel, but other repos that manually clear breadcrumbs may also find this useful (e.g. Python)

### DIFF
--- a/lib/features/steps/breadcrumb_steps.rb
+++ b/lib/features/steps/breadcrumb_steps.rb
@@ -1,5 +1,30 @@
 # @!group Breadcrumb steps
 
+# Tests whether the first event entry contains the specified number of breadcrumbs.
+#
+# @step_input expected [Integer] The expected number of breadcrumbs
+Then("the event has {int} breadcrumb(s)") do |expected|
+  breadcrumbs = Maze::Server.errors.current[:body]['events'].first['breadcrumbs']
+
+  Maze.check.equal(
+    expected,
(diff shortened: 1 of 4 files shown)


lib/features/steps/breadcrumb_steps.rb
```diff
@@ -1,5 +1,30 @@
...
 
+# Tests whether the first event entry contains the specified number of breadcrumbs.
+#
+# @step_input expected [Integer] The expected number of breadcrumbs
+Then("the event has {int} breadcrumb(s)") do |expected|
+  breadcrumbs = Maze::Server.errors.current[:body]['events'].first['breadcrumbs']
+
+  Maze.check.equal(
+    expected,
+    breadcrumbs.length,
```
twometresteve, line 11: If some notifiers may omit breadcrumbs entirely, then this would error in an unfriendly way.

imjoehaines: Fixed in a1abcd7

twometresteve (approved):

PR #435, opened 2022-12-09
Add missing assertion to step

imjoehaines:
## Goal

This step wasn't actually checking the name of the breadcrumb, so it would always pass if there was at least 1 breadcrumb with the right type

### DIFF
--- a/lib/features/steps/breadcrumb_steps.rb
+++ b/lib/features/steps/breadcrumb_steps.rb
@@ -30,10 +30,12 @@ end
 # @step_input type [String] The expected breadcrumb's type
 # @step_input name [String] The expected breadcrumb's name
 Then('the event has a {string} breadcrumb named {string}') do |type, name|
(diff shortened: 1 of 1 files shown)


imjoehaines:
Not sure if this is going to break some repo relying on this, but I guess that's a bug in the test anyway

twometresteve:
> Not sure if this is going to break some repo relying on this, but I guess that's a bug in the test anyway

Absolutely, it could a test - I'll give the SDK team a heads-up.

This isn't great to have - would you mind back-porting it to v6 too (by PRing into the `v6` branch) and I'll release that as well?

imjoehaines:
> This isn't great to have - would you mind back-porting it to v6 too (by PRing into the `v6` branch) and I'll release that as well?

Sure

twometresteve:
Discussed IRL, but for the paper trail - we actually just broke this last week with a bad copy/paste in a cleanup, so there's no need to fix v6.

twometresteve (approved):

PR #436, opened 2022-12-12
Change schema validation library from `json-schema` to `json-schemer`, add trace schema

Cawllec:
## Goal

Adds the correct trace schema, and enables automatic schema validation for the trace endpoint.

## Changes

As `json-schema` was failing to validate our payloads correctly it has been replaced with `json-schemer`, which now functions as expected.


Cawllec:
@lemnik would you be able to quickly review the trace-schema change, making sure that matches what you'd written

lemnik (approved):
LGTM

PR #438, opened 2022-12-12
Add --repeater option

twometresteve:
## Goal

Adds the `--repeater` option, which allows POST requests received by the mock server to be forwarded onto bugsnag.com for display in the dashboard.

## Design

Forwarding of requests is performed by a simple new class, `RequestRepeater`, called into from the `Servlet`'s `do_POST` method.  I wasn't massively happy adding addition arguments to the `server.mount` calls in the `Server`, so open to better suggestions here.

## Documentation

Online docs updated, with a new markdown page to cover the mock server in general.

## Changeset

- New `--repeater` CL option
- New config values to store repeater details
- `RequestRepeater` class
- Plumbing from `Server` and `Servlet` adding to `RequestRepeater`
- Tests and docs updated

## Tests

I've added a few unit tests, but this has mostly been tested locally.  I've successfully run the full set of bugsnag-android e2e tests as well as the e2e tests for iOS/Android performance, with data nicely displayed in the dashboard.


lib/maze/option/parser.rb
```diff
@@ -32,10 +32,15 @@ def parse(args)
...
             opt Option::ENABLE_BUGSNAG,
-                'Enables reporting to Bugsnag on scenario failure (Require MAZE_BUGSNAG_API_KEY)',
+                'Enables reporting to Bugsnag on scenario failure (requires MAZE_BUGSNAG_API_KEY)',
                 type: :boolean,
                 default: true
 
+            opt Option::REPEATER,
+                'Enables forwarding of all received POST requests to Bugsnag using the API key held in MAZE_REPEATER_API_KEY',
+                type: :boolean,
+                default: false
```
imjoehaines, line 42: Have you considered making the API Key the option and enabling/disabling it based on the presence of the key?

i.e. instead of:

```
$ MAZE_REPEATER_API_KEY=abc bundle exec maze-runner --repeater
```

do:

```
$ bundle exec maze-runner --repeater-api-key abc
```

There might be a reason to use an environment variable, but I think it's a little nicer not to have to specify two things. Though maybe it's also useful to always provide an API key, but allow controlling whether to actually use it separately?

DOCS.md
```diff
@@ -10,28 +10,30 @@ to draft semantic tests.
 
 [Getting Started](./docs/Getting_Started.md)
 
+## Mock server
+
+In all modes of operation, Maze Runner automatically starts an HTTP server that mocks Bugsnag's server endpoint.
```
imjoehaines, line 15:
```suggestion
In all modes of operation, Maze Runner automatically starts an HTTP server that mocks Bugsnag's server endpoints.
```

DOCS.md
```diff
@@ -10,28 +10,30 @@ to draft semantic tests.
...
-
-## Exit codes
-
-In normal usage maze runner will follow the same exit code patterns that cucumber would in normal operation.
-
-When the `--fail-fast` mode is specified, maze-runner will potentially override this default behaviour with a set exit code.
-A list of codes that will automatically be set for errors can be found in [the Error Code Hook class](./lib/maze/hooks/error_code_hook.rb).
-
-## Troubleshooting
+  Bitbar or BrowserStack.n## Troubleshooting
```
imjoehaines, line 29:
```suggestion
  Bitbar or BrowserStack.
  
## Troubleshooting
```

DOCS.md
```diff
@@ -10,28 +10,30 @@ to draft semantic tests.
 
 [Getting Started](./docs/Getting_Started.md)
 
+## Mock server
+
+In all modes of operation, Maze Runner automatically starts an HTTP server that mocks Bugsnag's server endpoint.
+This mock server provides a URL for each endpoint that Bugsnag does, allowing it to receive errors, session, builds,
+source map uploads and traces.
+
+See [here](./docs/Mock_Server.md) for more details of the options available for controlling the behaviour of the mock
```
imjoehaines, line 19:
```suggestion
[See `./docs/Mock_Server.md`](./docs/Mock_Server.md) for more details of the options available for controlling the behaviour of the mock 
```

DOCS.md
```diff
@@ -10,28 +10,30 @@ to draft semantic tests.
...
-When the `--fail-fast` mode is specified, maze-runner will potentially override this default behaviour with a set exit code.
-A list of codes that will automatically be set for errors can be found in [the Error Code Hook class](./lib/maze/hooks/error_code_hook.rb).
-
-## Troubleshooting
+  Bitbar or BrowserStack.n## Troubleshooting
 
 ### Logging
 
-Maze-runner contains a Ruby logger connected to `STDOUT` that will attempt to log several events that occur during the  testing life-cycle.  By default, the logger is set to report `INFO` level events or higher, but will log `DEBUG` level events if the `VERBOSE` or `DEBUG` flags are set.  If the `QUIET` flag is set it will instead log at the `ERROR` level and above.
+Maze Runner contains a Ruby logger connected to `STDOUT` that will attempt to log several events that occur during the
```
imjoehaines, line 33: What's with the hard wrapping in markdown files?

twometresteve: Hmm, I think there are arguments both ways:
https://stackoverflow.com/questions/41513565/markdown-and-line-wrapping
(just as an example).

For me consistency is the bigger driver and there were already some hard-wrapped lines in other files - so I'd found myself just falling into that habit without thinking too hard about it.

If people felt strongly enough about it we could go to soft-wrap?  How much does it irk you?  What's the contect for you - annoying presentation in an editor that soft-wraps by default?

twometresteve:
> Will take a proper look tomorrow
> 
> One way to avoid the extra parameter would be to give the `RequestRepeater` knowledge of the URLs, based on the `request_type`
> 
> You could also avoid modifying `Servlet` by using `Module#prepend` to allow `RequestRepeater` to decorate `do_POST`
> 
> Something like this:
> 
> ```ruby
> module Maze
>   module RequestRepeater
>     def do_POST(request, response)
>       repeat(request) if enabled?
> 
>       super(request, response)
>     end
> 
>     private
> 
>     def enabled?
>       # enabled if the config option is on and this request type should be repeated
>       Maze.config.repeater && url_for_request_type
>     end
> 
>     # @param request [HTTPRequest] The request to be repeated
>     def repeat(request)
>       # TODO: Forwarding of internal errors to be considered later
>       return if request.header.keys.any? { |key| key.downcase == 'bugsnag-internal-error' }
> 
>       url = url_for_request_type
> 
>       http = Net::HTTP.new(url.host)
>       bugsnag_request = Net::HTTP::Post.new(url.path)
>       bugsnag_request['Content-Type'] = 'application/json'
> 
>       # Set all Bugsnag headers that are present
>       bugsnag_request.body = request.body
>       request.header.each {|key,value| bugsnag_request[key] = value if key.downcase.start_with? 'bugsnag-' }
>       bugsnag_request['bugsnag-api-key'] = Maze.config.repeater_api_key
> 
>       # TODO Also overwrite apiKey in the payload, if present, and recalculate the integrity header
> 
>       http.request(bugsnag_request)
>     end
> 
>     def url_for_request_type
>       case @request_type
>       when :errors then 'https://notify.bugsnag.com'
>       when :sessions then 'https://sessions.bugsnag.com'
>       when :traces then 'https://otlp.bugsnag.com/v1/traces'
>       end
>     end
>   end
> end
> 
> module Maze
>   module Servlets
>     class Servlet < BaseServlet
>       prepend RequestRepeater
> 
>       # otherwise unchanged from 'main'
>     end
>   end
> end
> ```
> 
> This is a good read for what `prepend` does: https://dev.to/abbiecoghlan/ruby-modules-include-vs-extend-vs-prepend-4gmc
> 
> Not necessarily saying this is better, but it avoids most changes to `Server` or `Servlet`

That's actually how I implemented it at first, but unfortunately it will break down for the `:uploads` request type, which corresponds to multiple URL paths under upload.bugsnag.com, depending on the technology in use.

I guess I could have ignored that for the time being, but I opted instead for having a holistic review of the setup in the context of the new Bugsnag CLI that's in development at the moment - which is likely to impact this area in itself.

imjoehaines:
> That's actually how I implemented it at first, but unfortunately it will break down for the :uploads request type, which corresponds to multiple URL paths under upload.bugsnag.com, depending on the technology in use.

I think that's still fine — you can [read the path from the request with `request#path`](https://rubyapi.org/3.1/o/net/httpgenericrequest#path) so it's only a small change to support that

twometresteve:
> > That's actually how I implemented it at first, but unfortunately it will break down for the :uploads request type, which corresponds to multiple URL paths under upload.bugsnag.com, depending on the technology in use.
> 
> I think that's still fine — you can [read the path from the request with `request#path`](https://rubyapi.org/3.1/o/net/httpgenericrequest#path) so it's only a small change to support that

Ah, interesting - I had not considered that.  I suspect we'll still have to change things for the Bugsnag CLI, but I may as well refactor the implementation so far to make the code cleaner and worry about potential complications when they actually come up later.

lib/maze/request_repeater.rb
```diff
@@ -0,0 +1,41 @@
...
+      bugsnag_request['bugsnag-api-key'] = Maze.config.repeater_api_key
+
+      # TODO Also overwrite apiKey in the payload, if present, and recalculate the integrity header
+
+      http.request(bugsnag_request)
+    end
+
+    private
+
+    def url_for_request_type
```
twometresteve, line 33: Logic now implemented here based on request_type, rather than passing it down.  I ran into trouble trying the `prepend` approach, as `do_POST` exists at two level and both need be be run.  I'm sure it's possible, but this felt the most easily achievable and better than it was before.

lib/maze/servlets/servlet.rb
```diff
@@ -19,6 +19,8 @@ def initialize(server, request_type, schema=nil)
         @request_type = request_type
         @requests = Server.list_for request_type
         @schema = JSONSchemer.schema(schema) unless schema.nil?
+
+        @repeater = Maze::RequestRepeater.new(@request_type) if [:errors, :sessions, :traces].include? @request_type
```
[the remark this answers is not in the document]
twometresteve: Moved to REPEATED_REQUEST_TYPES

lib/maze/servlets/servlet.rb
```diff
@@ -36,6 +38,8 @@ def do_GET(request, _response)
       # @param request [HTTPRequest] The incoming GET request
       # @param response [HTTPResponse] The response to return
       def do_POST(request, response)
+        @repeater.repeat(request) if Maze.config.repeater_api_key
```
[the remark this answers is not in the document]
twometresteve: Yes, you're right - great catch.  I've refactored the two guards so they appear together now, and nil-guarded the use of `@repeater`.

lib/maze/option/parser.rb
```diff
@@ -32,10 +32,14 @@ def parse(args)
                 default: true
 
             opt Option::ENABLE_BUGSNAG,
-                'Enables reporting to Bugsnag on scenario failure (Require MAZE_BUGSNAG_API_KEY)',
+                'Enables reporting to Bugsnag on scenario failure (requires MAZE_BUGSNAG_API_KEY)',
                 type: :boolean,
                 default: true
 
+            opt Option::REPEATER_API_KEY,
+                'Eables forwarding of all received POST requests to Bugsnag, using the API key provided',
```
imjoehaines, line 40:
```suggestion
                'Enables forwarding of all received POST requests to Bugsnag, using the API key provided',
```

lib/maze/request_repeater.rb
imjoehaines: Calling `super` in a prepended module will call the class' method:

```ruby
module A
  def x
    puts "A#x"
    super
  end
end

class B
  prepend A
  
  def x
    puts "B#x"
  end
end

B.new.x

# => A#x
# => B#x
```

That should be all you need to do to run both methods

lib/maze/request_repeater.rb
```diff
@@ -0,0 +1,41 @@
+module Maze
+  # Repeats POST requests
+  class RequestRepeater
+
+    # @param request_type [String] The type of request being handled by this repeater
+    def initialize(request_type)
+      @request_type = request_type
```
imjoehaines, line 7:
```suggestion
      @url = URI.parse(url_for_request_type(request_type))
```

lib/maze/request_repeater.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    # @param request [HTTPRequest] The request to be repeated
+    def repeat(request)
+
+      # TODO: Forwarding of internal errors to be considered later
+      return if request.header.keys.any? { |key| key.downcase == 'bugsnag-internal-error' }
+
+      url = URI.parse(url_for_request_type)
+      http = Net::HTTP.new(url.host)
+      bugsnag_request = Net::HTTP::Post.new(url.path)
```
imjoehaines, line 18:
```suggestion
      http = Net::HTTP.new(@url.host)
      bugsnag_request = Net::HTTP::Post.new(@url.path)
```

lib/maze/request_repeater.rb
```diff
@@ -0,0 +1,41 @@
...
+
+      # TODO Also overwrite apiKey in the payload, if present, and recalculate the integrity header
+
+      http.request(bugsnag_request)
+    end
+
+    private
+
+    def url_for_request_type
+      case @request_type
```
imjoehaines, line 34:
```suggestion
    def url_for_request_type(request_type)
      case request_type
```

lib/maze/servlets/servlet.rb
```diff
@@ -19,6 +22,10 @@ def initialize(server, request_type, schema=nil)
         @request_type = request_type
         @requests = Server.list_for request_type
         @schema = JSONSchemer.schema(schema) unless schema.nil?
+
+        if Maze.config.repeater_api_key && REPEATED_REQUEST_TYPES.include?(@request_type)
+          @repeater = Maze::RequestRepeater.new(@request_type)
+        end
```
imjoehaines, line 28: I don't like that this is nilable and the logic for the request type is duplicated (here & `url_for_request_type)

I think it'd be better for `RequestRepeater` to handle whether it is enabled so the `Servlet` can just call it unconditionally

lib/maze/request_repeater.rb
twometresteve: Yep, I don't know what I was up to there.

twometresteve:
I've refactored the code now to keep all repeater logic in the `RequestRepeater`, using `prepend` - so the Servlet is no longer cluttered with it.  _Hopefully_ that resolves all the comments above.

lib/maze/request_repeater.rb
```diff
@@ -0,0 +1,48 @@
...
+      http.request(bugsnag_request)
+    end
+
+    def url_for_request_type
+      url = case @request_type
+            when :errors then 'https://notify.bugsnag.com/'
+            when :sessions then 'https://sessions.bugsnag.com/'
+            when :traces then 'https://otlp.bugsnag.com/v1/traces'
+            end
+      URI.parse(url) if url
```
imjoehaines, line 45: Not sure if this is actually nicer so feel free to ignore, but you can return straight from this case so can skip the inline `if`:

```suggestion
            else
              return nil
            end

      URI.parse(url)
```

lib/maze/request_repeater.rb
```diff
@@ -0,0 +1,48 @@
...
+      http = Net::HTTP.new(url.host)
+      bugsnag_request = Net::HTTP::Post.new(url.path)
+      bugsnag_request['Content-Type'] = 'application/json'
+
+      # Set all Bugsnag headers that are present
+      bugsnag_request.body = request.body
+      request.header.each {|key,value| bugsnag_request[key] = value if key.downcase.start_with? 'bugsnag-' }
+      bugsnag_request['bugsnag-api-key'] = Maze.config.repeater_api_key
+
+      # TODO Also overwrite apiKey in the payload, if present, and recalculate the integrity header
```
imjoehaines, line 34: How much of an issue is this / do we know what the pipeline does if the API keys are different?

I wonder if we should make this an error case so that something unexpected doesn't happen, though I guess the worst-case is that this uses the API key in the payload and gets dropped since it doesn't match a project

imjoehaines (approved):

lib/maze/request_repeater.rb
twometresteve: I spoke to the pipeline team about it and they said the one in the body is legacy and not inspected.  So the change here would be just be for consistency's sake.  We wouldn't want to make it and error case as some notifiers still set both the header and value in the body (consistently, but will be a dummy key in test scenarios).

lib/maze/request_repeater.rb
twometresteve: Yeah, I do prefer that actually.

PR #444, opened 2022-12-22
Refactor Selenium driver code into class hierarchy

twometresteve:
## Goal

Make the browser/Selenium client code more OO and consistent with the Appium client code already present in `Maze::Client::Appium`.  

## Design

This was largely driven by an issue on BitBar that was caused by capabilities we set for BrowserStack.  Making the code more OO makes it much easier to see what is common and what is specific to either BB or BS.

I started by adding the new classes in `Maze::Client::Selenium, with the same methods used on the Appium side.  I then cut/pasted the relevant arm of each `case` into place.  `appium_hooks.rb` and `browser_hooks.rb` should now follow the same pattern.

## Tests

BB/BS use covered by CI and I've also tested local browser use manually (fixing several issues I introduced along the way!).


lib/maze/hooks/browser_hooks.rb
```diff
@@ -4,105 +4,24 @@ module Hooks
...
         when :local
-          Maze.driver = Maze::Driver::Browser.new Maze.config.browser.to_sym
+          @client = Maze::Client::Selenium::LocalClient.new session_uuid
         end
 
-        # Write links to device farm sessions, where applicable
-        write_session_link
+        @client.prepare_session
+        @client.start_session
+        @client.log_session_info
```
imjoehaines, line 20: If this is the only place using the clients, does it make sense to have three separate methods? Could the others be put into `start_session` so this can be simpler:

```suggestion
        @client.start_session
```

Then the client interface can be simplified to match:

```ruby
class Client
  def start_session -> void
  def end_session -> void
end
```

imjoehaines (commented):
I think this could be simpler by providing a single public method that starts the correct client, e.g.:

```ruby
module Maze
  module Client
    module Selenium
      def self.start(session_uuid)
        client_class =
          case Maze.config.farm
          when :bb then BitBarClient
          when :bs then BrowserStackClient
          when :local then LocalClient
          end

        client_class.new(session_uuid).tap(&:start_session)
      end
    end
  end
end
```

Then the hook's `before_all` can be a one-liner:

```ruby
def before_all
  @client = Maze::Client::Selenium.start(SecureRandom.uuid)
end
```

lib/maze/hooks/browser_hooks.rb
twometresteve: Yes, you're right.  I following the implementation for the Appium too closely (which was only done to that granularity because the code bounces around between command and farm-specifics).

twometresteve:
> I think this could be simpler by providing a single public method that starts the correct client, e.g.:
> 
> ```ruby
> module Maze
>   module Client
>     module Selenium
>       def self.start(session_uuid)
>         client_class =
>           case Maze.config.farm
>           when :bb then BitBarClient
>           when :bs then BrowserStackClient
>           when :local then LocalClient
>           end
> 
>         client_class.new(session_uuid).tap(&:start_session)
>       end
>     end
>   end
> end
> ```
> 
> Then the hook's `before_all` can be a one-liner:
> 
> ```ruby
> def before_all
>   @client = Maze::Client::Selenium.start(SecureRandom.uuid)
> end
> ```

Yes, I like that.  I've update the existing Appium side to more-or-less match that in a separate PR.

lib/maze/client/selenium/local_client.rb
```diff
@@ -0,0 +1,24 @@
...
+        end
+
+        def start_session
+          Maze.driver = Maze::Driver::Browser.new Maze.config.browser.to_sym
+          Maze.driver.start_driver
+        end
+
+        def log_session_info
+          # Nothing to do
+        end
```
imjoehaines, line 16:
```suggestion
```

lib/maze/client/selenium/local_client.rb
```diff
@@ -0,0 +1,24 @@
+module Maze
+  module Client
+    module Selenium
+      class LocalClient < BaseClient
+        def prepare_session
+          # Nothing to do
+        end
+
```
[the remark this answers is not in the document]
twometresteve: Ah yes - removing on my next branch.

PR #445, opened 2022-12-23
Support `docker compose cp` commands

imjoehaines:
## Goal

In a few repos we need to read the contents of a file on a running container, e.g.:

- in Laravel to read the installed version: https://github.com/bugsnag/bugsnag-laravel/blob/1c0fd6783c2d5763c483cea271d88f5ed1733dfa/features/lib/laravel.rb#L86-L101
- in the RN CLI to verify changes to files: https://github.com/bugsnag/bugsnag-js/blob/0599faeb4019e571bfd7111c2663b5ddcbb78901/test/react-native-cli/features/steps/steps.rb#L73-L88

This can be accomplished with `docker compose cp` and a temporary file, but can also be useful more generically e.g. copying log files out of a container to upload as CI artefacts

This PR adds two new methods to `Maze::Docker`:

- `copy_from_container(service, from:, to:)` for copying a file from a running container
- `copy_to_container(service, from:, to:)` for copying a file into a running container

I haven't defined any steps for them as I don't think we'd necessarily need them — the existing uses of this kind of thing don't actually need the files, but just their contents. For this reason I think this will only be used in custom steps per-repo for specific purposes


twometresteve (approved):
LGTM.  I've pushed a commit to add a changelog entry, so explicitly adding @joshedney to the PR to check my changes.

joshedney (approved):
LGTM!

PR #447, opened 2023-01-05
Decode all gzipped POST requests

lemnik:
## Goal
Transparently decode gzipped requests

## Design
Introduce a MazeRunner specific `HttpRequest` class that is a `Delegator` around the existing WEBrick request objects. The `HttpRequest.body` property will transparently decompress the content if the `Content-Encoding` header is `gzip`, allowing all other parts of MazeRunner to continue as normal.

## Tests
Manually tested with `bugsnag-android-performance`, otherwise relying on existing tests.


lib/maze/servlets/servlet.rb
```diff
@@ -3,8 +3,27 @@
...
+    def decode_body
+      delegate = __getobj__
+      if %r{^gzip$}.match(delegate['Content-Encoding'])
+        gz_element = Zlib::GzipReader.new(StringIO.new(delegate.body))
+        gz_element.read
+      else
+        delegate.body
+      end
+    end
+  end
```
twometresteve, line 25: I'd prefer this was in its own file, `http_request.rb` - it'll make its existence more obvious and I think there's a good chance that we could wrap more functionality into it.

twometresteve: If you're happy to move it, we'll also need a new `require_relative` line in `bin/maze-runner`.

twometresteve (approved):

PR #449, opened 2023-01-06
Fix legacy browser mode

twometresteve:
## Goal

Fixes an issue in the browser client code introduced in a recent refactor.

## Design

Before making any code changes I updated the pipeline to be sure of reproducing the error and to avoid future regression.

## Tests

Covered by new and existing CI tests.

### DIFF
--- a/lib/maze/client/selenium/bs_client.rb
+++ b/lib/maze/client/selenium/bs_client.rb
@@ -13,7 +13,7 @@ module Maze
           browser = browsers[config.browser]
 
           if config.legacy_driver?
-            capabilities = Selenium::WebDriver::Remote::Capabilities.new
+            capabilities = ::Selenium::WebDriver::Remote::Capabilities.new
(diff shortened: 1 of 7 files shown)


lib/maze/client/selenium/bs_client.rb
```diff
@@ -11,7 +13,7 @@ def start_session
           browser = browsers[config.browser]
 
           if config.legacy_driver?
-            capabilities = Selenium::WebDriver::Remote::Capabilities.new
+            capabilities = ::Selenium::WebDriver::Remote::Capabilities.new
```
twometresteve, line 16: This was the main issue.  Because the line is in a module called `Something::Something::Selenium`, by default Ruby thinks I want want a namespace relative to the current one.  The `::` at the start makes it an absolute one.

gingerbenw (approved):

PR #450, opened 2023-01-09
Only set build and project capabilities for BrowserStack

twometresteve:
## Goal

Avoid problems setting information in the BitBar dashboard caused by setting the `project` and `build` capabilities.

## Design

BitBar have their own capability, `bitbar_project`,for setting the project name in the dashboard, but it doesn't take effect if these capabilities are set.  This PR pushes the logic for setting the capabilities into the BrowserStack client rather than apply them for all device farms.

## Tests

The main thing this change risked breaking was the links to the BrowserStack dashboard that we log at the start of each session.  I've run both browser and device sessions locally, in both legacy and w3c modes to check those.  I've also checked the links created on Buildkite.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# TBD
+
+## Fixes
+
+- Only set `project` and `build` capabilities for BrowserStack [450](https://github.com/bugsnag/maze-runner/pull/450)
+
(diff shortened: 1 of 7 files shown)


lemnik (approved):
LGTM

PR #451, opened 2023-01-09
Generate and use a single UUID for for the application run

twometresteve:
## Goal

Previously we were generating UUIDs in different part of the code for the Appium/Selenium session and for returning in `/command` messages.

This change unifies all of those under a single "run UUID".

## Design

Storing the UUID in the globally accessible `Maze.run_uuid` means that we can do away much of the previous plumbing.  It also means that the UUID returned in `/command`s will match that used for the BrowserStack session, which could be useful for analysis purposes.

## Changeset

- `Maze.run_uuid` added
- Generation of UUIDs collapsed to just one instance
- Various plumbing removed

## Tests

Covered by CI and I've performed some basic checks that the BrowserStack dashboard link are still correct.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # TBD
 
+## Enhancements
+
+- Generate and use a single "run UUID" throughout the app [451](https://github.com/bugsnag/maze-runner/pull/451)
+
 ## Fixes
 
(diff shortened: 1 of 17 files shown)


lemnik (approved):
LGTM

PR #452, opened 2023-01-11
Public IP address awareness

twometresteve:
## Goal

Makes Maze Runner aware of the public IP address when running on Buildkite with the Elastic CI Stack for AWS.

## Design

We now have public IP addresses available on our Buildkite instances.  This change provides and option to make Maze Runner aware of that, with the address for now being recorded in `Maze.public_address`.

This is the foundation to re-architecting the model for using BitBar, eliminating the need for the secure tunnel and TMS.  Subsequent PRs will add functionality that takes the address recorded and writes it to a file, before pushing it to the test fixture app for use.

## Documentation

Basic docs have been added for the command line option.  Full docs will be added in a later PR, once the fully-fledged feature is in place for use with BitBar.

## Changeset

- New command line option and processing
- New config option
- `Maze.public_address` added

## Tests

Tested by adding to the BitBar browser step in CI and inspecting the logged output.


lib/maze/aws_public_ip.rb
```diff
@@ -0,0 +1,51 @@
...
+
+      ip = determine_public_ip
+      port = determine_public_port
+
+      @address = "#{ip}:#{port}"
+    end
+
+    # Determines the public IP address of the running AWS instance
+    def determine_public_ip
+      `curl --silent -XGET http://169.254.169.254/latest/meta-data/public-ipv4`
```
Cawllec, line 18: Is this a fixed address, and should it be parameterised and made configurable if not?

Cawllec (approved):
Makes sense to me - left a small comment on the address left in the code.  If this is definitely the route we're taking then I think this is fine, but I'm wondering if we can start to make plugins of some sort, rather than growing the codebase continuously.
Probably something to discuss separately at some point.

lib/maze/aws_public_ip.rb
twometresteve: It is fixed, but it's a good point - I should have included a comment as to what it is.  I
'll do that (see https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html in the meantime).

twometresteve:
> 

The idea of using Cucumber "worlds" to split functionality up has been discussed.  To me code structure is more important than the amount of code in one place, but there is certainly room for continuing to improve that - and a plugin/modular architecture could well be a good thing to put in place to help encourage/enforce good structure.

PR #454, opened 2023-01-13
Add trace/span steps from Android/Cocoa repos

twometresteve:
## Goal

Consolidates a number of Cucumber steps and methods that we added to the Android and Cocoa performance repos.

## Design

This is mostly a cut and paste job, but the steps had diverged slightly in the two repos.  The resulting effect on each repo can be seen in these PRs, which will be updated and presented as soon as Maze Runner has been released to include this PR.
https://github.com/bugsnag/bugsnag-android-performance/pull/50
https://github.com/bugsnag/bugsnag-cocoa-performance/pull/38

## Tests

If creating these steps from scratch, we would look to add a new test fixture to the Maze Runner pipeline.  However, given when we are, it was far more effective to test it by updating the performance repos at the same time - see builds for the above PRs.


lemnik (approved):
LGTM

PR #456, opened 2023-01-17
Mobile test fixture config files

twometresteve:
## Goal

Write and push a config file for mobile test fixtures to read the address of Maze Runner from. 

## Design

Mobile test fixtures currently contain a hard-coded address for reaching Maze Runner.  This needs to change in order to be device farm agnostic, as well as allow the use of public IP addresses, which must be determined dynamically.

This changes introduces the notion of a test fixture config file, which is pushed onto the device using Appium.  The file contains the appropriate Maze Runner address for the device farm and whether public IPs are being used.  Its presence on the device will of course not break existing tests and fixtures can be gradually migrated over to using it.

I'm also testing the water on having `Maze::Api` and `Maze::Internal` modules.  The intention would be to gradually move classes into these in order to form a clear layer that clients can/should be using and steer them aware from accessing internals.  This should make future refactors easier as many changes will not be considered a breaking change.

## Documentation

Short section added to existing docs for Appium mode.

## Changeset

New classes:
- `Maze::Api::Appium::FileManager`
- `Maze::Internal::FixtureConfig`

Extra plumbing to write the file and push to the device.

## Tests

Tested locally against both BitBar and BrowserStack on both Android and iOS (smoking out a problem with the initial Android path chosen in the process).


lib/maze/client/appium/bb_client.rb
```diff
@@ -18,6 +18,14 @@ def prepare_session
           end
         end
 
+        def maze_address
+          if Maze.config.aws_public_ip
+            Maze.public_address
+          else
+            "local:#{Maze.config.port}"
+          end
```
imjoehaines, line 26: IIUC there will only be a `Maze.public_address` if there's also a `Maze.config.aws_public_ip`, so this is equivalent:

```suggestion
          Maze.public_address || "local:#{Maze.config.port}"
```

It was a little confusing to me why we were checking `Maze.config.aws_public_ip` only to read `Maze.public_address`

lib/maze/internal/fixture_config.rb
```diff
@@ -0,0 +1,27 @@
...
+      def initialize
+        @config = {}
+      end
+
+      def add(key, value)
+        @config[key] = value
+      end
+
+      def write
+        FileUtils.makedirs(WORKING_DIRECTORY)
```
imjoehaines, line 20: If we're just making a file to push to the device and don't actually care about it, I think this should make a temporary file instead of cluttering up the working directory (could always log it with DEBUG if you want it visible for debugging)

Maybe `FileManager` could then take responsibility for creating the file, so you can just ask it to push a string to a device?

lib/maze/client/appium/bb_client.rb
twometresteve: Yes, I think that's a fair change to make.  FWIW I think it's still reasonable for both `Maze.config.aws_public_ip` and `Maze.public_address`, as one if set by the user provided options and the other is dynamically determined.

lib/maze/internal/fixture_config.rb
twometresteve: In fact it's not necessary to write to a file at all - the Appium API simply accepts a string.  I've now refactored to take that into account, removing a bunch of code.

.gitignore
```diff
@@ -14,4 +14,3 @@ doc/
 **/src/main/gen/
 .DS_Store
 bb.ready
-sb.ready
```
twometresteve, line 17: We don't write this file anymore (it's changed to bb.ready)

lib/maze/internal/fixture_config.rb
```diff
@@ -0,0 +1,22 @@
+require 'json'
+
+module Maze
+  module Internal
+    # Handles the Maze Runner configuration file for test fixtures
+    class FixtureConfig
```
imjoehaines, line 6: Do we envisage adding many more things here?

I'm wondering if this should just be a hash in `base_client`, or if it should be more aware of what's being put in it, e.g. rather than a generic `add` method, should there be a specific method for `maze_address`?

twometresteve: It's a good question.  Initially I liked having it as its own specific abstraction, but it's over-engineered if we never need to add anything else - and in general we have commands to convey information to the fixture.  For now I'll just write a JSON string inline, saving the need for it completely.  We can factor it out again later if really necessary.

yousif-bugsnag:
Currently the config file only seems to be there on the first scenario of a run - it looks like the config file is being deleted when the app is reset between scenarios, but isn't being subsequently re-written.

lib/maze/api/appium/file_manager.rb
```diff
@@ -0,0 +1,29 @@
...
+        # Creates a file with the given contents on the device (using Appium).  The file will be located in the app's
+        # Documents directory for iOS and /sdcard/Android/data/<app-id>/ for Android.
+        # @param contents [String] Content of the file to be written
+        # @param filename [String] Name (with no path) of the file to be written on the device
+        def write_app_file(contents, filename)
+          path = case Maze::Helper.get_current_platform
+                 when 'ios'
+                   "@#{@driver.app_id}/Documents/#{filename}"
+                 when 'android'
+                   "/sdcard/Android/data/#{@driver.app_id}/#{filename}"
```
yousif-bugsnag, line 20: I'm not sure how robust this will be if we're running tests on a wide range of devices - I think the path to external storage can differ between manufacturers

lib/maze/client/appium/base_client.rb
```diff
@@ -18,13 +18,23 @@ def start_session
           # Ensure the device is unlocked
           Maze.driver.unlock
 
+          # Write Maze's address to file and push to the device
+          config = Maze::Internal::FixtureConfig.new
+          config.add('maze_address', maze_address)
+          Maze::Api::Appium::FileManager.new.write_app_file(config.to_s,
+                                                            Maze::Internal::FixtureConfig::DEVICE_FILENAME)
+
```
twometresteve, line 26: Thanks to @yousif-bugsnag for realising that this needs to be executed before every scenario, not just once at the start.  Some new plumbing is needed to achieve that, so that the client is aware of each scenario starting, but it makes sense to have as there are other things we need to run before each scenario.

lib/maze/client/appium/base_client.rb
```diff
@@ -86,6 +91,15 @@ def start_driver(config)
           end
         end
 
+        def start_scenario
+          # Write Maze's address to file and push to the device
+          Maze::Api::Appium::FileManager.new.write_app_file("{\"maze_address\": \"#{maze_address}\"}",
+                                                            FIXTURE_CONFIG)
```
imjoehaines, line 97: Might be nicer to avoid escaping the quotes, e.g.:

```suggestion
          Maze::Api::Appium::FileManager.new.write_app_file(
            JSON.generate({ maze_address: maze_address }),
            FIXTURE_CONFIG
          )
```

or:

```suggestion
          Maze::Api::Appium::FileManager.new.write_app_file(
            %Q({ "maze_address": #{maze_address} }),
            FIXTURE_CONFIG
          )
```

imjoehaines (approved):

lib/maze/client/appium/base_client.rb
twometresteve: Agreed.  I wasn't keen on it when I did it, but didn't stop long enough to consider an alternative.

PR #460, opened 2023-01-25
Add a step to assert no spans were received

lemnik:
## Goal
Add a step to assert that no spans have been received, regardless of the number of traces sent (ie: only empty traces were sent). This is required to ensure that client-side sampling is working as expected.

## Tests
Tested locally with a dependant feature.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 7.14.0 - 2023/01/25
+
+## Enhancements
+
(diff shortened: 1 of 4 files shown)


twometresteve:
Can you add a changelog entry and update the `VERSION` in `maze.rb`.  v7.4.0 has an example of the wording to follow.

You can either set the release date as TBD, or today if you'd like me to release it once merged.

twometresteve (approved):

PR #463, opened 2023-01-27
Add support for non-JSON content types to the reflective servlet

twometresteve:
## Goal

Allow content types other than `application/json` to be used with POST requests in the reflective servlet.

## Design

Although it's not usual in the real world, the servlet will extract its instructions from the query string in this case, avoiding the need to consider specific content types now or over time.

## Tests

e2e tests updated.


test/fixtures/doc-server/features/steps/steps.rb
```diff
@@ -9,7 +9,7 @@
   assert_compare millis, '<', duration
 end
 
-When('I make a reflective {word} request on port {string} with status {string} and delay of {string}') do |method, port, status, delay_ms|
+When('I make a reflective {string} request on port {string} with status {string} and delay of {string}') do |method, port, status, delay_ms|
```
twometresteve, line 12: I needed to change to a `{string}` here as apparently Cucumber doesn't allow hyphens in a `{word}`.

lib/maze/servlets/reflective_servlet.rb
```diff
@@ -26,17 +27,17 @@ def do_GET(request, response)
...
-          $logger.error msg
-          response.status = 415
-          response.body = msg
-          return
-        end
 
-        body = JSON.parse(request.body)
-        delay_ms = body['delay_ms']
-        status = body['status']
+        # For JSON, pull the instruction from the body, other take them from the query parameters
```
yousif-bugsnag:
```suggestion
        # For JSON, pull the instruction from the body, otherwise take them from the query parameters
```

twometresteve: You beat me to my follow-on commit 🙂

yousif-bugsnag (approved):

PR #465, opened 2023-01-30
Map `--device` option to BitBar device group names 

twometresteve:
## Goal

Enhance the existing mechanism for selecting an available device on BitBar.

## Design

The `--device` option is designed to map directly to the name of a device group configured in the BitBar dashboard.  A series of BitBar API calls are used to find the list of devices currently available in that group, using the first in the list.

## Documentation

Docs updated.

## Tests

Tested locally against a number of configured Android device groups.


lib/maze/client/appium/bb_devices.rb
```diff
@@ -71,6 +52,45 @@ def get_device(device_group, platform, platform_version, api_key)
...
+          def find_device_in_group(device_group_id)
+            path = "device-groups/#{device_group_id}/devices"
+            query = {
+              'filter': "online_eq_true"
+            }
+            all_devices = query_api(path, query)
+
+            $logger.debug "All available devices in group #{device_group_id}: #{JSON.pretty_generate(all_devices)}"
+            filtered_devices = all_devices['data'].reject { |device| device['locked'] }
+            return filtered_devices.size, filtered_devices.first
```
imjoehaines, line 77: If the log & exception were in this method we wouldn't need to return two things here

imjoehaines (approved):

lib/maze/client/appium/bb_devices.rb
twometresteve: Haha, I actually refactored this prior to review, moving the log out.  I preferred the multiple return value to having to pass in the group name to log (or not pass it in and only be able to log the group id).

PR #469, opened 2023-02-09
Add iOS 11 and 12 browsers

twometresteve:
## Goal

Add iOS 11 and 12 browsers on BrowserStack.

## Tests

Testing locally using the browser test suite in the repo, e.g:
```
cd test/fixtures/browser
bundle exec maze-runner --farm=bs --browser=iphone_x
```


.buildkite/pipeline.yml
```diff
@@ -281,7 +281,7 @@ steps:
     timeout_in_minutes: 20
     depends_on:
       - "appium-test-fixture"
-      - "push-branch-cli-legacy"
+      - "push-branch-cli"
```
twometresteve, line 284: Corrects a mistake from an earlier PR - the build for that only passed by chance because the step it needs happened to have finished.

gingerbenw: d'oh!

gingerbenw (approved):
LGTM!

PR #470, opened 2023-02-12
Log link to BitBar for each Appium session created

twometresteve:
## Goal

Log link to BitBar for each Appium session created.

## Design

This PR is a little larger than I would have liked, but I found the need to refactor a couple of things as part of it.

A single "maze run" may have multiple Appium session created for it (due to retry behaviour).  For BrowserStack, the link that we already log lands on a search page for the build (a UUID), which all Appium session for the run with share.

For BitBar, however, each Appium session will have a different dashboard link.

I've therefore renamed the logging methods to refer to a "run" rather than "session" (because there could be multiple sessions in a run).  The BrowserStack link is now logged before and after the run (because it's the same link).  The BitBar link(s) are only logged at the end because, whilst most runs only have a single Appium session, we can't be sure of that.

## Changeset

I've also:
- Move the existing BitBar API client code into its own class.
- Refactored the "`linkify` logging method to avoid duplication of code
- Added `--fail-fast` to the Maze Runner CI steps

## Tests

Covered by CI - inspect the output for both the BrowserStack and BitBar steps.


lib/maze/client/bb_api_client.rb
```diff
@@ -0,0 +1,65 @@
...
+
+      def get_device_session_ui_link(session_id)
+        query = {
+          'filter': "clientSideId_eq_#{session_id}"
+        }
+        data = query_api('device-sessions', query)['data']
+        if data.size == 1
+          data[0]['uiLink']
+        else
+          # TODO
```
tomlongridge, line 40: Planning to handle now or in future?

.buildkite/pipeline.yml
```diff
@@ -185,6 +185,7 @@ steps:
           - "--app=build/outputs/apk/release/app-release.apk"
           - "--farm=bs"
           - "--device=ANDROID_7_1"
+          - "--fail-fast"
```
tomlongridge, line 188: I wasn't sure on the justification for adding these fast-fails – is it because we have a retry mechanism, so it should have been on fast fail before?

tomlongridge (approved):
Change LGTM - couple of comments which I'll leave up to your discretion

lib/maze/client/bb_api_client.rb
twometresteve: Thanks, that was an omission - now completed (logging a suitable warning).

.buildkite/pipeline.yml
twometresteve: It's just to get feedback of a failing Buildkite step as soon as possible.  If the MR tests fail it's generally because we've forgotten something obvious and there's no point running the rest of the tests.  It's really annoying when you realise a pipeline has been failing for several minutes before you get any feedback.

One of the main motivations was that the BitBar test, for now, use the secure tunnel and can often have bad runs.  I want them to fail asap so I can retry in that case.  (We can/will move them not to use the secure tunnel, I just wanted something in place quickly and not have to refactor the test fixture).

PR #471, opened 2023-02-15
Set Project and Test Run in the BitBar dashboard

twometresteve:
## Goal

Set the Project and Test Run fields in the BitBar dashboard for each Appium session.

## Design

The capabilities being set here are picked up by the BitBar dashboard to display test runs, grouping them by project.

## Changeset

Also corrected an error in the Changelog introduced in #470 .

## Tests

Covered by unit tests and by inspection of the BitBar dashboard. E.g:

image

(They are all the same due to retries on the job).


lib/maze/client/appium/base_client.rb
```diff
@@ -58,6 +58,7 @@ def start_driver(config)
                   end
                   result
                 rescue => start_error
+                  $logger.error "Session creation failed: #{start_error}"
```
twometresteve, line 61: Bit of an aside - I messed up during the development of this and it was falling in a heap but giving no feedback to me.

yousif-bugsnag (approved):

PR #475, opened 2023-02-20
Allow --device to specify a BitBar device name

twometresteve:
## Goal

Let `--device` specify a device, rather than just a device group, name on BitBar.

## Design

The main approach for using devices on BitBar is to provide a device group name, e.g. `ANDROID_10`.  We're finding that some test scenarios behave differently on specific devices, so this change makes it possible for the `--device` option to also specify a particular device.

## Documentation

Existing docs tweaked.

## Changeset

- New BitBar REST api method added and existing routines refactored 
- Device allocation logic updated to look for a device if no device group with the name if found

## Tests

Device group use (the primary use) covered by CI.  Specific device use tested locally, e.g:

```
$ bundle exec maze-runner --farm=bb --device='Google Pixel 4 -US' ...
[17:05:19]  INFO: Starting SBSecureTunnel
[17:05:21]  INFO: 1 of 3 devices with name 'Google Pixel 4 -US' are available
[17:05:21]  INFO: Selected device: Google Pixel 4 -US (android 10)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,10 @@
 
 ## Enhancements
  
-- Add link to BitBar dashboard for each Appium session. [470](https://github.com/bugsnag/maze-runner/pull/470)
-- Set Appium capabilities for the BitBar dashboard. [471](https://github.com/bugsnag/maze-runner/pull/471)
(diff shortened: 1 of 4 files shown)


yousif-bugsnag (approved):
LGTM

PR #484, opened 2023-03-01
Start the document server after client hooks run

imjoehaines:
## Goal

Currently the document server is started in Maze Runner's `BeforeAll` block. This runs before any Maze Runner hooks are run, which is usually where Maze Runner configuration would be applied

For example, this doesn't currently work but will after this PR:

```rb
Maze.hooks.before_all do
  Maze.config.document_server_root = 'some/path'
end
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# TBD
+
+## Fixes
+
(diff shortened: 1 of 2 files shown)


twometresteve (approved):
Please just add a CHANGELOG entry.

PR #485, opened 2023-03-07
Add Bugsnag-Span-Sampling to Access-Control-Allow-Headers

gingerbenw:
## Goal

Enable required headers in trace payload


twometresteve (approved):
LGTM, but please can you add a changelog entry.  Also, if you want it released asap, you could update the `VERSION` in `maze.rb` and run `bundle install` (committing the updated `Gemfile.lock`).  I can then release it for you immediately after merge.

PR #487, opened 2023-03-08
Fix integer attribute step looking at stringValue

imjoehaines:


twometresteve (approved):
Hurray for the original reviewer of these steps (me).

PR #490, opened 2023-03-09
Add 'latest' versions of Chrome, Firefox & Edge to BitBar browser list

imjoehaines:
## Goal

BitBar now support using `latest` as a version to get the latest version of a given browser:

image

This PR adds the latest versions of Chrome, Firefox & Edge[^1] to `bb_browsers.yml` so we can use them in other repos

## Tests

Updated the Firefox version to `firefox_latest` in the pipeline

[^1]: I've not added Safari as (presumably) it'd also need the `osVersion` to be `latest` and I'm not sure if that's possible — there are basically no docs on this AFAICT. Safari only updates its major version once per year, so it's not as onerous to update manually compared to other browsers

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -277,7 +277,7 @@ steps:
   # BitBar tests
   #
 
-  - label: 'BitBar web browser test - Firefox v107'
+  - label: 'BitBar web browser test - Firefox latest'
     depends_on: "push-branch-cli"
     timeout_in_minutes: 10
     plugins:
@@ -291,7 +291,7 @@ steps:
         verbose: true
         command:
           - "--farm=bb"
-          - "--browser=firefox_107"
+          - "--browser=firefox_latest"
           - "--aws-public-ip"
           - "--fail-fast"
     env:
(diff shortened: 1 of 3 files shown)


lib/maze/client/selenium/bb_browsers.yml
```diff
@@ -1,3 +1,10 @@
+chrome_latest:
+  platform: 'Windows'
+  osVersion: '11'
+  browserName: 'chrome'
+  version: '108'
```
twometresteve, line 5: Should be latest?

.buildkite/pipeline.yml
```diff
@@ -277,6 +277,29 @@ steps:
...
+          - "--farm=bb"
+          - "--browser=chrome_latest"
+          - "--aws-public-ip"
+          - "--fail-fast"
+    env:
+      RUBY_VERSION: "2.7"
+    concurrency: 5
+    concurrency_group: 'bitbar-web'
+    concurrency_method: eager
+
```
twometresteve, line 302: I'd be happy to remove the `BitBar web browser test - Firefox v107` step now.  The thing it's really testing is general support for browser testing on BitBar, as opposed to specific browser versions.

lib/maze/client/selenium/bb_browsers.yml
imjoehaines: 🙃

twometresteve (approved):

twometresteve (approved):

PR #494, opened 2023-03-10
Add android_pixel7 to bs_browsers.yml

gingerbenw:
## Goal

Support testing browser and device combinations required for js-performance testing


twometresteve:
Let's revisit this in future

PR #496, opened 2023-03-13
Add a passthrough in trace validation for initial sampling payloads

Cawllec:
## Goal

Recently added trace payload validation breaks with full setups due to not allowing an exception through.
This PR adds that exception to the validation logic so a specific payload with an a single key of `resourceSpans` with no elements in the array will pass.

## Tests

Added proper E2E tests to cover the scenario.


twometresteve (approved):
LGTM - I also ran some extra checks locally, doctoring the code to prove that the validation checks are definitely run when they should be.

PR #499, opened 2023-03-17
Add new step and associated method to search for table-defined spans

Cawllec:
## Goal

When testing traces we need to locate spans with a particular name, and verify they contain certain attributes.  This allows that test to be set up via a gherkin table in the style of:

```
        Then a span named "TestSpan" contains the attributes:
            | attribute               | type        | value       |
            | test.bool_value         | boolValue   | true        |
            | test.string_value       | stringValue | frayed_knot |
            | test.int_value          | intValue    | 50          |
            | test.double_value       | doubleValue | 6.4         |
            | test.bytes_value        | bytesValue  | deadbeef    |
```

## Tests

End to end tests have been updated to verify the tests work correctly.  Practical testing has been carried out by @yousif-bugsnag

### DIFF
--- a/lib/features/steps/trace_steps.rb
+++ b/lib/features/steps/trace_steps.rb
@@ -156,12 +156,60 @@ Then('a span field {string} matches the regex {string}') do |attribute, pattern|
   Maze.check.false(selected_attributes.empty?)
 end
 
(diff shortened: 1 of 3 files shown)


lemnik (approved):
LGTM - Can't wait to use it

PR #503, opened 2023-03-22
Add DogStatD metrics logging to report BitBar devices availability to Datadog

Cawllec:
## Goal

Allows us to track the amount of available devices for a particular device group every time we attempt to start a BitBar appium driver.  Also lays a foundation for future Maze-runner metrics as required.

## Tests

This has been tested manually to verify that the required metrics are passed onto Datadog.


lib/maze/client/bb_api_client.rb
```diff
@@ -41,7 +41,9 @@ def find_device_in_group(device_group_id)
         all_devices = query_api(path, query)
 
         $logger.debug "All available devices in group #{device_group_id}: #{JSON.pretty_generate(all_devices)}"
+        Maze::Plugins::DatadogMetricsPlugin.send_gauge('bitbar.device.available', all_devices['data'].size, ['device_group_id'])
         filtered_devices = all_devices['data'].reject { |device| device['locked'] }
+        Maze::Plugins::DatadogMetricsPlugin.send_gauge('bitbar.device.unlocked', filtered_devices.size, ['device_group_id'])
```
twometresteve, line 46: Aren't these just sending the literal string "device_group_id"?  In any case, I think we should send the device group name, not the id.

Cawllec: Yes, you're absolutely correct 🤔 .  I'm just trying to figure out how that got there like that...

lib/maze/client/bb_api_client.rb
```diff
@@ -41,7 +41,9 @@ def find_device_in_group(device_group_id)
         all_devices = query_api(path, query)
 
         $logger.debug "All available devices in group #{device_group_id}: #{JSON.pretty_generate(all_devices)}"
+        Maze::Plugins::DatadogMetricsPlugin.send_gauge('bitbar.device.available', all_devices['data'].size, [device_group_id])
         filtered_devices = all_devices['data'].reject { |device| device['locked'] }
+        Maze::Plugins::DatadogMetricsPlugin.send_gauge('bitbar.device.unlocked', filtered_devices.size, [device_group_id])
```
twometresteve, line 46: can we make it the device group name?  it'll be awkward to map ids to names in datadog, won't it?

twometresteve (approved):

PR #504, opened 2023-03-23
Update upload-app script for use with BitBar

Cawllec:
## Goal

This allows us to upload an app to BitBar early during a build process then use that application ID later in the build to avoid uploading build artefacts more than once.

While the `farm` argument will allow us to switch to BitBar, the default will remain as BrowserStack to maintain continuity with our current pipelines.

## Tests

Tested multiple times to ensure the uploads function as expected with varying amounts of arguments present.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Enhancements
 
 - Add logic for step `a span field {string} equals {int}` [501](https://github.com/bugsnag/maze-runner/pull/501)
+- Update upload-app script to allow BitBar uploads []()
 
 ## Fixes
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Enhancements
 
 - Add logic for step `a span field {string} equals {int}` [501](https://github.com/bugsnag/maze-runner/pull/501)
+- Update upload-app script to allow BitBar uploads []()
```
twometresteve, line 6: Remember to add the link to the PR

lib/maze/client/bb_client_utils.rb
```diff
@@ -17,7 +17,8 @@ class << self
         # Uploads an app to BitBar for later consumption
         # @param api_key [String] The BitBar API key
         # @param app [String] A path to the application file
-        def upload_app(api_key, app)
+        # @param app_id_file [String] the file to write the uploaded app url to BrowserStack
```
twometresteve, line 20:
```suggestion
        # @param app_id_file [String] the file to write the uploaded app url into
```

twometresteve (approved):

PR #505, opened 2023-03-27
Add increment metrics for appium tests starting, successes and failures

Cawllec:
## Goal

Lets us keep track of our overall amount of appium tests, and how many are successful/failing for different device farms.

## Tests

Metrics are successfully being reported to the Datadog dashboard


lib/maze/hooks/appium_hooks.rb
```diff
@@ -30,6 +31,11 @@ def after(scenario)
 
       def after_all
         @client&.log_run_outro
+        if $success
+          Maze::Plugins::DatadogMetricsPlugin.send_increment('appium.test_succeeded')
+        else
+          Maze::Plugins::DatadogMetricsPlugin.send_increment('appium.test_failed')
```
twometresteve, line 37: I can't see that the device farm in use is being included?  We need to be able to distinguish between BB and BS (and I would not send any metrics at all for `:local`).

twometresteve (approved):

PR #506, opened 2023-03-28
Set the default HTTP verb for status setting to 'POST' request only

Cawllec:
## Goal

We've found that when setting status responses for incoming requests, OPTIONS requests can cause inconsistent behaviour between tests, and currently have to be specially accounted for.  Following discussion we believe that desiring a set response to effect the OPTIONS request is generally rare or undesirable behaviour.

This change introduces a default verb of POST for any step that sets a specific number of responses (next/list) if a verb isn't already specified.  This should have minimal impact across the vast majority of tests, but will put the status code steps more in line with their expected purpose.

## Tests

Existing HTTP status tests have been updated to cover the new defaulting behaviour where it isn't covered already.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# TBD
+
+## Enhancements
+
+- Set default verb for HTTP status code steps to "POST" [506](https://github.com/bugsnag/maze-runner/pull/506)
+
 # 7.24.0 - 2023/03/30
 
 ## Enhancements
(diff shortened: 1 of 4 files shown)


imjoehaines (commented):
Would we ever want to be able to change the status code of an OPTIONS request?

Just wondering if it's easier to always respond to OPTIONS with a 200

imjoehaines:
Should this apply to "`I set the HTTP status code to {int}`" as well?

Cawllec:
I've agreed with your approach @imjoehaines and set POST as the default verb for all steps that don't specify a verb explicitly, and updated the tests accordingly

imjoehaines (approved):

PR #508, opened 2023-03-30
Use URI.open directly

twometresteve:
## Goal

Fix an error in the `I open the URL {string}` step when running with Ruby 3.  Corresponds to the following warning in Ruby 2:
```
warning: calling URI.open via Kernel#open is deprecated, call URI.open directly or use URI#open
```

## Tests

Reran the test in `bugsnag-java` that uncovered it.


Cawllec (approved):

DavidPetran (approved):
LGTM

PR #512, opened 2023-04-11
Add appium: prefix to non-specific capabilities

Cawllec:
## Goal

Updates capabilities to be appium 2.0 compatible, required for supporting newer versions of OSs going forward.  These should still be backwards compatible with older versions of appium.

This is required for us to move our iOS 16 tests from BrowserStack to BitBar

## Tests

A full cocoa test run is present [here](https://buildkite.com/bugsnag/bugsnag-cocoa/builds/6815#_)
A full android test run is present [here](https://buildkite.com/bugsnag/bugsnag-android/builds/9253#_)


lib/maze/client/appium/bb_client.rb
```diff
@@ -27,16 +27,16 @@ def start_scenario
...
               # Some capabilities probably belong in the top level
               # of the hash, but BitBar picks them up from here.
               'apiKey' => config.access_key,
               'app' => config.app,
               'findDevice' => false,
-              'testTimeout' => 7200,
-              'newCommandTimeout' => 0
-            }
+              'testTimeout' => 7200
+            },
```
twometresteve, line 38: I've got some doubts here about moving the capability out of `bitbar:options`, having seen the BitBar can be very particular about where capabilities are put.  Having said that, the behaviour of BitBar at times was very surprising and perhaps this is a good change for ironing that out.

Are we certain that BitBar picks up the capabilities in these locations and that they actually take effect?  If so, I'm happy to go with it.

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# TBD
+
+## Enhancements
+
+- Update appium capabilities for appium 2.0 compatibility [512](https://github.com/bugsnag/maze-runner/pull/512)
```
twometresteve, line 5: I still haven't released 7.27.0, so you could add this into there and bump the date.

lib/maze/client/appium/bb_client.rb
Cawllec: I don't believe this should be a problem, as it's a core appium capability and any custom setup BitBar uses shouldn't effect appium picking it up.

twometresteve (approved):

PR #513, opened 2023-04-12
Improve handling when failing to get BitBar credentials

imjoehaines:
## Goal

Currently nothing checks if we successfully get BitBar credentials, so if the 30 minute timeout fails then we crash as `account_credentials` returns `false` and the code tries to use that as the credentials:

image

This PR changes this to raise an error explaining that fetching the credentials failed

I've also fixed the retry interval saying it's 30 seconds when it's actually 10 seconds


twometresteve (approved):
CHANGLOG please, but looks ok.  FWIW I am hoping to dispense with the need for this mechanism entirely in due course.

PR #514, opened 2023-04-12
Fix issue causing nil access error in some span steps

Cawllec:
## Goal
Fixes an issue blocking Cocoa performance tests using a couple of specific steps.

Note - this won't fix everything across the board, and further work is being ticketed to refactor the steps for proper testing.


twometresteve (approved):
LGTM, just the PR number/link top add in the changelog of course.

PR #515, opened 2023-04-17
Fix crash when trying to retry browser tests

imjoehaines:
## Goal

Fixes a crash when trying to retry browser tests:
image

`refresh` doesn't exist on the Selenium WebDriver, it's on the `navigate` object: https://www.selenium.dev/documentation/webdriver/interactions/navigation/#refresh

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# TBD
+
+## Fixes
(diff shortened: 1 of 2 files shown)


twometresteve (approved):
looks good, but please add a changelog

PR #516, opened 2023-04-19
Improve JSON schema validation failure messages

imjoehaines:
## Goal

Use `JSONSchemer::Errors.pretty` to output a nicer JSON schema validation failure message:

- https://www.rubydoc.info/gems/json_schemer/0.2.24/JSONSchemer/Errors
- https://github.com/davishmcclurg/json_schemer/blob/master/lib/json_schemer/errors.rb

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,16 +3,13 @@
 ## Enhancements
 
 - Allow BitBar Selenium server URL to be configured [524](https://github.com/bugsnag/maze-runner/pull/524)
+- Improve JSON schema validation failure messages [516](https://github.com/bugsnag/maze-runner/pull/516)
 
 ## Removals
 
 - Remove absolute browser versions no longer supported by BitBar [523](https://github.com/bugsnag/maze-runner/pull/523)
   
 # 7.26.1 - 2023/04/20
-
-## Enhancements
-
-- Improve JSON schema validation failure messages [516](https://github.com/bugsnag/maze-runner/pull/516)
 
 ## Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # TBD
 
+## Enhancements
+
+- Improve JSON schema validation failure messages [516](https://github.com/bugsnag/maze-runner/pull/516)
+
 ## Fixes
 
 - Fix crash when trying to retry browser tests [515](https://github.com/bugsnag/maze-runner/pull/515)
--- a/lib/features/steps/request_assertion_steps.rb
+++ b/lib/features/steps/request_assertion_steps.rb
@@ -44,9 +44,9 @@ def verify_schema_matches(list, list_name)
     if schema_errors.size > 0
       passed = false
       $stdout.puts "\n"
-      $stdout.puts "\e[31m--- #{list_name} #{index} failed validation with errors at the following locations:\e[0m"
+      $stdout.puts "\e[31m--- #{list_name} #{index} failed validation:\e[0m"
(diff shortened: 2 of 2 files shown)


imjoehaines:
The messages depend on the way it failed, but basically instead of always being:

> /some/json/path failed to match /some/other/path

You can get a more useful message based on the way it failed validation, e.g.:

> property '/some/json/path' is missing required keys: a, b, c

> property '/some/json/path' is not of type: integer

Sometimes the errors are no better though, e.g. the error Karl got this morning:

> /resourceSpans/0/scopeSpans/0/spans/0/parentId failed to match /definitions/opentelemetry.proto.trace.v1.Span/additionalProperties

is equally unhelpful:

> property '/resourceSpans/0/scopeSpans/0/spans/0/parentId' is invalid: error_type=schema

Cawllec:
> Sometimes the errors are no better though, e.g. the error Karl got this morning:

> /resourceSpans/0/scopeSpans/0/spans/0/parentId failed to match 
 /definitions/opentelemetry.proto.trace.v1.Span/additionalProperties

> is equally unhelpful:

> property '/resourceSpans/0/scopeSpans/0/spans/0/parentId' is invalid: error_type=schema

That was mostly my reasoning not to use pretty mode - while it may be slightly more helpful in some comparisons, having a path-to-path comparison will always point to the solution so long as both paths are present, which isn't always the case in pretty mode.

imjoehaines:
> That was mostly my reasoning not to use pretty mode - while it may be slightly more helpful in some comparisons, having a path-to-path comparison will always point to the solution so long as both paths are present, which isn't always the case in pretty mode.

I don’t think:

> /resourceSpans/0/scopeSpans/0/spans/0/parentId failed to match /definitions/opentelemetry.proto.trace.v1.Span/additionalProperties

is better than:

> property ‘/resourceSpans/0/scopeSpans/0/spans/0/parentId’ is invalid: error_type=schema

Both require you to know something very unclear to determine what the problem is (`additionalProperties`/`error_type=schema`)

If you know json schemas then `additionalProperties` is kind of decipherable as it’s what determines if the object is allowed to have properties that aren’t specified in the schema — if you get an `additionalProperties` error then you can guess it means that there was an extra property and `additionalProperties` is `false`, but you have to know about json schemas first

`error_type=schema` seems similarly understandable: the json didn’t match the schema 🤷‍♀️

The other errors are way better and I think they would be more common as validation failures going forward — it’s harder to use the wrong property name than sending a badly formed value

For example, take Karl's error from this morning:

> /resourceSpans/0/scopeSpans/0/spans/0/attributes/0/value/intValue failed to match /definitions/opentelemetry.proto.common.v1.AnyValue/properties/intValue

This doesn't tell you anything until you dig into what `/definitions/opentelemetry.proto.common.v1.AnyValue/properties/intValue` is

With pretty errors it would be the much more understandable:

> "property '/resourceSpans/0/scopeSpans/0/spans/0/attributes/0/value/intValue' is not of type: string"

Cawllec:
If you believe that pretty errors are more legible for debugging, I'm happy to go with that going forward

Cawllec (approved):

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@
   
 # 7.26.1 - 2023/04/20
 
+## Enhancements
+
+- Improve JSON schema validation failure messages [516](https://github.com/bugsnag/maze-runner/pull/516)
+
```
twometresteve, line 16: Ah, this will now need to go into the 7.27.0 section.

PR #518, opened 2023-04-19
Fail upload-app with non-zero exit code

twometresteve:
## Goal

Fail upload-app with non-zero exit code if the BitBar access key is not provided.

## Design

When first using this in a pipeline it took me a while to realise that the previous Buildkite step had errored, but passed in Buildkite due to the exit code of 0.

## Tests

Tested locally.

### DIFF
--- a/bin/upload-app
+++ b/bin/upload-app
@@ -59,9 +59,8 @@ class UploadAppEntry
     end
 
     if options[:access_key].nil?
-      $logger.warn 'An access_key is required to upload the app'
       Optimist::with_standard_exception_handling parser do
-        raise Optimist::HelpNeeded
+        raise Optimist::CommandlineError.new('An access_key is required to upload the app', 1)
       end
     end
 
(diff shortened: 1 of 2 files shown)


bin/upload-app
```diff
@@ -59,9 +59,8 @@ class UploadAppEntry
     end
 
     if options[:access_key].nil?
-      $logger.warn 'An access_key is required to upload the app'
       Optimist::with_standard_exception_handling parser do
-        raise Optimist::HelpNeeded
+        raise Optimist::CommandlineError.new('An access_key is required to upload the app', 1)
```
imjoehaines, line 63: Looks like this is equivalent to:

```suggestion
        parser.die('An access_key is required to upload the app', 1)
```

https://github.com/ManageIQ/optimist/blob/f022eb85c24a30e195b234e5b7722711e27c36d2/lib/optimist.rb#L960-L963

https://github.com/ManageIQ/optimist/blob/f022eb85c24a30e195b234e5b7722711e27c36d2/lib/optimist.rb#L991-L993

twometresteve: Changed

PR #519, opened 2023-04-19
Read app option from file

twometresteve:
## Goal

Allow the id for an app previously uploaded to BitBar to be read from a file.

## Design

This mechanism has been in place for BrowserStack for some time.  If the `--app` options starts with an `@` then Maze Runner will read the contents of the file.  The change had just not been applied on the BitBar side.

## Tests

Tested locally.


richardelms (approved):
Very clever 👍

PR #520, opened 2023-04-19
Enable span validation in span receipt step

Cawllec:
## Goal

Moves validation methods into a separate class that can be invoked from multiple points in maze-runner, adding an invocation to the span expectation step to ensure we validate all incoming trace payloads.


imjoehaines:
Wouldn't it be better to do this outside of any step (i.e. in an after hook), so the validation is always guaranteed to run?

Cawllec:
I've refactored to run the validation after each scenario instead of in specific steps, and I've run the tests against both Android and Cocoa pipelines (and performance pipelines) to verify it works correctly.

twometresteve (approved):

PR #521, opened 2023-04-19
Fix crash in local selenium client

twometresteve:
## Goal

Fixes a recently introduced crash in local selenium client.

## Tests

This crash was readily reproducible by running `bundle exec maze-runner --farm=local --browser=firefox` in `test/fixtures/browser` - fixed after the change.


richardelms (approved):
Super! thanks

PR #523, opened 2023-04-21
Update BitBar browsers list

gingerbenw:
## Goal

BitBar no longer supports some of the browsers contained in our browsers list, this PR updates the list to contain all currently supported chromium browsers.


twometresteve:
As discussed IRL, I think you're better using the `_latest` variants for each browser on BitBar and we can add "latest minus n" if needed.  For this PR, however, I would be happy to see absolute versions of browsers no longer supported by BitBar removed, without adding new absolute versions.  No test could succeed using them if they have been removed, so they are just cluttering the file.

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 7.26.2 - 2023/04/21
+
+## Fixes
+
+- Update BitBar browsers list to match capabilities creator [523](https://github.com/bugsnag/maze-runner/pull/523)
```
twometresteve, line 5:
```suggestion
## Removals

- Remove absolute browser versions no longer supported by BitBar [523](https://github.com/bugsnag/maze-runner/pull/523)
```

twometresteve (approved):

PR #525, opened 2023-05-02
Provide access to document server via AWS public IP addresses

twometresteve:
## Goal

Allow (primarily browser-based) clients access to the document server through AWS EC2 public IP addresses.

## Design

Follows the same pattern used for access to the mock server through public IPs.  The document server just wasn't considered when it was introduced.

## Changeset

In addition to new plumbing, I've updated the CI step for browser tests to use the new functionality.  I've had to put it onto Safari as only those browsers are allowed by our firewall at present. 

## Tests

Covered by CI updaes.

### DIFF
--- a/lib/features/support/internal_hooks.rb
+++ b/lib/features/support/internal_hooks.rb
@@ -46,8 +46,14 @@ BeforeAll do
 
   # Determine public IP if enabled
   if Maze.config.aws_public_ip
-    Maze.public_address = Maze::AwsPublicIp.new.address
+    public_ip = Maze::AwsPublicIp.new
+    Maze.public_address = public_ip.address
     $logger.info "Public address: #{Maze.public_address}"
+
+    unless Maze.config.document_server_root.nil?
+      Maze.public_document_server_address = public_ip.document_server_address
+      $logger.info "Public document server address: #{Maze.public_document_server_address}"
+    end
   end
 
   # Start mock server
--- a/lib/features/support/internal_hooks.rb
+++ b/lib/features/support/internal_hooks.rb
@@ -44,18 +44,6 @@ BeforeAll do
   Maze.run_uuid = SecureRandom.uuid
   $logger.info "UUID for this run: #{Maze.run_uuid}"
 
-  # Determine public IP if enabled
-  if Maze.config.aws_public_ip
-    public_ip = Maze::AwsPublicIp.new
-    Maze.public_address = public_ip.address
-    $logger.info "Public address: #{Maze.public_address}"
-
-    unless Maze.config.document_server_root.nil?
-      Maze.public_document_server_address = public_ip.document_server_address
-      $logger.info "Public document server address: #{Maze.public_document_server_address}"
-    end
-  end
-
   # Start mock server
   Maze::Server.start
   Maze::Server.set_response_delay_generator(Maze::Generator.new [Maze::Server::DEFAULT_RESPONSE_DELAY].cycle)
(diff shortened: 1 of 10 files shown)


lib/features/support/internal_hooks.rb
```diff
@@ -66,6 +60,18 @@
...
+    public_ip = Maze::AwsPublicIp.new
+    Maze.public_address = public_ip.address
+    $logger.info "Public address: #{Maze.public_address}"
+
+    unless Maze.config.document_server_root.nil?
+      Maze.public_document_server_address = public_ip.document_server_address
+      $logger.info "Public document server address: #{Maze.public_document_server_address}"
+    end
+  end
+
```
twometresteve, line 74: This needed moving to after the hooks are run, so that any config we set is taken into account (`Maze.config.document_server_root` in this case).

test/fixtures/browser/features/steps/browser_steps.rb
```diff
@@ -1,6 +1,35 @@
...
+def get_private_hostname
+  case Maze.config.farm
+  when :local
+    'localhost'
+  when :bs
+    'bs-local.com'
+  else # :bb
+    'local'
+  end
+end
```
imjoehaines, line 26: Is this right? What are `bs-local.com` and `local` and how do they translate into a reachable host?

We just use `Maze.config.bind_address` and `Maze.config.document_server_bind_address` in the SDK and it works as expected on BrowserStack, BitBar and locally

test/fixtures/browser/features/steps/browser_steps.rb
```diff
@@ -1,6 +1,35 @@
+def get_document_server_url
+  if Maze.config.aws_public_ip
+    "http://#{Maze.public_document_server_address}"
+  else
+    "http://#{get_private_hostname}:#{Maze.config.document_server_port}"
+  end
```
imjoehaines, line 6: Presumably we'd need to do this in the SDKs, so can Maze Runner provide an API?

Currently we're doing this:

```ruby
maze_uri = URI("http://#{Maze.config.bind_address}:#{Maze.config.port}")
fixtures_uri = URI("http://#{Maze.config.document_server_bind_address}:#{Maze.config.document_server_port}")
```

lib/maze/client/selenium/bb_client.rb
```diff
@@ -4,7 +4,7 @@ module Selenium
       class BitBarClient < BaseClient
         def start_session
           config = Maze.config
-          if ENV['BUILDKITE']
+          if ENV['BUILDKITE'] && !Maze.config.aws_public_ip
```
imjoehaines, line 7: This could be in the `if` on line 19 to avoid checking for the public IP twice

lib/maze/client/appium/bb_client.rb
```diff
@@ -6,7 +6,7 @@ def prepare_session
           config = Maze.config
           config.app = Maze::Client::BitBarClientUtils.upload_app config.access_key,
                                                                   config.app
-          if Maze.config.start_tunnel
+          if Maze.config.start_tunnel && !Maze.config.aws_public_ip
```
imjoehaines, line 9: This is duplicated 4 times and seems to be the new requirement for whether to start & stop the tunnel

Can we add a method for this, e.g. `Maze::Client::BitBarClientUtils.use_local_tunnel?`

imjoehaines: This could also be:

```suggestion
          if Maze::Client::BitBarClientUtils.use_local_tunnel?
            @started_local_tunnel = true
            # ...
```

Then `stop_session` can just check for `@started_local_tunnel`

test/fixtures/browser/features/steps/browser_steps.rb
twometresteve: That would seem to make sense, but I think I'll add it separately when I come to move the JS/perf pipelines over.

lib/maze/client/selenium/bb_client.rb
twometresteve: Yup, that's pretty nasty.

tomlongridge (approved):

PR #526, opened 2023-05-03
Only use separate BitBar accounts on CI

twometresteve:
## Goal

Fix an issue running browser tests on BitBar on local machines.

## Design

On CI, if the secure tunnel is used, we use multiple BitBar users as each user can only start a single secure tunnel.  Locally, there is no need to have separate users - local user accounts are sufficient.

This change was introduced in a recent PR, during a time that the secure tunnel was broken and so we couldn't test it locally.

## Changeset

I've made some changes to the browser test fixture so that the tests can be run on Chrome, Firefox or Safari on a local machine.

I've also bumped the version of Android in use for the legacy protocol (JsonWP) tests, as BrowserStack are increasingly short of Android 7 devices, causing unnecessary test failures.  We don't need to test on an old device to test the legacy protocol.

## Tests

Tested locally and CI covers that usage.

### DIFF
--- a/test/fixtures/browser/features/steps/browser_steps.rb
+++ b/test/fixtures/browser/features/steps/browser_steps.rb
@@ -24,7 +24,7 @@ def get_private_hostname
     'bs-local.com'
   else # :bb
     # Get the IP address of the local machine - this is the only way I could
-    # find to make it with with Firefox, Chrome and Safari over the tunnel.
+    # find to make it work with Firefox, Chrome and Safari over the tunnel.
     addr_infos = Socket.ip_address_list.reject( &:ipv6? )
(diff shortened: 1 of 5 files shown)


test/fixtures/browser/features/steps/browser_steps.rb
```diff
@@ -21,14 +23,20 @@ def get_private_hostname
   when :bs
     'bs-local.com'
   else # :bb
-    'local'
+    # Get the IP address of the local machine - this is the only way I could
+    # find to make it with with Firefox, Chrome and Safari over the tunnel.
```
twometresteve, line 27: Typo `with with` corrected and ready to push

tomlongridge (approved):

PR #530, opened 2023-05-10
Allow multiple device groups to be listed

twometresteve:
## Goal

Allow multiple device group name for BitBar, from which an available device will be selected at random.

E.g. `--device=ANDROID_11|ANDROID_12`.

## Design

This change was slightly more complicated than I anticipated, as I needed to refactor some methods a little to handle multiple API calls.

## Documentation

Docs updated.

## Tests

Covered by CI and I'm also performed some testing locally with both single and multiple device group names.


Cawllec (approved):
LGTM 👍

PR #531, opened 2023-05-11
Forward trace requests to Aspecto

twometresteve:
## Goal

Allow trace requests received in tests to be forwarded to Aspecto.

## Design

This takes the same basic approach as the existing BugSnag repeater mechanism. I've just refactored the existing module that does the repeating into a class hierarchy to allow common code to be shared across the Aspecto and BugSnag repeaters.

Aspecto does not appear to accept gzipped payloads, so if the `Content-Encoding` header is set we'll decompress it and remove the header before repeating.

## Changeset

- Existing Repeater module refactored into a class hierarchy
- The BugSnag repeater will now use SSL.  Whilst it was aiming at an `https` endpoint, it was not actually using SSL.  BugSnag tolerates that, but Aspecto does not.

## Tests

regression of basic MR functionality covered by CI.  I've also tested locally that errors can be repeated to BugSnag and traces to Aspecto, then the relevant environments variables are set.


lib/maze/option/parser.rb
```diff
@@ -41,7 +41,11 @@ def parse(args)
                 type: :boolean,
                 default: true
 
-            opt Option::REPEATER_API_KEY,
+            opt Option::ASPECTO_REPEATER_API_KEY,
+                'Enables forwarding of all received POST requests to Aspecto, using the API key provided.  MAZE_ASPECTO_REPEATER_API_KEY may also be set.',
+                type: :string
+
+            opt Option::BUGSNAG_REPEATER_API_KEY,
                 'Enables forwarding of all received POST requests to Bugsnag, using the API key provided.  MAZE_REPEATER_API_KEY may also be set.',
```
tomlongridge, line 49: MAZE_BUGSNAG_REPEATER_API_KEY?

tomlongridge (approved):
Not sure whether you deliberately left some of the existing names without BUGSNAG prefix? But works either way I guess.

lib/maze/option/parser.rb
twometresteve: I considered it, but it would have been a breaking change.  I also considered having that as an alternative, but on balance found it unnecessary - the Aspecto forwarder is more niche and the existing one is implicitly "BugSnag".

PR #532, opened 2023-05-12
Correct setting of headers for Bugsnag repeater

twometresteve:
## Goal

Correct the setting of HTTP headers for the BugSnag repeater.

## Design

All Bugsnag headers were being stripped from requests.  This is correct for the Aspecto repeater as we don't want to leak an API key, for example, but of course they are needed when repeating to Bugsnag.  This change refactors the code to call into methods in each subclass.

## Changeset

Header logic pushed down into the subclasses.

## Tests

I neglected to test the repeating of traces to BugSnag in #531 (testing only the repeating of errors).  This time I've testing the repeating of traces to Aspecto as well as both errors and traces to BugSnag.  I've also verified that no trace servlet BugSnags are created by the change.


tomlongridge (approved):
Assume the logging was intended to stay in and not for debugging - LGTM

PR #537, opened 2023-05-17
Store sampling requests in their own list 

twometresteve:
## Goal

Store trace sampling requests in their own list, allowing them to be inspected separately and avoiding flakes caused by race conditions between the receipt of them and regular trace requests.

## Design

The `Servlet` has been refactored slightly allowing the `TraceServlet` to put received requests into the appropriate list.

A custom Cucumber parameter type of `request_type` has been created, allowing arguments to have space in them without using speech marks (this can't be done with either `{word}` or `{string}` parameters).

## Documentation

Upgrading guide updated.

## Tests

CI tests updated and I've also updated the entire set of `bugsnag-android-performance` tests to run against it.

### DIFF
--- a/lib/maze/schemas/trace_validator.rb
+++ b/lib/maze/schemas/trace_validator.rb
@@ -25,8 +25,6 @@ module Maze
       # Runs the validation against the trace given
       def validate
         @success = true
-        # Shortcut the validation if the body is empty for initial P gathering reasons
-        return if @body.keys.eql?(['resourceSpans']) && @body['resourceSpans'].empty?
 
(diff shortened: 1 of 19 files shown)


lib/maze/schemas/trace_validator.rb
```diff
@@ -25,8 +25,6 @@ def initialize(body)
       # Runs the validation against the trace given
       def validate
         @success = true
-        # Shortcut the validation if the body is empty for initial P gathering reasons
-        return if @body.keys.eql?(['resourceSpans']) && @body['resourceSpans'].empty?
```
twometresteve, line 29: This is no longer needed thanks to the code in `validator.rb`.  `validator_class` will be `nil` for p-value requests as they are now held on a separately named list.

tomlongridge (approved):
LGTM

PR #544, opened 2023-05-25
Add JS performance required BS devices to driver

Cawllec:
## Goal

Adds two devices required for testing JS performance to the BS driver:
- iphone_14, providing iOS 16
- Pixel 7, providing an up-to-date chrome driver

This _doesn't_ add an iOS 11 device, as that is already covered by the `iphone_x` device.

(Thoughts on changing the naming convention as a v8 release feature (in a separate PR)?  I was thinking just changing them to the OS version, so `iphone_14` just becomes `iOS_16` etc)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# TBD
+
+## Enhancements
+
+- Add iOS 16 and an up to date Android device browsers to the available BS devices [544](https://github.com/bugsnag/maze-runner/pull/544)
+
 # 7.32.0 - 2023/05/19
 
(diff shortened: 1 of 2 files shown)


twometresteve:
> (Thoughts on changing the naming convention as a v8 release feature (in a separate PR)? I was thinking just changing them to the OS version, so iphone_14 just becomes iOS_16 etc)

Yes, I agree that would be a good idea.  Could you raise a ticket and link it to the v8 epic please?

twometresteve (approved):

PR #545, opened 2023-05-26
Rename BrowserStack mobile browsers to OS and version format

Cawllec:
## Goal

To make it clearer which devices and versions are being targeted when running tests on mobile browsers, the existing browsers have been renamed to an `os_version` format.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -23,6 +23,8 @@ The `VERBOSE` environment variable no longer has any effect on the log level.  U
 
 The names for each Android device on BrowserStack have been shortened, e.g. ANDROID_10_0 becomes ANDROID_10.  This is for consistency with BrowserStack iOS devices and BitBar device groups.  Run `bundle exec maze-runner --farm=bs --list-devices` for a full list of options.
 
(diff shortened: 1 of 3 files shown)


UPGRADING.md
```diff
@@ -23,6 +23,8 @@ The `VERBOSE` environment variable no longer has any effect on the log level.  U
 
 The names for each Android device on BrowserStack have been shortened, e.g. ANDROID_10_0 becomes ANDROID_10.  This is for consistency with BrowserStack iOS devices and BitBar device groups.  Run `bundle exec maze-runner --farm=bs --list-devices` for a full list of options.
 
+In addition, browsers running on mobile devices are now named after the OS and version, e.g. `android_nexus5` becomes `android_4`.  This should make the desired device/browser combination more obvious in future.
+
```
twometresteve, line 27: Is there an argument for listing all the old and new names out in a table here?

twometresteve (approved):
Nice change - we might just make it a bit more friendly for people by putting all the old and new names in UPGRADING.

UPGRADING.md
Cawllec: Yes, and it's done

PR #547, opened 2023-06-08
Remove TMS functionality

Cawllec:
## Goal

This is everything AFAICT, and the existing tests should cover all scenarios involved (BitBar appium, BitBar Selenium) to ensure nothing breaks.


lib/features/support/internal_hooks.rb
```diff
@@ -90,8 +90,6 @@
   # Only add the retry plugin if --retry is not used on the command line
   config.filters << Maze::Plugins::GlobalRetryPlugin.new(config) if config.options[:retry].zero?
   config.filters << Maze::Plugins::BugsnagReportingPlugin.new(config)
-  cucumber_report_plugin = Maze::Plugins::CucumberReportPlugin.new
-  cucumber_report_plugin.install_plugin(config)
```
twometresteve, line 94: Ah, I didn't mean to blast this away necessarily, I forgot that the TMS did more than just distribute BitBar user account numbers.

We can keep the other functionality of the TMS - delivering of test reports.  We may be able to build on it over time, even if it did get parked previously.

twometresteve (approved):

PR #549, opened 2023-06-12
Reword steps for consistency

twometresteve:
## Goal

Change a few Cucumber step definitions to make for more consistency use of speech marks when using them.

## Documentation

UPGRADING updated.

## Tests

Covered by CI.


twometresteve:
Let's go the other way with this and make the list not have speech marks

PR #550, opened 2023-06-12
Remove speech marks from status codes list

twometresteve:
## Goal

Avoid the need for speech marks around the list of status codes, making it consistent with the Cucumber steps that accept 
just a single integer.

## Documentation

UPGRADING guide updated.

## Tests

Covered by CI.

### DIFF
--- a/lib/features/support/cucumber_types.rb
+++ b/lib/features/support/cucumber_types.rb
@@ -9,5 +9,5 @@ ParameterType(
   name:        'int_array',
   regexp:      /\d+(?:, ?\d+)+/,
   type:        String,
-  transformer: ->(s) { s.split(',').map(&:strip) }
+  transformer: ->(s) { s.split(',').map(&:strip).map(&:to_i) }
 )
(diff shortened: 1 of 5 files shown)


lib/features/support/cucumber_types.rb
```diff
@@ -4,3 +4,10 @@
   type:        String,
   transformer: ->(s) { s }
 )
+
+ParameterType(
+  name:        'int_array',
+  regexp:      /\d+(?:, ?\d+)+/,
+  type:        String,
+  transformer: ->(s) { s.split(',').map(&:strip) }
```
imjoehaines, line 12: Probably doesn't matter but this could also convert the codes to integers

```suggestion
  transformer: ->(s) { s.split(',').map(&:strip).map(&:to_i) }
```

imjoehaines (approved):

lib/features/support/cucumber_types.rb
twometresteve: Don't think it makes a difference for this change as we just go and join it back together into a string, but it might make a difference in future if we use the `int_array` type for something else.

PR #558, opened 2023-06-16
Fix and improve logging of received requests

twometresteve:
## Goal

Fixes a crash when trying to log gzipped trace requests, as well as making general improvements to the logging of requests.

## Design

Previous we tried to log the whole raw request in addition to the JSON formatted payload.  This failed for gzipped payloads, as the `request.body` was still in its binary form, but really logged the raw request wasn't that useful anyway.  Instead I've opted for logging specific fields within it, meaning we have more control of what is being logged.

Responses sent back from Maze Runner are now stored in the `RequestList`, allow us to log their headers and bodies too.

## Tests

Tested locally, inspecting the `maze_output` contents and by forcing failures in a couple of pipelines:
[Android Performance (traces)](https://buildkite.com/bugsnag/bugsnag-android-performance/builds/999#0188c486-b214-4f34-9d93-a51f4373a0d6/461)
[Unity (errors)](https://buildkite.com/bugsnag/bugsnag-unity/builds/3974#0188c478-f1b6-434f-979e-a0efe29a8708)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 8.1.0 - 2023/06/16
+
+## Enhancements
+
+- Fixes and enhancements to logging of received requests [558](https://github.com/bugsnag/maze-runner/pull/558)
+
 # 8.0.2 - 2023/06/15
 
 ## Fixes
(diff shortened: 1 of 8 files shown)


lemnik (approved):
LGTM :rocket:

PR #560, opened 2023-06-20
Update the delete method for purging projects on bitbar

joshedney:
## Goal

Add the `delete` method to the API request within the `delete_project` function

## Tests

Built and tested locally


twometresteve (approved):
Looks good, please just add a CHANGELOG entry

PR #566, opened 2023-07-03
Include traces when setting HTTP response codes

twometresteve:
## Goal

Trace requests got missed from #565 .

## Tests

Tested locally using a bugsnag-unity-performance test that found the issue.


richardelms (approved):
thanks!

PR #567, opened 2023-07-05
Correct logic for HTTP response codes to session requests

twometresteve:
## Goal

Third attempt following #565 and #566.

## Tests

Covered by CI and rerunning another Unity test.


richardelms (approved):
thanks

PR #568, opened 2023-07-06
build: security patch for nokogiri >1.13

tomlongridge:
## Goal

`nokogiri` has some CVEs (CVE-2022-2309, CVE-2022-40304, CVE-2022-40303) but `appium_lib` doesn't force the version to be high enough. This has been locked to 1.13 in MR, but as it's not declared in the Gemfile, all the downstream repos get the older version, and associated CVEs.

## Design

Therefore, I've added `nokogiri` as an explicit dependency of MR. Not entirely sure if this the best approach or a reasonable thing to do, but thought I'd put this up for discussion.


twometresteve (approved):
Please just add a CHANGELOG entry - just in case it does break something unforeseen.

tomlongridge:
Closing in favour of lockfile updates in downstream libraries.

PR #569, opened 2023-07-17
Add mechanism to allow hooks before scenario has completed in full

twometresteve:

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -118,7 +118,7 @@ GEM
     minitest (5.18.1)
     mocha (1.13.0)
     multi_test (0.1.2)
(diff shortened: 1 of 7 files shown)


Cawllec (approved):
LGTM, are you ok with the docker-compose output code being left in the change set, perhaps as an example for logs if we need it later, or should we ticket it and add these as definitive functionality when docker is used?

twometresteve:
> LGTM, are you ok with the docker-compose output code being left in the change set, perhaps as an example for logs if we need it later, or should we ticket it and add these as definitive functionality when docker is used?

Ah, thanks - that was a copy/paste error, now removed.

PR #570, opened 2023-07-18
Add endpoints for bugsnag-cli testing

joshedney:
## Goal

Add additional sourcemaps endpoints so that we can use MazeRunner to test the Bugsnag CLI.

## Changeset

Add the following endpoints to the mock server:
`/dart-symbol`
`/ndk-symbol`
`/proguard`

## Tests

Updated tests, covered by CI

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 8.1.5 - TBD
+
+- Add additional sourcemap endpoints [570](https://github.com/bugsnag/maze-runner/pull/570)
+
 # 8.1.4 - 2023/07/05
 
 ## Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
-# 8.1.5 - TBD
+# 8.2.0 - 2023/07/18
+
+## Enhancements
 
 - Add additional sourcemap endpoints [570](https://github.com/bugsnag/maze-runner/pull/570)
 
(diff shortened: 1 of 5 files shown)


twometresteve:
No objection as such, but could the CLI be configured to send to a different endpoint instead?

Also, please add a changelog entry.

joshedney:
> No objection as such, but could the CLI be configured to send to a different endpoint instead?
> 
> Also, please add a changelog entry.

We can overwrite the server address but the endpoints themselves are hardcoded as these don't change regardless if its SASS or on prem. We could look at adding a `endpoint` flag but this would only be used for interacting with MR. Which would you prefer ?

lib/maze.rb
```diff
@@ -7,7 +7,7 @@
 # Glues the various parts of MazeRunner together that need to be accessed globally,
 # providing an alternative to the proliferation of global variables or singletons.
 module Maze
-  VERSION = '8.1.4'
+  VERSION = '8.1.5'
```
twometresteve, line 10: I think I'd got for 8.2.0 - whilst minor, it does constitute and Enhancement rather than a Fix.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 8.1.5 - TBD
+
+- Add additional sourcemap endpoints [570](https://github.com/bugsnag/maze-runner/pull/570)
```
twometresteve, line 3: Please add an `## Enhancements` header.

twometresteve (approved):

PR #571, opened 2023-07-19
Add received span steps to allow greater flexibility in performance tests

Cawllec:
## Goal

In addition, add a step to allow a ranged amount of any request to be specified.

## Tests

Both positive and negative test steps have been added to our test fixtures.


Cawllec:
This won't pass test-wise at the moment, as a fix in a previous PR needs to be merged first.  This can then be rebased on that.

twometresteve (approved):
Nice change - I like the precedent it continues to set over the testing of new steps.

PR #573, opened 2023-07-21
Add validation of trace request headers

twometresteve:
## Goal

Adds validation of the following headers on every trace request received
- Bugsnag-Api-Key – 32 hex char guid
- Bugsnag-Sent-At – ISO 8601 date
- Bugsnag-Span-Sampling – of the format x:y where
  - x is a decimal between 0 and 1 (inclusive)
  - y is an integer 

## Design

Existing mechanism extended to include steps for validating headers.

## Tests

Unit and e2e tests updated and I've also run the whole set of Android and Cocoa performance e2e tests against it.

### DIFF
--- a/lib/maze/schemas/trace_validator.rb
+++ b/lib/maze/schemas/trace_validator.rb
@@ -47,11 +47,6 @@ module Maze
         )
       end
 
-      def check(date)
-        Date.iso8601(date)
-      rescue Date::Error
-      end
-
       def validate_header(name)
         value = @headers[name]
         if value.nil? || value.size > 1
--- a/lib/maze/schemas/trace_validator.rb
+++ b/lib/maze/schemas/trace_validator.rb
@@ -7,7 +7,8 @@ module Maze
 
     HEX_STRING_16 = '^[A-Fa-f0-9]{16}$'
     HEX_STRING_32 = '^[A-Fa-f0-9]{32}$'
-    SAMPLING_HEADER = '^(1(.0)?|0(\.[0-9]+)?):[0-9]+$'
+    SAMPLING_HEADER_ENTRY = '((1(.0)?|0(\.[0-9]+)?):[0-9]+)'
+    SAMPLING_HEADER = "^#{SAMPLING_HEADER_ENTRY}(;#{SAMPLING_HEADER_ENTRY})*$"
 
     # Contains a set of pre-defined validations for ensuring traces are correct
     class TraceValidator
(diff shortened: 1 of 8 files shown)


lib/maze/schemas/trace_validator.rb
```diff
@@ -5,6 +5,10 @@
 module Maze
   module Schemas
 
+    HEX_STRING_16 = '^[A-Fa-f0-9]{16}$'
+    HEX_STRING_32 = '^[A-Fa-f0-9]{32}$'
+    SAMPLING_HEADER = '^(1(.0)?|0(\.[0-9]+)?):[0-9]+$'
```
imjoehaines, line 10: If there's more than 1 sampling probability then they are all listed in the header separated by semicolons, i.e. `a:b;c:d`, `a:b;c:d;e:f` etc&hellip;

twometresteve: Regex updated and new test cases added.

lib/maze/schemas/trace_validator.rb
```diff
@@ -40,6 +45,61 @@ def validate
         )
       end
 
+      def check(date)
```
Cawllec, line 48: This doesn't appear to be used, and has an ambiguous name, could it be removed or renamed?

lib/maze/schemas/trace_validator.rb
```diff
@@ -40,6 +45,61 @@ def validate
...
+          expected = Regexp.new(HEX_STRING_32)
+          unless expected.match(api_key)
+            @success = false
+            @errors << "bugsnag-api-key header was expected to match the regex '#{HEX_STRING_32}', but was '#{api_key}'"
+          end
+        end
+
+        # Bugsnag-Sent-at
+        validate_header('bugsnag-sent-at') do |date|
+          begin
```
Cawllec, line 75: Can this use the method below, `def validate_date(date)`?

lib/maze/schemas/trace_validator.rb
```diff
@@ -15,19 +20,19 @@ class TraceValidator
...
       #
-      #   @param body [Hash] The body of the trace to validate
-      def initialize(body)
-        @body = body
-        @success = nil
+      #   @param request [Hash] The trace request to validate
+      def initialize(request)
+        @headers = request[:request].header
+        @body = request[:body]
+        @success = true
```
Cawllec, line 27: Could this introduce false positives if the validation isn't actually run?

Cawllec (approved):
Pending tests passing

PR #579, opened 2023-08-18
Move maze_output.zip so that it can be easily accessed from docker containers

joshedney:
## Goal

Make the `maze_output.zip` easily accessible from docker containers

## Changeset

Move the `maze_output.zip` to the `maze_output` folder when completed so that it can be easily accessed from within docker containers.

## Tests

Covered via CI


twometresteve (approved):
Gemfile.lock should also be updated, but I can do that in the next PR.

PR #580, opened 2023-08-18
Allow push_file location for Android to be overidden

twometresteve:
## Goal

Allow push_file location for Android to be overidden.

## Changeset

Adds a new config value that can be set to push files to a different location on Android.  iOS have no other viable location to use, so there is no point in adding a config option for that.

## Tests

Tested using a RN Perf branch that needs to read for the temp folder.

### DIFF
--- a/lib/maze/api/appium/file_manager.rb
+++ b/lib/maze/api/appium/file_manager.rb
@@ -9,7 +9,8 @@ module Maze
         end
 
         # Creates a file with the given contents on the device (using Appium).  The file will be located in the app's
-        # Documents directory for iOS and /sdcard/Android/data/<app-id>/ for Android.
+        # documents directory for iOS. On Android, it will be /sdcard/Android/data/<app-id>/files unless
+        # Maze.config.android_app_files_directory has been set.
         # @param contents [String] Content of the file to be written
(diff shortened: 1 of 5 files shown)


lib/maze/api/appium/file_manager.rb
```diff
@@ -9,15 +9,16 @@ def initialize
...
+        # Maze.config.android_app_files_directory has been set.
         # @param contents [String] Content of the file to be written
         # @param filename [String] Name (with no path) of the file to be written on the device
         def write_app_file(contents, filename)
           path = case Maze::Helper.get_current_platform
                  when 'ios'
                    "@#{@driver.app_id}/Documents/#{filename}"
                  when 'android'
-                   "/sdcard/Android/data/#{@driver.app_id}/files/#{filename}"
+                   Maze.config.android_app_files_directory || "/sdcard/Android/data/#{@driver.app_id}/files/#{filename}"
```
yousif-bugsnag, line 21: should this be `Maze.config.android_app_files_directory + filename`?

yousif-bugsnag (approved):

lib/maze/api/appium/file_manager.rb
twometresteve: Yes!

PR #581, opened 2023-08-29
Allow Bugsnag-Uncompressed-Content-Length in CORS header policy

twometresteve:
## Goal

Allow `Bugsnag-Uncompressed-Content-Length` in CORS header policy.

## Tests

Reran the bugsnag-unity-performance test that found the issue.


richardelms (approved):
thanks!

PR #582, opened 2023-09-05
Prefix `autoGrantPermissions` option

lemnik:
## Goal
Fix the error:
```
Selenium::WebDriver::Error::InvalidArgumentError occurred with message: All non-standard capabilities should have a vendor prefix. The following capabilities did not have one: autoGrantPermissions
```

## Tests
Relied on existing tests passing again


Cawllec (dismissed):
I think we should set the default prefix for BitBar to `appium:` from now on, changing that in the `caps_prefix` method further down in this file.  It should look like:

```
def caps_prefix(appium_version)
            appium_version.nil? || (appium_version.to_i >= 2) ? 'appium:' : ''
end
```

tomlongridge (approved):

PR #584, opened 2023-09-14
Do not retry Appium session creation on BitBar devices

twometresteve:
## Goal

Do not retry Appium session creation on BitBar devices, as what appear to be failures can result in orphaned sessions on the bitBar side - leading to us exceeding our concurrency limit.

## Design

This is a light refactor to push the criteria for whether we retry or not down into the farm-specific classes.  If the session creation fails completely then Maze Runner will return a specific exit code, allowing us to perform a calculated retry at the Buildkite level.

## Documentation

Online docs updated.

## Tests

Covered by CI and I've tested the retry mechanism with both BB and BS locally by trying to install an iOS app on an Android phone.


twometresteve:
Actually, I'm not sure this is in line with what `ErrorCodeHook` already does - I need to understand that more I think.

lib/maze/errors.rb
```diff
@@ -24,29 +24,29 @@ def initialize(message='Element not found', element='No element specified')
...
       },
       Maze::Error::AppiumElementNotFoundError => {
         retry: true,
-        error_code: 11
+        error_code: Maze::Api::ExitCode::AUTOMATION_ELEMENT_NOT_FOUND
       },
       ::Selenium::WebDriver::Error::NoSuchElementError => {
         retry: true,
-        error_code: 12
+        error_code: Maze::Api::ExitCode::AUTOMATION_ELEMENT_NOT_FOUND
```
twometresteve, line 39: You'll notice here that the underlying exit code is the same in two case.  We could certainly have different codes for NoSuchElementError and AppiumElementNotFoundError, but I get the feeling that no one is likely to care at the exit code level - and having different codes to retry on would more likely create extra work/noise.

Cawllec (approved):
LGTM based on our earlier discussion 👍

PR #588, opened 2023-09-27
Improvements to error logging and monitoring

twometresteve:
## Goal

A few improvements to the logging and monitoring of errors:
- Errors logged to the console for invalid requests made clearer
- App version added to the BugSnag integration, allowing us to "fix" errors as appropriate
- Removed a source of double reporting
- Disabled BugSnag for self-tests that force an error condition 

## Tests

Covered by CI and inspection of the BugSnag dashboard.

### DIFF
--- a/lib/maze/servlets/servlet.rb
+++ b/lib/maze/servlets/servlet.rb
@@ -102,9 +102,9 @@ module Maze
         set_response_header response.header
         response.status = post_status_code
       rescue JSON::ParserError => e
-        Bugsnag.notify e
         msg = "Unable to parse request as JSON: #{e.message}"
         if Maze.config.captured_invalid_requests.include? @request_type
+          Bugsnag.notify e
           $logger.error msg
           Server.invalid_requests.add({
             reason: msg,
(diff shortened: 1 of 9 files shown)


lib/maze/servlets/servlet.rb
```diff
@@ -115,9 +115,9 @@ def do_POST(request, response)
           $logger.warn msg
         end
       rescue StandardError => e
-        Bugsnag.notify e
         if Maze.config.captured_invalid_requests.include? @request_type
-          $logger.error "Invalid request: #{e.message}"
+          Bugsnag.notify e
+          $logger.error "Invalid #{@request_type} request: #{e.message}"
```
tomlongridge, line 120: I don't quite follow why this is an error and if it's not in the `captured_invalid_requests` list then it's a warning, but presumably you're intentionally not bugsnagging the latter?

My consistency alarm is sounding though as in the rescue block above you are bugsnagging for the warning case too (L105).

twometresteve: You're right, the error handlers should be consistent with each other - corrected.

There are some circumstance where we don't care if a request we receive is invalid.  For example, a request made to `/command` from a test fixture polling away, or in a test scenario where we don't care about session requests but can't turn them off completely for some reason.  We still log it as a warning though to avoid it being completely invisible to the test scenario - just in case there is a significance that we didn't realise.

tomlongridge (approved):

PR #591, opened 2023-10-10
update attribute_value_matches to handle regex values

gingerbenw:


gingerbenw:
Please see https://github.com/bugsnag/bugsnag-js-performance/pull/348

PR #593, opened 2023-10-12
Allow BitBar browser versions to be set on the command line

twometresteve:
## Goal

Allow BitBar browser versions to be set on the command line. 

## Design

Adds new entries in `bb_browsers.yml` for each browser, which the `version` field missing.  For these browsers, the new `--browser-version` command line option must be provided to set a suitable version.

## Documentation

Command line help updated for the new option.

## Changeset

New command line option with validation, new entries in the browsers file and updated logic to set the capability when needed.

## Tests

Unit tests updated and verified locally that that the `--browser-version` can be used with each of the three browsers.


joshedney (approved):
LGTM

PR #594, opened 2023-10-16
Avoid shell interpretation of characters in file names in Maze::MacosUtils.capture_screen

twometresteve:
## Goal

Fixes an error on macOS when screenshots are taken for scenarios containing characters interpreted by the shell (such as `(`).

## Changeset

I've also taken the opportunity to reuse the code that determines the output folder for the scenario.

## Tests

Tested by temporarily changing one of the e2e tests to contain parenthesis and always take a screenshot.


richardelms (approved):
thanks!

PR #600, opened 2023-10-25
Retry creation of BitBar Appium sessions

twometresteve:
## Goal

Refactor the Appium client so that a different retry strategy can be employed for each device farm.  For BitBar, we will not retry after waiting for different times depending on the error that was experienced.

## Changeset

Relatively light refactor to push `handle_error` into each farm-specific client, allowing them to retry based on different logic.

## Tests

Some manual testing performed, but the retry logic is primarily covered by a new unit test.


Cawllec (approved):
LGTM

PR #602, opened 2023-10-26
Fix crash when logging message for invalid requests

twometresteve:
## Goal

Fixes a crash when logging the message for invalid requests received.

## Tests

Verified by peer review.


joshedney (approved):
LGTM!

PR #608, opened 2023-11-16
Modify retries for BitBar app uploads and appium test failures

Cawllec:
## Goal

This PR comprises two changes, both relating to maze-runner attempting to recover in the event of a failure when running on BitBar.

1. When uploading the test app to bitbar, in the event of a failure the upload will retry up to a default maximum of 5 times.  If all of these attempts fail the app will exit, raising the last received error as the reason (all other errors will be logged and notified).
2. If the appium session errors in the midst of a BitBar test run, instead of attempting to restart the driver maze-runner will now set a specific exit code (200: Appium session failure) and fail.  This should help us in debugging and feeding back on BitBar over time.


twometresteve:
Can you target the `v8` branch with this?

twometresteve (approved):

PR #609, opened 2023-11-17
Work around Event Machine issue on ARM macOS

twometresteve:
## Goal

Works around the following crash on ARM-based macOS machines:
```
bash-3.2$ bundle exec maze-runner --version
Unable to load the EventMachine C extension; To use the pure-ruby reactor, require 'em/pure_ruby'
bundler: failed to load command: maze-runner (/Users/administrator/.rbenv/versions/2.7.7/lib/ruby/gems/2.7.0/bin/maze-runner)
```

## Design

Previously we can manually added this to the installed Gem on our ARM build servers, but we're changing how Gems and managed and so Maze Runner needs to work out of the box.

## Tests

Covered by CI and by running the tool locally on an ARM machine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,12 @@
-# TBD
+# 8.13.0 - 2023/11/17
 
 ## Enhancements
 
 - Modify retry logic for app upload failures and BitBar appium failures [608](https://github.com/bugsnag/maze-runner/pull/608)
+
+## Fixes
+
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 8.13.0 - 2023/11/17
+
+## Fixes
+
+- Work around Event Machine issues on ARM macOS [606](https://github.com/bugsnag/maze-runner/pull/609)
+
```
twometresteve, line 6: I'll be rebasing onto another PR before merging

Cawllec (approved):

PR #611, opened 2023-11-28
Auto-accept alerts on iOS

twometresteve:
## Goal

Auto accept iOS alert this this one:
image

## Tests

Tested locally with both BrowserStack and BitBar, verifying that the alert is dismissed in the video recording.


richardelms (approved):
Thanks!!!

PR #612, opened 2023-12-15
Only output buildkite collapses on buildkite

imjoehaines:


twometresteve (commented):
Good change, I'm surprised it's gone unnoticed for this long.  Just a couple of things:
- Change log, please
- If you want it in v8, then the PR will need to target the v8 branch

twometresteve:
Rolled into #616

PR #613, opened 2023-12-18
Idempotent command endpoints

twometresteve:
## Goal

Adds endpoints to allow commands to be retrieve without changing the internal state of Maze Runner, meaning that they can be called repeatedly.

## Design

Each command that's added to the list now gets assigned a UUID.

Two additions are being made:
- Optional `after` parameter added to the existing `/command` endpoint.  When used, always returns the command following the one with the UUID given (or no-op or error).
- New `/commands` endpoint that always returns all commands that have been added to the list.

## Documentation

Markdown docs added for Commands in general.

## Tests

Unit tests updated and risk of regression covered by existing tests.  For the next functionality I've successfully updated the test fixture for the


lib/maze/servlets/command_servlet.rb
```diff
@@ -8,25 +8,69 @@ module Servlets
...
+            response.status = 200
+            commands.next
+          end
+        else
+          $logger.info "request.query = #{request.query}"
+
+
+          # Idempotent mode
+          after_uuid = request.query['after']
+          if after_uuid.nil?
```
Cawllec, line 39: Is `after` intended to be the _only_ recognised query parameter?  If not, can we rename the `return_command` method to be indicative that it returns the command after the uuid?

twometresteve: Renamed to `command_after(uuid,...`.

PR #618, opened 2024-01-09
Add combined file & STDOUT logging

Cawllec:
## Goal

We want to have a pair of loggers, one with a configurable log level that logs to STDOUT according to the user, and another which always logs everything to a file that can be consulted when errors occur.

Unimplemented features include:
- Splitting log files between:
  - Set up
  - Individual scenarios
  - Tear down
- Outputting the log file to the maze_output directory for automatic uploading to buildkite (perhaps depending on BUILDKITE environment variable?)

## Design

This follows previous split logging usage as seen in the Bugsnag PHP PSR logger - a pseudo "logger" sits at the top level (replacing the previous logger) as the `Maze::Logger` class and `$logger` global variable.  This then captures all incoming method calls and distributes them to the component `STDOUTLogger` and `FileLogger` sub-loggers.

This setup should allow us to configure both actual loggers independently, while passing on shared configuration (such as `datetime_format`) simultaneously.  Both loggers will also be completely accessible via the accessors added to the parent logger.

## Changeset

I've moved all loggers into separate files within a new `loggers` directory, and broken the `LogUtils` class into a separate file.  I've also updated all references pointing (absolutely or relatively) to `maze/logger` to `maze/loggers/logger`.

## Tests

Tests have been minimally updated, but will need to be refactored to properly test the new setup, rather than just expecting both loggers to output identically.


twometresteve:
> Unimplemented features include

Can this section of the PR description be removed now?

lib/maze/loggers/file_logger.rb
```diff
@@ -0,0 +1,30 @@
...
+      File.delete(Maze::FileLogger::LOG_LOCATION) if File.exist?(Maze::FileLogger::LOG_LOCATION)
+
+      super(LOG_LOCATION, level: ::Logger::TRACE)
+
+      @datetime_format = '%H:%M:%S'
+
+      @formatter = proc do |severity, time, _name, message|
+        formatted_time = time.strftime(@datetime_format)
+
+        "\e[2m[#{formatted_time}]\e[0m #{severity.rjust(5)}: #{message}\n"
```
twometresteve, line 26: I don't think we want ANSI colour codes into files, as most editors won't recognise them?

lib/maze/hooks/logger_hooks.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+module Maze
+  module Hooks
+    # Hooks for Browser mode use
+    class LoggerHooks
+      class << self
+        def before(scenario)
+          location = "\"# #{scenario.location}\""
+          $logger.trace "\n--- Begin Scenario: #{scenario.name} #{location}"
```
twometresteve, line 10: Looking at the resulting log file, the being/end aren't as clear as they could be imo.  Instead of a new line at the start here (and none in the end), how about an empty log line before this one and an empty one after the end.  Ie.:
```
$logger.trace ''
$logger.trace "\n--- Begin Scenario: #{scenario.name} #{location}"
```

and 

```
$logger.trace "--- End Scenario: #{scenario.name} #{location}"
$logger.trace ''
```

Just an idea to try and make separate scenarios stand out more.

twometresteve (approved):

PR #627, opened 2024-02-07
Set capabilities for BitBar using Appium v1

twometresteve:
## Goal

Set capabilities for BitBar using Appium v1.

## Design

Appium v1 needs some capabilities set in the top level of the hash.  We thought we had moved on from Appium 1, but recently found that Appium 2 isn't as stable as we hoped yet.  

Capabilities are now set in the root of the hash and are yet ignored by Appium 2.

## Tests

Tested locally that the following scenario doesn't result in a session timeout.  Before this change it would consistently fail after the 65s sleep:
```
  Scenario: Timeout test
    When I wait for 55 seconds
    And I tap the screen 1 times
    And I wait for 65 seconds
    And I tap the screen 1 times
```

Appium 2 use is covered by CI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,12 @@
+# 9.1.1 - TBD
+
+## Fixes
+
+- Set capabilities for BitBar using Appium v1 [625](https://github.com/bugsnag/maze-runner/pull/625)
+
 # 9.1.0 - 2024/01/26
+
+## Enhancements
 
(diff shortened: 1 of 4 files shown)


twometresteve:
> Following discussion, we want to make sure we test this setup works with both appium 1 & appium 2. To this end we should add iOS tests running on BitBar to our CI setup.

Now added and passing.  I've also checked the appium.log on BitBar to check there were no warnings/errors for `newCommandTimeout`.

Cawllec (approved):

PR #628, opened 2024-02-09
Add app hang exit code

Cawllec:
## Goal

For use in pipelines where we may want to automate test restarts when an appium connection is lost.


lib/maze/api/exit_code.rb
```diff
@@ -12,6 +12,8 @@ class ExitCode
       TUNNEL_FAILURE = 101
       SESSION_CREATION_FAILURE = 102
       APPIUM_SESSION_FAILURE = 103
+      # A catch-all for certain errors related to Appium failures when the app is running app hang or ANR tests
+      APPIUM_APP_HANG_ERROR = 104
```
twometresteve, line 16: How will this actually get used - just by the Cocoa repo?  What's the underlying error that occurs that we denote as an "Appium app hang error?"

Cawllec: Hmm, good point on the naming.  I'll re-label it a failure instead.  There isn't a specific error that points to this, but if a test marked as `app_hang` or `anr`  fails there are conditions to fail with this code, so that buildkite can pick it up.  Thinking about it, there's no actual reason it can't be in this repo, so I'll modify this.

Cawllec: Following discussion, the re-name will be done, but the retry logic will be kept specific to the repos, as it won't be universal.

twometresteve (approved):

PR #630, opened 2024-02-15
Update span schemas to include missing verifications

Cawllec:
## Goal

Adds to the span schema validation to cover the following:
- `service.version`, required in the attributes span
- `device.id`, required in the attributes span
- `bugsnag.sampling.p`, required in the attributes of each individual span

In addition this tightens the validation on timestamps, ensuring they are occurring within a 1 hour allowable period around the current time.  This should allow them to pass even when delayed or running in differing time-zones, but will flag errors if the timestamp is orders of magnitude away from the current time.

## Tests

The validator tests have been updated to ensure that the additional functionality works as required on an individual basis.


tomlongridge (approved):
LGTM 👍

tomlongridge (approved):

PR #633, opened 2024-02-28
Apply fix for app_id missing in BrowserStack tests

Cawllec:
## Goal

When running BrowserStack appium tests we've been seeing errors with the test fixture trying to access the config file that should be present on the device.  This is caused by the `app_id` being missing when uploaded the config file, causing it to upload in the wrong location.

It appears the session capabilities which would normally indicate the app_id are now missing in BrowserStack, namely the `CFBundleIdentifier` capability, so a check has been added to see if this is present, and if not will use the `bundleID` capability.

## Releasing

I've added this as a fix, but I'd like it to go out with #632 if possible, so once this is merged I'll rebase that PR and prepare the release in there.

### DIFF
--- a/lib/maze/client/appium/base_client.rb
+++ b/lib/maze/client/appium/base_client.rb
@@ -25,6 +25,8 @@ module Maze
                                  Maze.driver.session_capabilities['CFBundleIdentifier'] # Present on BS and locally
                                end
 
+          pp "App id is set to #{Maze.driver.app_id}"
+
           # Ensure the device is unlocked
           begin
             Maze.driver.unlock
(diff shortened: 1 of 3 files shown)


lib/maze/client/appium/base_client.rb
```diff
@@ -22,7 +22,10 @@ def start_session
                                when 'android'
                                  Maze.driver.session_capabilities['appPackage']
                                when 'ios'
-                                 Maze.driver.session_capabilities['CFBundleIdentifier'] # Present on BS and locally
+                                 unless app_id = Maze.driver.session_capabilities['CFBundleIdentifier']
+                                    app_id = Maze.driver.session_capabilities['bundleID']
+                                 end
+                                 app_id
```
twometresteve, line 28: Wow, nice detective work on this.  I'm guessing we can't just use `bundleID` every time?  I didn't even remember us having this logic in here to start with to know.

twometresteve (approved):

PR #639, opened 2024-03-11
Enable logging captured reflect requests

Cawllec:
## Tests

I've tested locally, but have some ideas about a maze_output test fixture which could be added to verify this and all other post-test outputs if we think it's necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 9.5.0 - TBD
+
+## Enhancements
+
+- Update reflective server to output received requests [639](https://github.com/bugsnag/maze-runner/pull/639)
+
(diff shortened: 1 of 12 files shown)


twometresteve:
The changes so far look good to me.  From memory of when I added the `traces` endpoint, it wasn't too hard to add a basic to (I think) the payload helper tests.  It would be good if we can show that this PR satisfies the whole initial ask without having to make another PR.  The original request was for something like this (not necessarily the precise wording of course):
```
        And I wait to receive 1 reflect request

        Then the reflect request method equals "GET"
        And the reflect request "traceparent" header equals "00-abc123-xyz456-01"
```

twometresteve (approved):
Nice work

PR #641, opened 2024-03-28
Remove unsupported BrowserStack devices

twometresteve:
## Goal

Remove various devices that will no longer be supported by BrowserStack, as notified by email on 22nd March.

## Changeset

I've also removed all the iOS 11 devices as we no longer have nay need for them.

## Tests

Covered by CI.


joshedney (approved):
LGTM!

PR #643, opened 2024-03-28
Allow the command server to provide the next ID when given the last from the previous scenario

Cawllec:
## Goal

Enables the command servlet to respond with the next command in the list when an idempotent request for a command contains the previously sent commands UUID.  This indicates that the app may not have cleared the cache properly, for a number of reasons, mainly because the test fixture was closed suddenly as part of the previous test.

The command servlet will always check to see if there's a matching command for the UUID given _before_ checking against the previously given UUID, so we shouldn't run into any scenario where an incorrect command is given by mistake.

In addition, the previous UUID is stored on the server class as the servlet resets in some way between scenarios.

## Tests

Additional tests will be added before merging (as well as changelog and version updates)

### DIFF
--- a/lib/maze/server.rb
+++ b/lib/maze/server.rb
@@ -15,6 +15,12 @@ module Maze
     DEFAULT_STATUS_CODE = 200
 
     class << self
+
+      # Records the previous command UUID sent to the test fixture
+      #
+      # @return [String] The UUID of the last command sent
+      attr_accessor :last_command_uuid
+
       # Sets the response delay generator.
       #
       # @param generator [Maze::Generator] The new generator
--- a/lib/maze/servlets/command_servlet.rb
+++ b/lib/maze/servlets/command_servlet.rb
@@ -11,8 +11,6 @@ module Maze
 
       NOOP_COMMAND = '{"action": "noop", "message": "No commands queued"}'
 
-      attr_reader :last_uuid
-
       # Serves the next command, if these is one.
       #
       # @param request [HTTPRequest] The incoming GET request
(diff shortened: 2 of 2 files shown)


Cawllec:
I believe this is ready for review now.

test/fixtures/command-workflow/features/commands.feature
```diff
@@ -0,0 +1,77 @@
...
+    And the error payload field "command_response.uuid" equals "2"
+
+  Scenario: An error is sent when no commands are added
+    When I run the "bounce_command" test script
+    And I wait to receive an error
+    Then the error payload field "command_response.message" equals "No commands queued"
+    And the error payload field "command_response.action" equals "noop"
+    And the error payload field "command_status" equals 200
+
+  Scenario: An error is sent when commands run out
```
twometresteve, line 32: nit
```suggestion
  Scenario: A noop is sent when commands run out
```

test/fixtures/command-workflow/features/commands.feature
```diff
@@ -0,0 +1,77 @@
...
+    Then the error payload field "command_response.message" equals "second"
+    And the error payload field "command_response.uuid" does not equal the stored value "uuid"
+    And the error payload field "command_response.run_uuid" equals the stored value "run_uuid"
+
+  Scenario: Idempotent commands are sent correctly
+    Given I generate a series of commands with sequential UUIDs
+    Then I run the "bounce_idempotent_command" test script with UUID "1"
+    And I wait to receive an error
+    Then the error payload field "command_response.message" equals "second"
+    And the error payload field "command_response.uuid" equals "2"
```
twometresteve, line 23: I'm thinking you could discard the oldest error and then repeat all of these lines to prove that we really do get the same thing again for a second call with the same UUID?

PR #646, opened 2024-04-16
Use BitBar device group IDs where possible

Cawllec:
## Goal

Allows us to use a device group ID instead of having to interrogate the bitbar API to attempt to get a device, which should be more reliable and prevent race conditions occurring where multiple of our instances attempt to acquire the same device at the same time.

## Tests

A previous iteration contained extra pipeline steps to test the use of a device group ID, [as can be seen here](https://buildkite.com/bugsnag/maze-runner/builds/3981#018ee76b-8b40-42a7-adbd-62ae0e48e302/201-218)


lib/maze/client/appium/bb_devices.rb
```diff
@@ -18,22 +18,44 @@ def get_available_device(device_or_group_names)
...
+            if device
+              device_name = device['displayName']
+              platform = device['platform'].downcase
+              platform_version = device['softwareVersion']['releaseVersion']
 
-            $logger.info "Selected device: #{device_name} (#{platform} #{platform_version})"
+              $logger.info "Selected device: #{device_name} (#{platform} #{platform_version})"
+            else
+              # If a device group has been identified, extrapolate characteristics from the group name
+              if android_match = Regexp.new('(ANDROID|android)_(\d{1,2})').match(device_or_group_names)
```
Cawllec, line 49: I'm not happy with this, as I think it's too inflexible in the event we added a different device group names, however it should be fine for how we access devices at the moment.

twometresteve:
There's a curious effect of using the `deviceGroupId` capability in the BitBar dashboard - all devices in the device group are displayed, even if they were not selected for use in the test.  I wonder if that's what the BitBar team expect and perhaps it's worth raising a support ticket to ask?

twometresteve:
The down side of using deviceGroupId is that the Maze Runner output no longer includes the device model that was actually selected.  Is it possible to get that from what's returned when the session is created? If not then perhaps it's a change we can request?

twometresteve:
Everything I've tested works, including use of multiple device groups, e.g .`ANRDOID_13|ANDROID_14`.  We might just want to see if we can get and log the actual device that was used in the run when `deviceGroupId` before committing ourselves to the change.  Perhaps we can look up the device using the BitBar API at the end of the run, when we work out the dashboard link?

I wouldn't add it to this PR - raise a ticket to add it separate. I just think we should assess whether we can do it or not before merging.

twometresteve (approved):

PR #648, opened 2024-04-19
Only update documentation if the release is the most recent

Cawllec:
## Goal

Adds an in-process check to see if a given release matches (or exceeds) the release marked as 'latest' on github.  We'll only attempt to build and update the docs if that's the case.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -2,6 +2,8 @@ require 'bundler/gem_tasks'
 require 'rake/testtask'
 require 'yard'
 require 'open3'
+require 'net/http'
+require 'json'
 
 def run_command(command)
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -19,6 +21,28 @@ def run_command(command)
   end
 end
 
+def parse_release_tag(tag)
+  regex = /^v?(\d+)\.(\d+)\.(\d+).*$/
```
twometresteve, line 25: Might be an idea to allow an upper case V at the start.  I have made releases with one by mistake in the past.

twometresteve (approved):
Cunning approach, just the one minor comment.

PR #651, opened 2024-05-03
Add span step for checking properties

Cawllec:
## Goal

Allows us to create a table of properties to be verified in a specifically named span, in the form:

```
Then a span named {name} has the following properties:
            | property                       | value                          |
            | path.to.property          | here i am                   |
            | spanId                          | 7af51275a21aa300   |
            | attributes.0.key           | bugsnag.sampling.p |
```

Note, it verifies these by converting the properties into strings, as the table values will also be strings.  If this is a potential issue (i.e. numbers need to be within a tolerance) we can look at changing the behaviour.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ Pods
 *.gem
 coverage/
 maze_output/
+maze-runner.log
 Gemfile.lock
 package-lock.json
 test/fixtures/init-test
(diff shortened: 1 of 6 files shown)


twometresteve (approved):
LGTM.  I performed a little exact testing for the failure cases too, for example by having multiple spans with the same name that match the criteria between them, but none satisfied them all by themselves.

djskinner (approved):

PR #652, opened 2024-05-07
Update AWS lambda support for more recent SAM CLI releases

Cawllec:
## Goal

This updates the AWS sam functionality to work with a newer version of SAM CLI, where test runs don't quite work in the same way - instead of _always_ finishing cleanly regardless of the state of execution, whereas the newer version fails with an error.  This change primarily turns our earlier failure conditions into a warning, which will increase noise and shift the onus onto the reader to figure out that the test is failing due to an error in the test fixture, but I can't see another way around this at the time.

I've also moved the tests on MacOS 14, rather than Github Actions, as the name of this branch and some commits may reflect.


joshedney (approved):
LGTM!

twometresteve (approved):

PR #654, opened 2024-05-30
Use selective test runs to reduce build time

Cawllec:
## Goal

This will be updated when we're ready to review


Cawllec:
This is no longer going to be merged, closing.

PR #655, opened 2024-06-21
Improve handling of Selenium `ServerError`s

twometresteve:
## Goal

Improve the handling of Selenium `ServerError`s (also applicable to Appium tests.

## Design

Our monkey patch was obfuscating the interesting information held in the error, so when reported to our BugSnag dashboard we just saw lots of "Server Error 500" errors all grouped together, each with potentially very different underlying causes.

This change does two things:
- Ensures the error that is sent to BugSnag has a meaningful messing
- Allows the error handler that rescues the errors to decide if it should be logged to the console or not.

Before examples:
https://buildkite.com/bugsnag/bugsnag-android/builds/11506#019038f5-fdb1-4090-a3b3-e3fb0f442e6c/273-299
https://app.bugsnag.com/bugsnag/maze-runner/errors/661363735a3c130008080b72?filters[error.status]=open&filters[event.since]=30d

After examples:
https://buildkite.com/bugsnag/bugsnag-android/builds/11507#01903b6b-25e4-4d82-92ce-c64fab22aadb/206-230
https://app.bugsnag.com/bugsnag/maze-runner/errors/66759d7eaf1a6b00086daab5?filters[error.status]=open&filters[event.since]=30d

## Changeset

An unrelated change is pinning the `ffi` Gem to v1.6.3 in the Ruby 2 container.  We should probably explore this more and assess if we still need a Ruby 2 container at all anymore.

## Tests

We may want to make further tweaks to error handlers catching `ServerError`s to log more, but for now the BugSnag dashboard is a lot more useful and the console output is just as useful because the error's message was output anyway.


joshedney (approved):
LGTM!

PR #659, opened 2024-07-03
[PLAT-12352] Disable timestamp validation on spans

joshedney:
## Goal

Allow users to disable timespan validation using the step:

`When I disable timestamp validation for spans` 

## Tests

Covered by CI


lib/maze/schemas/trace_validator.rb
```diff
@@ -53,22 +53,24 @@ def validate
...
-        unless parsed_timestamp > 0
-          @success = false
-          @errors << "Timestamp was expected to be a positive integer, was '#{parsed_timestamp}'"
-          return
-        end
-        time_in_nanos = Time.now.to_i * 1000000000
-        unless (time_in_nanos - parsed_timestamp).abs < tolerance
-          @success = false
-          @errors << "Timestamp was expected to be within #{tolerance} nanoseconds of the current time (#{time_in_nanos}), was '#{parsed_timestamp}'"
+        if Maze.config.span_timestamp_validation != false
```
twometresteve, line 56:
```suggestion
        if Maze.config.span_timestamp_validation
```

lib/maze/configuration.rb
```diff
@@ -75,6 +75,9 @@ def initialize
     # Enables awareness of a public IP address on Buildkite with the Elastic CI Stack for AWS.
     attr_accessor :aws_public_ip
 
+    # Disables timestamp validation on spans
+    attr_accessor :span_timestamp_validation
```
twometresteve, line 79: We'll need to make sure this is set to `true` in one of the `Before` hooks, so it doesn't bleed over into subsequent scenarios.

lib/maze/configuration.rb
```diff
@@ -75,6 +75,9 @@ def initialize
     # Enables awareness of a public IP address on Buildkite with the Elastic CI Stack for AWS.
     attr_accessor :aws_public_ip
 
+    # Disables timestamp validation on spans
```
twometresteve, line 78: Wording of the comment is potentially misleading, to me it makes it sound like setting it to true will disable it. 
```suggestion
    # Whether timestamp validation should be performed on spans
```

twometresteve (commented):
Also remember the CHANGELOG and as discussed IRL we'll need some form of test(s) for it.

test/schemas/trace_validation_test.rb
```diff
@@ -278,7 +278,19 @@ def test_greater_than_failure_lesser
...
+      'timestamp' => '12345000000000'
+    }))
+    Time.expects(:now).never
+    validator.validate_timestamp('timestamp', Maze::Schemas::HOUR_TOLERANCE)
+    assert_nil(validator.success)
+    assert(validator.errors.empty?)
+  end
+
   def test_validate_timestamp_success_exact
+    Maze.config.span_timestamp_validation = true
```
twometresteve, line 293: I think we can avoid having this in every unit test by setting it to true in the `initialize` method of `Maze::configuration`.  Then I think it will be more consistent with things like `enforce_bugsnag_integrity`.

lib/maze/schemas/trace_validator.rb
```diff
@@ -53,22 +53,24 @@ def validate
...
-        unless parsed_timestamp > 0
-          @success = false
-          @errors << "Timestamp was expected to be a positive integer, was '#{parsed_timestamp}'"
-          return
-        end
-        time_in_nanos = Time.now.to_i * 1000000000
-        unless (time_in_nanos - parsed_timestamp).abs < tolerance
-          @success = false
-          @errors << "Timestamp was expected to be within #{tolerance} nanoseconds of the current time (#{time_in_nanos}), was '#{parsed_timestamp}'"
+        if Maze.config.span_timestamp_validation
```
twometresteve, line 56: Sorry not to mention this in the initial review, but you could use `return unless Maze.config.span_timestamp_validation` on the first line to avoid indenting the whole method.

twometresteve (approved):

PR #666, opened 2024-07-26
Captures server-side error and removes appium driver

Cawllec:
This should prevent multiple instances of server side errors appearing in the dashboard.  Afterwards any attempt to access the appium driver should fail without delivering a report, and the session will close.


twometresteve (approved):
Looks sensible to me, if you could just add a CHANGELOG entry for it.

PR #669, opened 2024-08-13
Add support for unmanaged traces

twometresteve:
## Goal

Add support for unmanaged traces.

## Design

Adds a new configuration property and step to enter unmanaged traces mode, which disables checking of the `bugsnag-span-sampling` header.

## Tests

e2e test updated.


joshedney (approved):
LGTM!

PR #670, opened 2024-08-29
Use API key as subdomain when repeating trace requests

twometresteve:
## Goal

Use the API key as the subdomain in the URKL when repeating trace requests to Bugsnag.

## Design

This approach reflects a change in system design, where the performance libraries themselves will be taking the same approach.

## Tests

Tested locally by running the test suite, verifying that trace data appears in the dashboard and no relevant Bugsnags are generated.  Two Bugsnags were generated due to the `Bugsnag-Span-Sampling` header missing, but that's because we have a test scenario for unmanaged traces mode.


Cawllec (approved):
LGTM from a functional perspective

PR #671, opened 2024-08-30
Modify validation framework for custom use

Cawllec:
## Goal

Aims to rework our existing validation setup, which currently only includes trace validation, to be extendable through configuration options, allowing us to impose custom validation rules in each test environment as required.

## Changes

This PR primarily refactors the trace validation class to be more generic, building from a `ValidatorBase` class with the generic methods attached.  This base class will be used as a jumping off point for any future validators.

## Tests

The unit tests have been updated, and existing tests for the trace validation logic remains in place.


Cawllec:
Closed in favour of #672

PR #674, opened 2024-09-19
[PLAT-12770] Override `--fail-fast` via an environment variable

joshedney:
## Goal

Add the ability to override the `--fail-fast` option via an environment variable `MAZE_NO_FAIL_FAST`

## Changeset

Check if we're passing the `--fail-fast` option and that we've set the environment variable `MAZE_NO_FAIL_FAST` and remove the `--fail-fast` option before we pass it to cucumber.

## Tests

Covered by CI

### DIFF
--- a/bin/maze-runner
+++ b/bin/maze-runner
@@ -188,6 +188,12 @@ class MazeRunnerEntry
     remove_maze_runner_args
     add_cucumber_args
 
+    # Check if we've set ENV['MAZE_NO_FAIL_FAST'] to override the fail fast behaviour
+    # And remove the --fail-fast option if it's set
+    if ENV['MAZE_NO_FAIL_FAST'] && @args.include?('--fail-fast')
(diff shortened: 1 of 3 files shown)


bin/maze-runner
```diff
@@ -132,6 +132,12 @@ class MazeRunnerEntry
       @args << "--expand"
     end
 
+    # Check if we've set ENV['MAZE_NO_FAIL_FAST'] to override the fail fast behaviour
+    # And remove the --fail-fast option if it's set
+    if ENV['MAZE_NO_FAIL_FAST'] && @args.include?('--fail-fast')
+      @args = @args - ['--fail-fast']
```
twometresteve, line 138: You could perhaps add this here:
```
$logger.info 'Suppressing --fail-fast option as MAZE_NO_FAIL_FAST is set.'
```

PR #675, opened 2024-09-19
Add extra validation steps and unit tests

Cawllec:
## Goal

Adds some extra functions to shortcut a bunch of more complicated logic that had to be used in the [Go tests](https://github.com/bugsnag/bugsnag-go/pull/248).

### DIFF
--- a/lib/maze/schemas/validator_base.rb
+++ b/lib/maze/schemas/validator_base.rb
@@ -45,9 +45,6 @@ module Maze
         if element.nil?
           @success = false
           @errors << "Element '#{path}' was not found"
-        elsif element.kind_of?(Array) && element.empty?
-          @success = false
-          @errors << "Element '#{path}' was an empty array"
         end
       end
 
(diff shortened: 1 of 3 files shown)


lib/maze/schemas/validator_base.rb
```diff
@@ -32,6 +32,58 @@ def validate
...
+      end
+
+      def element_exists(path)
+        element = Maze::Helper.read_key_path(@body, path)
+        if element.nil?
+          @success = false
+          @errors << "Element '#{path}' was not found"
+        elsif element.kind_of?(Array) && element.empty?
+          @success = false
+          @errors << "Element '#{path}' was an empty array"
```
twometresteve, line 50: I _think_ this makes sense - we always consider empty arrays as "missing"?  I can imagine that quirks in various platforms could cause an empty array to be serialised from "nothing".

twometresteve (approved):
LGTM, seems clean and well thought out.

lib/maze/schemas/validator_base.rb
Cawllec: I've removed this, and have set it to allow arrays to be marked as present.  Otherwise I can see this potentially being a tripping point as it's not obvious what's happening.

PR #677, opened 2024-09-20
Modify default setup for how validation works

Cawllec:
## Goal

Accomplishes 3 things:
- Changes how validation will be run by default, ensuring a given config block won't prevent the default validators being run
- Adds a new configuration option that will disable the default validation if called, but won't disable any custom configured validators:
```
Maze.config.skip_default_validation('errors')
```
- Adds a new configuration option that denotes that the test is in client mode, and indicates different kinds of verification could be used.  This is intended to be used to modify the internal validators in the case where certain attributes are not present in different notifier circumstances.

## Tests

Not implemented yet...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# 9.14.0 - TBD
+# 9.14.0 - 2024-09-23
 
 ## Enhancements
 
@@ -6,6 +6,8 @@
 - Add support for defining custom validation blocks in configuration
   - [672](https://github.com/bugsnag/maze-runner/pull/672)
   - [675](https://github.com/bugsnag/maze-runner/pull/675)
+  - [677](https://github.com/bugsnag/maze-runner/pull/677)
+- Add client mode configuration option [677](https://github.com/bugsnag/maze-runner/pull/677)
 - Add support for command line options files [676](https://github.com/bugsnag/maze-runner/pull/676)
 
 # 9.13.1 - 2024-08-30
--- a/lib/maze/configuration.rb
+++ b/lib/maze/configuration.rb
@@ -96,6 +96,20 @@ module Maze
       @custom_validators[endpoint] = validator
     end
 
+    # Whether default validation should be skipped for a given endpoint
+    attr_reader :skipped_validators
+
+    # Sets whether to skip default validation for a given endpoint
+    #
+    # @param endpoint [String] The endpoint to skip default validation for
+    def skip_default_validation(endpoint)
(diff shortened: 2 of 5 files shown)


Cawllec:
@twometresteve I'm not sure if the client mode config option makes sense at a glance?  I can't think of any other name for it though

lib/maze/configuration.rb
```diff
@@ -96,6 +96,20 @@ def add_validator(endpoint, &validator)
       @custom_validators[endpoint] = validator
     end
 
+    # Whether default validation should be skipped for a given endpoint
+    attr_reader :skipped_validators
+
+    # Sets whether to skip default validation for a given endpoint
+    #
+    # @param endpoint [String] The endpoint to skip default validation for
+    def skip_default_validation(endpoint)
```
twometresteve, line 105: Passing the endpoint as a string is kind of brittle - is it "error" or "errors", for example.  That's not the end of the world as it will only mean false failures not false passes, I just wonder if we should create ticket to introduce an enum type for the endpoints and apply it across the codebase (not part of this PR).

lib/maze/schemas/trace_validator.rb
```diff
@@ -34,6 +33,10 @@ def validate
           'resourceSpans.0.scopeSpans.0.spans.0.endTimeUnixNano',
           'resourceSpans.0.scopeSpans.0.spans.0.startTimeUnixNano'
         )
+
+        unless Maze.config.client_mode_validation
```
twometresteve, line 37: Other way around - it should be present for client mode
```suggestion
        if Maze.config.client_mode_validation
```

lib/maze/configuration.rb
```diff
@@ -96,6 +96,20 @@ def add_validator(endpoint, &validator)
...
+      @skipped_validators ||= {}
+      @skipped_validators[endpoint] = true
+    end
+
+    # Sets whether validation should be run in client mode
+    attr_accessor :client_mode_validation
```
twometresteve, line 111: Naming is fine to me, but I would default it to true in the initialiser.

CHANGELOG.md
```diff
@@ -1,11 +1,13 @@
-# 9.14.0 - TBD
+# 9.14.0 - 2024-09-23
```
twometresteve, line 1: Looks like `lib/maze.rb`  needs updating with the new release version too.

twometresteve (approved):
Happy to merge and release once the VERSION is bumped.

PR #684, opened 2024-10-17
Add outline of steps to get events from data-access-api

Cawllec:
## Goal
Adds a couple of steps to wrap getting the last sent event from the pipeline via data-access-api.

Note: NON-FUNCTIONAL, DO NOT MERGE until error ID acquisition is implemented

### DIFF
--- a/lib/maze/server.rb
+++ b/lib/maze/server.rb
@@ -101,6 +101,8 @@ module Maze
           sourcemaps
         when 'reflect', 'reflects', 'reflection', 'reflections'
           reflections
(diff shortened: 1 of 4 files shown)


lib/maze/server.rb
```diff
@@ -101,6 +101,8 @@ def list_for(type)
           sourcemaps
         when 'reflect', 'reflects', 'reflection', 'reflections'
           reflections
+        when 'pipeline event', 'pipeline events, pipeline_events'
```
sazap10, line 104:
```suggestion
        when 'pipeline event', 'pipeline events', 'pipeline_events'
```

PR #694, opened 2024-10-29
Add extra guard when terminating appium session app

Cawllec:
## Goal

When calling terminate_app we sometimes get the sessionId error.  We can safely ignore it, and prevent it calling into our testing dashboard.

## Tests

Should be covered by existing tests


twometresteve (approved):
LGTM, just needs rebasing and version bumping

PR #695, opened 2024-10-31
Add basic https support

twometresteve:
## Goal

Adds support for running the mock endpoints and document server with https on the main port (9339 by default).

## Design

Support for https on the document server running on 9340 has not been added.  Now that we can use host:94339/docs, the original document server on 9340 can be removed in the next major version.

## Documentation

Command line `--help` naturally updated`

## Changeset

Added the `--https` command line option, which will start the main mock server with SSL and self-signed certificate.  Updated Selenium capability to ignore invalid certificates, allowing the updated tests to work.

I've also bumped the bugsnag-cocoa submodule and changed the pipeline to only build the iOS test fixture, not the macOS one as well.

## Tests

Tests updated to now be covered by CI.


joshedney (approved):
LGTM!

PR #699, opened 2024-11-06
Ignore system-generated ANRs

twometresteve:
## Goal

Ignore system-generated ANRs, logging an INFO line and pushing them to a separate `RequestList`.

## Design

Android 14 and 15 can cause ANRs to be generated if the "This app keeps stopping" dialog is not dismissed within 5 seconds.  Such ANRs are not related to the Busnag notifier and we cannot control them.  This change ignores such ANRs (identifiable by key frames of the stack frame) to avoid flaking tests.

As a future change we might consider making this mechanism more generic and configurable on a repo-by-repo basis.

## Changeset

- Added new internal `RequestList` for storing and later logging out ignored requests
- Added logic for ignoring the ANRs described.

## Tests

Ran a the entire bugsnag-android test suite locally to guard against regression and reran a scenario that was flaking on Android 14 several times to ensure the ANR is ignored.


lemnik (approved):
LGTM

Cawllec (approved):

PR #707, opened 2024-11-26
[PLAT-13243fix: correct case for setting bundleId

joshedney:
## Goal

Correct the name of the `bundleId` pulled from `Maze.driver.session_capabilities` 

Add a check to ensure that it is set and error if it is not.

## Tests

Covered by CI


lib/maze/client/appium/base_client.rb
```diff
@@ -23,11 +23,16 @@ def start_session
...
                                  unless app_id = Maze.driver.session_capabilities['CFBundleIdentifier']
-                                    app_id = Maze.driver.session_capabilities['bundleID']
+                                    app_id = Maze.driver.session_capabilities['bundleId']
                                  end
                                  app_id
                                end
 
+          if Maze.driver.app_id.nil?
+            $logger.error "Failed to determine app id."
+            $logger.debug "Capabilities: #{Maze.driver.session_capabilities.inspect}"
```
joshedney, line 33: @twometresteve not sure if we want this, I thought that outputting what capabilities are set when we've got debug enabled might help 🤷🏼

twometresteve: Logging as debug seems like a good idea to me, but I would just change it as follows to avoid confusion with the capabilities that we set.
```suggestion
            $logger.debug "session_capabilities: #{Maze.driver.session_capabilities.inspect}"
```

lib/maze/client/appium/base_client.rb
```diff
@@ -23,11 +23,16 @@ def start_session
                                  Maze.driver.session_capabilities['appPackage']
                                when 'ios'
                                  unless app_id = Maze.driver.session_capabilities['CFBundleIdentifier']
-                                    app_id = Maze.driver.session_capabilities['bundleID']
+                                    app_id = Maze.driver.session_capabilities['bundleId']
```
twometresteve, line 26: I understand we started seeing the issue with Appium 2.5 - does this fix work before and after 2.5 on BB and also on BS?  I don't think CI covers that.

twometresteve (commented):
Remember a CHANGELOG entry and to bump the version number.

lib/maze/client/appium/base_client.rb
joshedney: I've checked BB and can confirm that it works for Appium versions lower than 2.5.

joshedney: I've also checked BS and its working on Appium 1.21.0

lib/maze.rb
```diff
@@ -7,7 +7,7 @@
 # Glues the various parts of MazeRunner together that need to be accessed globally,
 # providing an alternative to the proliferation of global variables or singletons.
 module Maze
-  VERSION = '9.20.0'
+  VERSION = '9.21.0'
```
twometresteve, line 10: Should be a patch version bump for a fix.
```suggestion
  VERSION = '9.20.1'
```

twometresteve (approved):
Looks good once the version numbers are corrected.

PR #708, opened 2024-12-16
Skip remaining scenarios if the Appium session fails

twometresteve:
## Goal

Skip remaining scenarios if the Appium session fails.

## Design

Previously we were nilling out `Maze.driver` when it failed and guarding any uses of it with the `&.`.  That was generally ok, but the problem with `&.` is that operations essentially fail silently.  This change introduces an explicit failed state, which we can build more rigour on top of.

## Tests

This change can be seen in action on this trial build:
https://buildkite.com/bugsnag/bugsnag-unity/builds/5355#0193cf49-9bcd-4d1a-adad-2d719f96ea89

Although there are many errors in amongst the skipped scenarios, they are due to the failed driver still attempting to be used.  I've raised PLAT-13337 to introduce a formal API for accessing driver methods instead of calling it directly.

### DIFF
--- a/lib/maze.rb
+++ b/lib/maze.rb
@@ -11,20 +11,11 @@ module Maze
   VERSION = '9.22.0'
 
   class << self
-    attr_accessor :check, :internal_hooks, :mode, :start_time, :dynamic_retry, :public_address,
+    attr_accessor :check, :driver, :internal_hooks, :mode, :start_time, :dynamic_retry, :public_address,
                   :public_document_server_address, :run_uuid, :scenario
 
     def config
       @config ||= Maze::Configuration.new
-    end
-
-    def driver
-      raise 'Cannot use a failed driver' if @driver&.failed?
-      @driver
-    end
-
-    def driver=(driver)
-      @driver = driver
     end
 
     def hooks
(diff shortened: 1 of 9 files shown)


lib/maze.rb
```diff
@@ -8,16 +8,25 @@
...
-    attr_accessor :check, :driver, :internal_hooks, :mode, :start_time, :dynamic_retry, :public_address,
+    attr_accessor :check, :internal_hooks, :mode, :start_time, :dynamic_retry, :public_address,
                   :public_document_server_address, :run_uuid, :scenario
 
     def config
       @config ||= Maze::Configuration.new
     end
 
+    def driver
+      raise 'Cannot use a failed driver' if @driver&.failed?
```
Cawllec, line 22: Will this raise cause each attempt to acquire the driver just raise an error when called, but not actually return anything for logical checks?

Cawllec: So trying to call `Maze.driver.failed?` will simply raise the error and not actually return the driver, so it won't trigger checks like `if Maze.driver.failed?`.

Could we simply log instead, but allow the driver to be returned and checked?

Cawllec (approved):

PR #709, opened 2024-12-20
Add failed driver guard to FileManager Appium API

twometresteve:
## Goal

Add failed driver guard to FileManager Appium API.

## Design

First of a number of PRs to shift logic away from `Maze::Appium::Driver` and protect clients from noise caused by a failed underlying Appium driver.

## Tests

Unit test added and some local testing performed, but further testing will be performed on the integration branch prior to release.

### DIFF
--- a/lib/maze/api/appium/file_manager.rb
+++ b/lib/maze/api/appium/file_manager.rb
@@ -31,6 +31,10 @@ module Maze
         rescue Selenium::WebDriver::Error::UnknownError => e
           $logger.error "Error writing file to device: #{e.message}"
           false
+        rescue Selenium::WebDriver::Error::ServerError => e
+          # Assume the remote appium session has stopped, so crash out of the session
+          fail_driver
+          raise e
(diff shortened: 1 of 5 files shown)


lib/maze/api/appium/file_manager.rb
```diff
@@ -45,7 +61,15 @@ def read_app_file(filename, directory = nil)
           end
 
           $logger.trace "Attempting to read file from '#{path}'"
-          file = @driver.pull_file(path)
+          @driver.pull_file(path)
+          true
```
Cawllec, line 65: Since this is attempting to retrieve a file this should be the file pulled rather than the success metric?

twometresteve: Gosh, you're absolutely right - I'd assumed the file is written straight to file, but it is returned in memory by the method call.  Good spot.

Cawllec (approved):

PR #711, opened 2024-12-31
[PLAT-13351] Update BrowserStack Session URL

joshedney:
## Goal

Update the URL used for sessions for BrowserStack.

I've also had to pin the version of the ruby 3 we're using for the docker image to 3.3.3 due to an issue with Cucumber and Ruby 3.4.0

`<internal:hash>:37:in 'initialize': unknown keywords: :strict, :proc (ArgumentError)` 

## Tests

Covered by CI


Cawllec (approved):
Looks good, can we add a changelog entry with the ruby version setting please, so we can track it in future?

lib/maze/client/appium/bs_client.rb
```diff
@@ -67,7 +67,7 @@ def device_capabilities
 
         def log_run_intro
           # Log a link to the BrowserStack session search dashboard
-          url = "https://app-automate.browserstack.com/dashboard/v2/search?query=#{Maze.run_uuid}&type=builds"
+          url = "https://automate.browserstack.com/projects/#{ENV['BUILDKITE_PIPELINE_NAME']}/builds/#{Maze.run_uuid}/1?tab=tests"
```
twometresteve, line 70: This won't work when running locally as `BUILDKITE_PIPELINE_NAME` won't be set.  What you could do though is make use of the existing method to get the project, something like `project_name_capabilities[:project]`. 

It would make sense to use that for the build as well, to future proof it and stop the link getting out of sync with what we actually set.

twometresteve:
> Looks good, can we add a changelog entry with the ruby version setting please, so we can track it in future?

+1 - changelog entry needed as well as a version bump.

lib/maze/client/appium/bs_client.rb
joshedney: Ah yes, I forget about MR running locally.

CHANGELOG.md
```diff
@@ -1,8 +1,10 @@
-# 9.22.0 - 2024/12/xx
+# 9.22.0 - 2025/01/xx
 
 ## Enhancements
 
 - Skip remaining tests after Appium session failure [708](https://github.com/bugsnag/maze-runner/pull/708)
+- Failed driver guards added to FileManager API [709](https://github.com/bugsnag/maze-runner/pull/709)
```
twometresteve, line 6: I'm not sure how this line crept in, but it should just be removed.

lib/maze/client/appium/bs_client.rb
```diff
@@ -67,7 +67,7 @@ def device_capabilities
 
         def log_run_intro
           # Log a link to the BrowserStack session search dashboard
-          url = "https://app-automate.browserstack.com/dashboard/v2/search?query=#{Maze.run_uuid}&type=builds"
+          url = "https://automate.browserstack.com/projects/#{project_name_capabilities[:project]}/builds/#{Maze.run_uuid}/1?tab=tests"
```
twometresteve, line 70: Look like a copy paste error? app-automate has changed to automate.
```suggestion
          url = "https://app-automate.browserstack.com/projects/#{project_name_capabilities[:project]}/builds/#{Maze.run_uuid}/1?tab=tests"
```

twometresteve (approved):

twometresteve (approved):

PR #712, opened 2025-01-07
Allow multiple browsers to be specified

Cawllec:
## Goal

Allows multiple browsers to be specified when running against a browser farm.  This will utilise them in the order given until either the driver successfully boots or we run out of connection attempts.

The main bulk of the changes consist of refactoring the selenium driver start process, which should now function by:

- start_session is called on the appropriate BB or BS driver
  - Local tunnels and initial capabilities are created according to the drivers needs
- start_driver is called, which sits in the base client
  - start_driver attempts to start the driver with the existing attributes
  - If this fails, the error is handled by the base client to determine if it is recoverable (it's assumed they all are currently)
  - If multiple targets exist in the config browser_list the next one will be cycled in and capabilities repopulated by the create_capabilities methods
  - Repeat the above until we succeed or fail more than the allowed amount of times.

## Tests

I've fixed up the unit tests, and extended the test pipeline to show it runs everything, but need to consider how to actually test it in anger given the browser tests are unlikely to randomly and identifiably run the code on their own.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -283,6 +283,7 @@ steps:
         command:
           - "--farm=bs"
           - "--browser=firefox_latest"
+          - "--browser=chrome_latest"
           - "--fail-fast"
     concurrency: 2
     concurrency_group: 'browserstack'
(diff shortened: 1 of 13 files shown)


twometresteve (approved):
Aside from the one comment, looks good to me.  I tested it against BB and BS with single and multiple browsers, hacking in an explicit `raise` for the first option to verify how it behaved.

twometresteve:
Just realised we need a changelog entry and version bump too.

PR #719, opened 2025-02-13
Initial move of trace logic into maze-runner and tests

Cawllec:
## Goal

The overarching goals of the step refactor are to:
- De-couple our testing logic from cucumber steps, allowing us to theoretically use maze-runner as a resource for different testing frameworks in the future
- Combine and reduce the code where we have duplicate steps or methods accomplishing the same thing
- Add tests to our code that accesses the tested payloads

This start of the refactor attempts to establish three places where steps will be refactored into:
- Accessors (name pending) - Code within maze-runner that accesses the payloads and other received information, processes it (better name - processors?) into either a set of values for assertions, or returns a boolean value if the process was a search or similar.  Note - These don't make assertions themselves
- Interfaces (name pending) - Code to consolidate a testing intent described in the scenario lines and the test responses via the accessors, making an assertion on whether that specific test is acceptable.
- Steps - The older gherkin steps that previously had more code involved, these are being refactor to exclusively reference interface methods, in an effort to remove meta-steps, and to move as much logic into the testable ruby code in the interfaces as possible.

Currently these changes mostly target the traces, rather than the span steps.  I'm not entirely happy with the separation of code at the moment, and have been trying to rethink the approach around how much logic is in the interfaces vs the accessors, but feel like the above statements are the best method for the moment, even if I don't love how I've implemented it.

## Tests

I've only added unit tests for the traces accessors at the moment.  I would expected in the integration branch to further add unit tests around the individual interface methods, to ensure they work correctly at that level.

## Going forward

I feel like this is a decent start in the process, but as I've gone through I've found some of the smaller implementation details haven't been as simple to translate or refactor as I initially thought.  I'd like this PR to become the base of the integration branch as it indicates direction more than it represents a final product.


lib/maze/accessors/trace_accessors.rb
```diff
@@ -0,0 +1,106 @@
...
+        # Traces
+
+        # Retrieves a value within a trace
+        #
+        # @param list [String] The name of the list to get spans from
+        # @param field [String] The field to check the attributes of
+        # @param attribute [String] The attribute to check
+        #
+        # @returns [String|Integer] A string or integer representation of the attribute
+        def attribute_by_key(list, field, attribute)
```
twometresteve, line 17: Just to demonstrate the idea of what we discussed IRL earlier (which I think you fully grasped anyway), this would be a good example for a method to sit on a `Trace` class.

twometresteve (approved):
Looks good and I can see the ideas are continuing to evolve.  Certainly going in the right direction.

PR #722, opened 2025-02-17
Move AWS tests from MacOS boxes to github actions

Cawllec:
## Goal

This will hopefully be slightly more reliable than our existing test setup with docker on the macs, as it'll be a new environment every time.

This will also serve as a pattern to be used on other repos testing using AWS lambda.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -8,8 +8,6 @@ steps:
 
   - label: 'AWS - SAM tests'
     timeout_in_minutes: 30
-    agents:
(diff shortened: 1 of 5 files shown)


Cawllec:
I've updated this branch with a fix for an issue we were experiencing with invalid responses from the lambda function, where the expected parsable output line wasn't where it was supposed to be.

You can see this tested here: https://github.com/bugsnag/maze-runner/actions/runs/13388375041/job/37390203890#step:6:174

twometresteve (approved):
Nice work!

PR #725, opened 2025-02-20
Ensure an empty string in the repeater key options is interpreted as null

Cawllec:
## Goal

This should fix the issue where empty environment variables and empty string environment variables may cause maze to crash.

## Tests

Added unit tests, @yousif-bugsnag it would be good if you could repeat your tests with this branch/build as a base for maze to ensure it works fully before we release it.


twometresteve (approved):
LGTM.

twometresteve (approved):

yousif-bugsnag (approved):

PR #726, opened 2025-02-25
Log BitBar device name used in Appium sessions

twometresteve:
## Goal

Log the device model used in BitBar Appium tests.

## Tests

Covered by CI, by inspection of both of the device and browser tests against BitBar.


joshedney (approved):
LGMT

PR #727, opened 2025-03-04
Add the `breakpad-symbol` endpoint to the mock server

joshedney:
## Goal

Add the `/breakpad-symbol` endpoint to the mock server so that we can test the breakpad-symbol uploads via the CLI

## Tests

Covered by CI


Cawllec:
Does this have to be a separate endpoint?  It seems to be functionally identical to the sourcemaps uploads.

joshedney:
> Does this have to be a separate endpoint? It seems to be functionally identical to the sourcemaps uploads.

Yes it does. In the CLI we hardcode the endpoint paths (/dart /dsym /react-native-source-map) and are only able to override the endpoint URL.

Cawllec (approved):

twometresteve (approved):

PR #728, opened 2025-03-04
Reporting of Appium Session statuses to Bugsnag

twometresteve:


lib/maze/api/appium/manager.rb
```diff
@@ -11,8 +11,8 @@ def failed_driver?
           @driver.failed?
         end
 
-        def fail_driver
-          @driver.fail_driver
+        def fail_driver(message)
+          @driver.fail_driver(message)
```
Cawllec, line 15: Would this work if we removed the `message` argument, and instead pulled up the last error to occur `$!` and pulled the message from that?  There are potential risks involved, with things like race conditions possible.

Cawllec: Mostly the former.  It allows us to put all of the failure logic in a single place, rather than having it in many places making it less easily tested.  I don't think it's a huge deal either way.

twometresteve: I suspect you're right that it would be better and more "Ruby like".  My head is just struggling to compute any potential risks (and having to retest any changes).  I think it could be something for me to look at in future changes, but on balance I think I feel more inclined to leave it as it.  If there was more code involved then I'd probably look into it more now.

Cawllec (approved):

PR #729, opened 2025-03-04
Remove Aspecto repeater

twometresteve:
## Goal

Remove the Aspecto repeater, as Aspecto has now been shut down.

## Documentation

Removed from docs too.

## Tests

Covered by CI.


twometresteve:
> Will this break any runs that continues to specify the repeater key as a command line option? I doubt it's much of an issue, as we're probably setting it using environment variables, but it could be an issue?

You're right, it is actually technically a breaking change.  In practice, however, the command line option was never used, only envars, so I'd be happy to avoid the overhead of moving repos up with a major version bump.

Cawllec (approved):

PR #736, opened 2025-04-03
changelog

richardelms:
## Goal

Add percentage validator step for span attributes


richardelms:
implemented in the unity steps file instead

PR #751, opened 2025-05-02
Remove deprecated Cucumber steps and driver methods

twometresteve:
## Goal

Remove deprecated Cucumber steps and driver methods.

## Design

These steps are no longer needed as our test fixtures us HTTP Commands instead.

## Documentation

Upgrading guide updated.

## Tests

Covered by CI and further testing will be performed with notifier repos before releasing v10.


UPGRADING.md
```diff
@@ -6,6 +6,15 @@
 
 The legacy JSON-WP protocol drivers (using Selenium v3 and Appium v11) are no longer supported. 
 
+### Removed steps
+
+| Removed step  | Replacement                     |
+|---|---------------------------------|
+| `I wait for {int} span(s)`  | I wait to receive {int} span(s) |
```
tomlongridge, line 13: I don't see this in this PR - is that already made?

twometresteve: Yes, basically.  `I wait to receive {int} span(s)` spans is already in the repo and the change here is just to tidy away the one we don't want (in the deleted deprecated_steps.rb file).

twometresteve: I later realised that that's not the right step to direct people to, as the deprecated one wait for _at least_ the number of spans.  I'll correct that.

gingerbenw (approved):
All good from the JS side - I think the only change we'll need is to replace `I wait for {int} span(s)` when we accept v10 in the notifier

PR #753, opened 2025-05-12
Improve grouping of reported Selenium/Appium errors

twometresteve:
## Goal

This change improves the reporting and grouping of Selenium (and by-proxy, Appium) errors.

## Design

I've repurposed the `AmbiguousErrorMiddleware` (now `SeleniumErrorMiddleware`)  to inspect all Selenium/Appium errors reported by Bugsnag and change the grouping hash for any errors that match one of the regexes listed. When a regex is matched, a substitution is made to homogenize the message.  For example, to mask the app name from messages that are inherently the same error.

## Changeset

I've also moved `Maze::BugsnagConfig` into `Maze::ErrorMonitor::Config` and split the middleware classes out into their own files as part of an ongoing effort to improve the folder structure of the code base.

One other change is to enhance the logic in our monkey patch of `Selenium::WebDriver::Error::ServerError` to extract and use the message field of the payload as the error message.  This should improve the presentation of errors that currently look like this:

image

## Tests

Unit tests added for the regex matching on known error message formats and grouping agnostic of the app name is working nicely.

image

Despite several full test runs, I've not been able to reproduce an error that will exercise the change to the monkey patch, so I'll need to monitor that over the coming days.


lib/maze/error_monitor/config.rb
```diff
@@ -0,0 +1,59 @@
...
+                'builder': ENV['BUILDKITE_BUILD_CREATOR'],
+                'message': ENV['BUILDKITE_MESSAGE'],
+                'step': ENV['BUILDKITE_LABEL']
+              }
+              if ENV['BUILDKITE_JOB_ID']
+                metadata['job url'] = ENV['BUILDKITE_BUILD_URL'] + "#" + ENV['BUILDKITE_JOB_ID']
+              end
+            end
+            config.middleware.use(AssertErrorMiddleware)
+            config.middleware.use(SeleniumErrorMiddleware)
```
twometresteve, line 43: This line and the move into the `ErrorMonitor` module are the only real changes to code - GitHub just can't present the diff properly on a file move.

lib/maze/error_monitor/assert_error_middleware.rb
```diff
@@ -0,0 +1,25 @@
+module Maze
+  module ErrorMonitor
```
twometresteve, line 2: The only change here is to move it into it's own file in a new module.

joshedney (approved):

PR #794, opened 2025-09-17
Avoid deprecated devices and add iOS 26

twometresteve:
## Goal

Avoid the latest round of deprecated BrowserStack devices whilst adding iOS 26.

## Tests

Tested locally that each device name changed/added works.


joshedney (approved):
LGMT!

PR #827, opened 2026-02-06
Refresh Appium client

twometresteve:
## Goal

Updates the version of `appium_lib_core` whilst also removing the use of the `appium_li` wrapper (as is now the recommendation and reduces complexity for use).

## Design

The versions of `appium_lib_core`, 7.5.0, was carefully chosen as the last one before deprecated methods start to be removed.  We still use some of these methods where we deliberately restrict ourselves to using Appium 1 (for the sake of better stability).  We may be able to move this forward further in future when BitBar add support for Appium 3.

The `Appium` (driver) class is no longer marked as extending `Appium::Driver` and instead composition is preferred.  Due to the nature of how `Appium::Driver` is written in the `appium_lib_core` Gem, the type of the `Maze.driver` instance that we create was never what you would expect intuitively coming from a typical OO background.

This is marked as a major version bump, partly for the API change, but more through an abundance of caution to avoid breaking test suites that may rely of driver methods that were previously provided implicitly my inheritance on the driver class.

## Documentation

UPGRADING guide updated.

## Changeset

- `DeviceManager` API improved to allow the available log types to be queried before retrieval.

## Tests

Existing unit and e2e tests updated.


twometresteve:
The UPGRADING guide will follow in the next commit, but I wanted to get the rest into review given the time of day.

joshedney (approved):

PR #831, opened 2026-04-03
Update the valid device mapping for ANDROID_13

SB-khushbum:
## Goal

Updated the device mapping for `ANDROID_13` from `Google Pixel 6 Pro` to `Google Pixel 7`, as the previous device is no longer valid on BrowserStack.

## Tests

- Cloned the Android repo and set the local Maze Runner path in the Gemfile.
- Ran a test with the device set to `ANDROID_13`.
- The test failed with the old mapping (`Google Pixel 6 Pro`).
- Updated the mapping to `Google Pixel 7`.
- Re-ran the test and it passed.


SB-jigneshR (approved):

joshedney (approved):
Looks good to me. Again, probably worth a `CHANGELOG` update.

PR #832, opened 2026-04-16
appium-version 2.19/session_capability for failed driver

SB-khushbum:


SB-khushbum:
This was for testing purpose for appium version CI/CD and session_capability action

